PR #4, opened 2020-01-13
Setting unit tests and regression tests

sylvchev:
Unit tests should be populated
For regression tests, there is an error with handling MNE events and the obtained precision is not correct according to the hard-coded numbers.
This first pass correct the handling of the MNE events.


sylvchev:
I used the conda environment set up in PR #3 and I have the following version of packages: 
h5py==2.10.0
matplotlib==3.1.2
mne==0.19.2
numpy==1.17.3
pandas==0.25.3
resampy==0.2.2
scipy==1.4.1
torch==1.3.1

PR #22, opened 2020-01-14
Skorch model added for training

sliwy:
This is the first version of skorch model application for training models. As for now we have working training loop, predictions and evaluation step. Example flow defined in `examples/skorch_trial_decoding.py`.


examples/skorch_trial_decoding.py
```diff
@@ -0,0 +1,170 @@
...
+        return dataset[:n_train_samples], dataset[n_train_samples:]
+
+
+set_random_seeds(20200114, True)
+
+# final_conv_length = auto ensures we only get a single output in the time dimension
+model = ShallowFBCSPNet(
+    in_chans=in_chans,
+    n_classes=n_classes,
+    input_time_length=train_set.X.shape[2],
```
agramfort: n_times

agramfort: or maybe `n_times_input`

PR #27, opened 2020-01-16
move things out of torch_ext

agramfort:


robintibor:
Amazing! Great work! No killing of optimizer yet?

PR #29, opened 2020-01-16
rm AdamW; rm notebooks and mv perturbation to .py

agramfort:


agramfort:
next time @robintibor don't merge my PR if CIs are red :)

robintibor:
ya true @agramfort at the time only looked at travis since circleci was failing on master and did not remember the fix for that. Maybe fixed now in https://github.com/braindecode/braindecode/pull/35

PR #32, opened 2020-01-16
WIP: making datasets in braindecode

sbbrandt:
This PR contains a new example showing how to construct a pytorch-compatible dataset that is essentially a wrapper around mne objects; this includes the new raw transformer, window transformer and windower objects. The example also shows how to use moabb to load data into mne objects.


robintibor:
Move windower.py into datautil, authors in all files

robintibor:
After small discussion with Lukas: We need to ensure that default behavior for datasets that are used to do deep learning on it is such that the final supercrop that is partly out of trial bounds is shifted back inside and not dropped. The reason is that for my main examples we have a situation where just have 2 supercrops in a trial, and dropping the second may result in ignoring 1/3 of each trial... which is really not intended...
Basic idea should be, you can do trialwise decoding or switch to cropped decoding and set the supercrop size without thinking much about it, and always use all the signal in the trials for training.

braindecode/datasets/dataset.py
```diff
@@ -0,0 +1,67 @@
...
+    ----------
+    windows : mne.Epochs
+        epoched data to wrap
+    target : str
+        target specified by user to be decoded
+    transformer : sklearn.base.TransformerMixin
+        preprocessor function applied on windowed data
+    """
+
+    def __init__(self, windows, target='target', transformer=None):
```
agramfort: pytorchvision calls this transform or transforms. To be discussed

examples/plot_dataset_example.py
```diff
@@ -0,0 +1,113 @@
...
+# information contained in the 'numpy.array' the filter later is applied to.
+filter_window = FilterWindowTransformer(sfreq=250, l_freq=8, h_freq=12)
+
+# zscore for normalization
+zscorer_window = ZscoreWindowTransformer()
+
+# again, transformers can be chained using the scikit-learn Pipeline object
+transform_window_pipeline = Pipeline(
+    [('FIR_filter', filter_window), ('zscorer', zscorer_window)]
+)
```
agramfort: if transforms are just callables you can for loop the transforms and don't need sklearn pipeline. torch design if just with callables and the transformers here are all stateless not like in supervised learning where fit_transform is more than  transform

braindecode/datasets/dataset.py
```diff
@@ -0,0 +1,71 @@
...
+            index of the window to return
+
+        Returns
+        -------
+        x : ndarray, shape (n_channels, window_size)
+            one window of data
+        y : int | float
+            window target
+        """
+        x = np.squeeze(self.windows[index].get_data())
```
agramfort: I recommend to avoid using squeeze by any mean. Here you know what index you want to kill so just do
```suggestion
        x = self.windows[index].get_data()[0]
```

agramfort: also that this will typically be unecessarily slow as indexing epochs constructs a new epochs and makes a copy of windows.info which will take time and not needed here. I would consider adding a parameter to get_data in mne to avoid such copy

braindecode/datasets/moabb_datasets.py
```diff
@@ -0,0 +1,264 @@
+"""
+BCI competition IV 2a dataset
+"""
+import numpy as np
+import mne
+
+from torch.utils.data import ConcatDataset
+from braindecode.datasets.dataset import WindowsDataset
```
agramfort: nested import please

braindecode/datasets/moabb_datasets.py
```diff
@@ -0,0 +1,264 @@
...
+        raw transformers applied before windowing
+    windower : sklearn.base.TansformerMixim
+        windower transformer
+    window_transformer : sklearn.base.TansformerMixim
+        window transformer applied after windowing
+    transform_online : bool
+        if True, apply window transformers on the fly. Otherwise apply on loaded data.
+    """
+
+    def __init__(self, dataset_name, subject, raw_transformer=None, windower=None,
```
agramfort: raw_transformer -> raw_transforms?

braindecode/datasets/moabb_datasets.py
```diff
@@ -0,0 +1,264 @@
...
+                 transformer=None, transform_online=False, path=None):
```
agramfort: transformer -> transforms?

braindecode/datautil/lazy_iterators.py
```diff
@@ -4,9 +4,7 @@
 import torch as th
 import numpy as np
 
-from braindecode.datautil.iterators import (
-    _compute_start_stop_block_inds,
-)
+from braindecode.datautil.iterators import _compute_start_stop_block_inds
 from braindecode.util import get_balanced_batches
```
agramfort: relative imports here too

braindecode/datautil/transforms.py
```diff
@@ -0,0 +1,115 @@
...
+#          Lukas Gemein <l.gemein@gmail.com>
+#          Simon Brandt <simonbrandt@protonmail.com>
+#          David Sabbagh <dav.sabbagh@gmail.com>
+#
+# License: BSD (3-clause)
+
+import mne
+
+
+class FilterRaw(object):
```
agramfort: I would try to have only one Filter class that is smart enough to call .filter if method is there or just work with ndarray.

braindecode/datautil/transforms.py
```diff
@@ -0,0 +1,115 @@
...
+            windowed data to normalize
+
+        Returns
+        -------
+        X : ndarray, shape (n_channels, window_size)
+            normalized windowed data
+
+        """
+        windows -= windows.mean(axis=-1, keepdims=True)
+        return windows / windows.std(axis=-1, keepdims=True)
```
agramfort: same here. We should have only object object that is smart to work with mne Raw or ndarrays

braindecode/datautil/transforms.py
agramfort: the only thing is that if you have windows you need to pass explicitly the sfreq which might deserve a discussion

braindecode/models/eegresnet.py
```diff
@@ -18,150 +18,226 @@ class EEGResNet(nn.Sequential):
...
+        final_pool_length,
+        n_first_filters,
+        n_layers_per_block=2,
+        first_filter_length=3,
+        nonlinearity=elu,
+        split_first_layer=True,
+        batch_norm_alpha=0.1,
+        batch_norm_epsilon=1e-4,
+        conv_weight_init_fn=lambda w: init.kaiming_normal_(w, a=0),
+    ):
```
agramfort: is it some some form of automatic tool that did this reformating? If so I am not super happy as it makes large diffs to review. If we go this route then it should be made part of the CI system so formatting is not anymore a matter of taste but enforced why a system we agree on.

braindecode/util.py
```diff
@@ -226,9 +226,9 @@ def dict_equal(d1, d2):
     intersect_keys = d1_keys.intersection(d2_keys)
     modified = {o: (d1[o], d2[o]) for o in intersect_keys if d1[o] != d2[o]}
     return (
-        intersect_keys == d2_keys and
-        intersect_keys == d1_keys and
-        len(modified) == 0
+        intersect_keys == d2_keys
+        and intersect_keys == d1_keys
+        and len(modified) == 0
```
agramfort: flake8 suggests the opposite formatting...

examples/perturbation.py
```diff
@@ -98,29 +121,43 @@
...
-          input_time_length=input_time_length,
-          validation_data=(valid_set.X, valid_set.y),)
+model.fit(
+    train_set.X,
+    train_set.y,
+    n_epochs=30,
+    batch_size=64,
+    scheduler="cosine",
+    input_time_length=input_time_length,
+    validation_data=(valid_set.X, valid_set.y),
```
agramfort: I personally don't like exploding like this all function calls it makes very long files. It's not part of the pep8 guidelines.

examples/bcic_iv_2a_cropped.py
```diff
@@ -24,7 +24,7 @@
 )
```
agramfort: can you call your examples plot_... so they get rendered in the doc. 

See here they are there https://164-232335424-gh.circle-artifacts.com/0/dev/auto_examples/index.html  but it's not executed by the CircleCI. thanks

braindecode/models/eegresnet.py
hubertjb: @agramfort yes, the formatting was done automatically with [black](https://github.com/psf/black). I agree that this way of writing function calls is not space efficient and also doesn't look that good. I think it would make sense to either modify black's settings (if that's a thing we can actually do) or use some other tool, e.g. autopep8. @gemeinl @robintibor what do you think?

agramfort:
I would setup flake8 on CIs and agree on certains errors to ignore like
here https://github.com/braindecode/braindecode/blob/master/setup.cfg#L17
then don't edit code for cosmetic reasons if the CIs don't complain

I would just do the formatting myself with an editor that raises errors
when setup.cfg is not respected

>

robintibor:
Hi! So let's  have a vote if we want to enforce an autoformatter such as [black](https://github.com/psf/black) or [yapf](https://github.com/google/yapf) in the future :))
Basic idea of using and enforcing autoformatters is to have a clear consistent formatting (that may go beyond pep8 as well) that can be automatically applied. And not have style discussions anymore. At least as far as I understand, see for example https://medium.com/3yourmind/auto-formatters-for-python-8925065f9505  https://blog.frank-mich.com/python-code-formatters-comparison-black-autopep8-and-yapf/

Black is supposed to make diffs as small as possible (once it is applied consistently). Of course for using an autoformatter, we need to at some point to have a single commit that reformats everything and then use CI to enforce it afterwards. 
About the exploding lines, if we don't want that, we could switch to [yapf](https://github.com/google/yapf) 

You can play with both here:
* https://yapf.now.sh/
* https://black.now.sh

So let's vote by responding with emoji on this post, if mid/long-term we want:

* enforce a specific to-be-decided autoformatter (yapf or black)  autoformatting
* only check PEP8 on CI  pep8

To vote, click the smiley on the top right of **this comment** and choose icon as above :) For example I chose auto-formatter: autoformatting

robintibor:
Following our experiences last friday and discussions with @gemeinl , I think we may have to change  where to expose the location of the supercrops (i.e., per supercrop: which trial and start/stop within trial) in order to have stable and understandable code.

I propose the following:

For a dataset of braindecode, `dataset[idx]` returns `X,y,i` on  where `i` is a 3-tuple (or torch.Tensor of length 3) of   `i_trial, i_start, i_stop`.

Here, `i_trial` is the trial the supercrop was cropped from and `i_start, i_stop` are the supercrop start/stop positions within that trial (stop is exclusive as is standard in Python).

For the integration to work on the skorch-side, in our `BraindecodeClassifier` class, we override the `get_iterator` method to:

* Create a wrapper around the dataloader that removes these supercrop indices again so that skorch only sees X,y as it expects
* For a validation set dataloader, catch these supercrop indices during the validation loop and cache them in any existing `CroppedTrialEpochScoring` callbacks with `on_train=False` so they can be used to compute the scores


The advantages of that approach:
* on the PyTorch/Dataloader/Sampler side, we introduce no special functionality, and stick to the [interface](https://pytorch.org/docs/stable/data.html#torch.utils.data.Dataset) by only implementing `get_item`. This should allow us to use stuff like ConcatDataset, all kinds of samplers etc without additional problems
* the code to correct overlapping predictions may become both simpler and shorter after the change
* overriding `get_iterator` is an intended way to extend skorch as per https://github.com/skorch-ev/skorch/issues/336#issuecomment-419471143

Disadvantages:
* The additional `i` output in `X,y,i` may be surprising for users
* Our skorch extension to cache the indices for the scorers may be a bit surprising, we may not cover all possible cases (user-supplied special loaders, different scorer classes) etc. in our implementation

Overall, I feel this may be the best way to have a somewhat clean and understandable implementation of the cropped decoding stuff that may not require endless fixes later.

Let me know what you think or if you see any other, even simpler/cleaner way for implementing cropped decoding with correct scoring/metric computation. 

Otherwise we will start to move towards that interface.

agramfort:
I can live with X,y,i or actually X,y,info as you may want to return more metadata like the hour of the day/night for sleep data or any contextual information (age, gender etc.)

agramfort:
I can live with X,y,i or actually X,y,info as you may want to return more metadata like the hour of the day/night for sleep data or any contextual information (age, gender etc.)

agramfort:
regarding pep8 a start would be to have travis run flake8 on the code base
and start fixing the errors

>

robintibor:
Also, just to clear up any remaining confusion about what is a supercrop in the context of EEG (of course would be same for MEG) and other terminology as I used it so far :) :

For any EEG dataset, you may first define `trials`. Trials I view as any contiguous block of EEG you want to have predictions for and where directly before or after the trial you either don't want predictions at all (e.g., breaks between trials) or you want to see it as a new trial (e.g., different tasks without break inbetween).

Examples of trials could be 4-second motor imagery tasks,  variable-length motor imagery tasks, whole runs of EEG where you want to regress a continuous movement from, etc. 

Now, what are `supercrops`? :) Supercrops are:
* time windows *within* a trial that you define **mostly or only for computational reasons** (i.e., to fit the supercrop in GPU memory during the foward pass, the only non-computational reason I can think of would be to not have too correlated samples within a mini-batch during training). 
* an therefore an input that will be passed to the network during training (and possibly during inference as well)

For example you may have 4-second 100-Hz motor imagery trials, however for a batch size of 64, your mini-batch cannot be forwarded through your network. Therefore, you decide to use supercrops of size `2.5 seconds = 250 samples` so the mini-batch fits on GPU. But still, since you only did this for computational efficiency, in one epoch you want to collect all the predictions as if you were able to directly pass the 4-second trials. This is possible in the following way for a "dense-prediction" "timestep-causal" network. Dense-prediction here means from a certain timestep on, you have predictions for every timestep until the last timestep in the supercrop, without gaps inbetween. Timestep-causal means a prediction for a specific timestep only uses data up until that timestep, for example the second-last prediction in the supercrop is only computed from the data at the timesteps until the second last timestep, and is not affected by the data at the last timestep):
1. Do the forward pass for the first `2.5 seconds = 250 timesteps`, so timesteps 1-150 (in 1-based inclusive indexing). 
2. Check how many predictions come out of your network, e.g., 100. Now you know you have predictions for timestep 151-250 . The predictions 1-150 you cannot get, they are lost due to convolution/pooling kernel sizes. 
3. Now you want to get predictions 251-400.  For that you first shift your supercrop by 100 (number of predictions per supercrop) to 101-350. Same logic as before, you now get predictions 251-350.  
4. Missing are 351-400. Shifting by 100 is no longer possible as you would go out of bounds to 251-500. Therefore you shift to fit inside the trial to 151-400. You now get predictions  301-400. So you have predictions 301-300 twice and need to remove them during inference to correctly compute any metrics you are interested in. 

Finally, how to get these dense-prediction timestep-causal networks if you are using convolutional networks?
1. All strides need to be 1 (instead you can use dilated convolutions/pooling to have larger receptive fields (see https://arxiv.org/abs/1511.07122) 
2. Use no padding or left-padding

For our shallow and deep networks, we use no padding, which has the benefit that the predictions you get from your supercrops are *identical* to the predictions you get from directly passing the entire trial to the network (after removal of duplicates as explained above). Still, no padding complicates network design and left padding can be an alternative as done in for example WaveNet (https://arxiv.org/abs/1609.03499).

robintibor:
> I can live with X,y,i or actually X,y,info as you may want to return more metadata like the hour of the day/night for sleep data or any contextual information (age, gender etc.)

Yes, in the future we may add more, for now I suggest to focus on implementing a complete (dataset + EEGClassifier) pipeline for the general cropped decoding case in the clearest/cleanest possible way. Then we can see for the other use cases. For other information, it depends if it will be used by the model as well, then it should possibly be inside the `X` variable as `X` will be passed to the model by skorch.

robintibor:
Closing this since it is integrated into https://github.com/braindecode/braindecode/pull/54

robintibor:
> regarding pep8 a start would be to have travis run flake8 on the code base and start fixing the errors
> […](#)

Ya but would wait with this for the finish of the integration we are trying at the moment, hopefully mid next week there should be sth running. If we stick to only pep8, we might also consider to then run once the [autopep8](https://github.com/hhatto/autopep8) in the least aggressive option to make it easier for us, and later enforce only pep8 in CI or whatever vote is saying then.

PR #33, opened 2020-01-16
resnet model

TonioBall:
added eegresnet.py with residual network model

### DIFF
--- a/braindecode/models/eegresnet.py
+++ b/braindecode/models/eegresnet.py
@@ -1,0 +1,238 @@
+import numpy as np
+from torch import nn
+from torch.nn import init
+import torch as th
+from torch.nn.functional import elu
+
+from braindecode.models.base import BaseModel
+
+from braindecode.models.modules import Expression, AvgPool2dWithConv
+from braindecode.models.functions import identity
+from braindecode.models.modules import np_to_var
+
+
+class EEGResNet(BaseModel):
+    """
+    Residual Network for EEG.
+    """
+    def __init__(self, in_chans,
+                 n_classes,
+                 input_time_length,
+                 final_pool_length,
+                 n_first_filters,
+                 n_layers_per_block=2,
+                 first_filter_length=3,
+                 nonlinearity=elu,
+                 split_first_layer=True,
+                 batch_norm_alpha=0.1,
+                 batch_norm_epsilon=1e-4,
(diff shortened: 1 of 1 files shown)


robintibor:
Great

braindecode/models/eegresnet.py
```diff
@@ -0,0 +1,243 @@
...
+from torch import nn
+from torch.nn import init
+import torch as th
+from torch.nn.functional import elu
+
+from braindecode.models.base import BaseModel
+
+from braindecode.models.modules import Expression, AvgPool2dWithConv
+from braindecode.models.functions import identity
+from braindecode.models.modules import np_to_var
```
agramfort: you should use relative imports in the library

braindecode/models/eegresnet.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # Initialize all weights
+        model.apply(lambda module: weights_init(module, self.conv_weight_init_fn))
+
+        # Start in eval mode
+        model.eval()
+        return model
+
+
+def weights_init(module, conv_weight_init_fn):
```
agramfort: I would make this function private starting with _

braindecode/models/eegresnet.py
```diff
@@ -0,0 +1,243 @@
...
+        x = x[:,:,0]
+    return x
+
+
+def _transpose_time_to_spat(x):
+    return x.permute(0, 3, 2, 1)
+
+
+# create a residual learning building block with two stacked 3x3 convlayers as in paper
+class ResidualBlock(nn.Module):
```
agramfort: I assume it's not a public class so I would call it _ResidualBlock

braindecode/models/eegresnet.py
```diff
@@ -0,0 +1,243 @@
...
+from braindecode.models.base import BaseModel
+
+from braindecode.models.modules import Expression, AvgPool2dWithConv
+from braindecode.models.functions import identity
+from braindecode.models.modules import np_to_var
+
+
+class EEGResNet(BaseModel):
+    """
+    Residual Network for EEG.
```
agramfort: for new code let's force ourself to add docstrings

agramfort: it should also have been added to the api.rst page

TonioBall: hope docstrings are ok/better now?

PR #34, opened 2020-01-16
pep8 and try to add coverage tests for models

agramfort:
@robintibor @TonioBall please have a look. See that except one test they all fail. Suggestions welcome


robintibor:
Hi,
the assumptions in the tests are not correct for our models.
They will not return as many predictions as inputs in the time domain, since the pooling layers and unpadded convolutions reduce time resolution. Additionally, they will just crash for inputs that are too short in the time domain. Meaningful tests can still be written for shallow and deep testing final_conv_length='auto' vs other options. For others we have to see what meaningful tests exist.

PR #38, opened 2020-01-17
Sklearn api

robintibor:


braindecode/datautil/lazy_iterators.py
```diff
@@ -6,8 +6,8 @@
 
 from braindecode.datautil.iterators import (
     _compute_start_stop_block_inds,
-    get_balanced_batches,
 )
+from braindecode.util import get_balanced_batches
```
agramfort: Please use relative imports in the package

PR #41, opened 2020-01-17
modified:   braindecode/models/eegresnet.py

TonioBall:
changed to relative imports and added docstrings


agramfort:
Sorry @TonioBall i should have let you know that I had done it this morning.

PR #50, opened 2020-01-17
added a FilterBankRaw transformer

VincentRouanne:
Created new transformer for raw for filter bank, called FilterBankRaw. Basically adds virtual 'channels' that contain data filtered with different filters


gemeinl:
#65 now adds a new way of applying transforms to data (either mne.Raw or mne.Epochs): https://github.com/braindecode/braindecode/blob/master/braindecode/datautil/transforms.py#L18-L77. Would you be interested in continuing your work on a filterbank transform with this new API @VincentRouanne ?

robintibor:
for now not, possibly reactivate in future

PR #54, opened 2020-01-20
Dataset integration

robintibor:
Just a hardfixed integration (skorch+new dataset) for now, just so we have a running example at all. Not intended to stay like that, not intended to be merged yet.


agramfort:
I think there is a Pb in the commits of the PR

robintibor:
what does pb mean in that context @agramfort ? I merged the other pull request dataset_dev into this branch, was that not smart, is this is what is causing the problems?

robintibor:
Probably we should merge both these PRs into one again? Doing this seperately was in the first place only result of trying to somehow share last state of the code we were writing together before we had to leave INRIA with others on friday. @sliwy you have more experience with this as well, do you know how to merge this one and https://github.com/braindecode/braindecode/pull/32 intelligently without merging anything into master?

agramfort:
I am not sure I follow what you want to do. Merge this PR and kill the old
dataset_dev? Merge dataset_dev first and then this one?

>

sliwy:
I merged changes from `master` into `dataset-integration` and resolved conflicts in `eegresnet.py` (it occurred because the file was automatically formatted on `dataset-integration` branch and then modified on `master` branch).
From my perspective, if we want to merge this branch we can just close PR #32 for `dataset_dev` because this branch contains all commits from `dataset_dev` (we merged those branches on Friday). In case we would like to merge it this way we should not push anything to `dataset_dev` as changes won't be automatically added to this branch.

robintibor:
ok amazing @sliwy ! CIs run, great! Have closed https://github.com/braindecode/braindecode/pull/32 ! now we will work towards https://github.com/braindecode/braindecode/pull/32#issuecomment-576432931 ! There are two things I think:
1) changing the dataset itself towards `X, y, i` which @gemeinl and me are doing atm
2) changing our "dummy" `EEGDataset` towards the `X, y, i`, so that we can already implement the new logic on the skorch side. This is open, if you have no other tasks you could start that, maybe we can also do it together in some way via skype or sth... what would be your preference?

robintibor:
Great! I think for quick questions we can use gitter later. I will also send you my number there privately.

A bit more detail of the thoughts I had in mind for the implementation on the skorch-side. You can see whether they make sense and/or whether there are better ways:

In our `BraindecodeClassifier` class, we override the `get_iterator` method to:

* Create a wrapper around the dataloader (which will give `X, y, i`) that removes these supercrop indices again so that skorch only sees `X, y` as it expects
  * So you can just have a local function there that has a yield `X, y` inside, or explicitly define some class that takes original generator as input... whatever looks nicer to you in the end
* For a validation set dataloader, catch these supercrop indices during the validation loop and cache them in any existing `CroppedTrialEpochScoring` callbacks with `on_train==False` so they can be used to compute the scores. This will be most tricky part I think, maybe the following can work
  * Inside the wrapping-generator code there would be some check for all the callbacks of the current skorch-net if it has `CroppedTrialEpochScoring` with `on_train==False` and if we are returning the validation-generator at the moment
  * If yes to all that, then store only the current `i` in a new variable of the `BraindecodeClassifier` (e.g., `last_valid_supercrop_index`)
  * Override `on_batch_end` in the `BraindecodeClassifier` [method of skorch-net here](https://github.com/skorch-dev/skorch/blob/09be626e74512124eb74c76b4cbabad4d3b1f274/skorch/net.py#L312) (used in [here](https://github.com/skorch-dev/skorch/blob/09be626e74512124eb74c76b4cbabad4d3b1f274/skorch/net.py#L760))  Loop through the `CroppedTrialEpochScoring` with `on_train==False`, there are maybe several options what to do next
      1) Call a newly-defined function of `CroppedTrialEpochScoring` that stores the indices (including asserting that they are not None and that `on_train==False`
      2) Supply the indices directly to the on_batch_end function as some keyword argument (maybe only for the `CroppedTrialEpochScoring` callbacks, but maybe also to all since they can ignore extra keyword arguments) and have the aforementioned storing logic in the on_epoch_end function
  * In any case, after the `on_batch_end`, delete the `last_valid_supercrop_index` in `BraindecodeClassifier` again to make any errors less likely

I lean towards second option. My impression is that this will overall actually be an implementation that might not break too many assumptions of skorch?

braindecode/datautil/windowers.py
```diff
@@ -134,7 +135,7 @@ class EventWindower(BaseWindower):
     ----------
     window_size_samples : int | None
         size of one window in samples
-    overlap_size_samples : int
+    stride_samples : int
         size of overlap for window in samples
     drop_last_samples : bool
         See BaseWindower
```
gemeinl: @hubertjb @DavidSabbagh @sbbrandt  We now have a new Windower, see:
https://github.com/braindecode/braindecode/blob/1f01bbca020786af8cf3d5e14de941ae210952da/braindecode/datautil/windowers.py#L237-L301 this one computes supercrops outside of mne as it makes everything a bit easier for us our supercrop logic does not map well to mnes internal logic with chunk_duration.

test/unit_tests/datasets/test_dataset.py
```diff
@@ -0,0 +1,55 @@
+# Authors: Maciej Sliwowski <maciek.sliwowski@gmail.com>
+#
+# License: BSD (3-clause)
+
+import unittest
```
robintibor: @sliwy Could you switch to using pytest? Everywhere else we are using pytest, for setup stuff it has pytest.fixture (https://docs.pytest.org/en/latest/fixture.html). We use it for example here https://github.com/robintibor/braindecode-1/blob/47185b891ca96532095c71d80df9741b3aad77dc/test/unit_tests/datautil/test_windowers.py#L17

braindecode/datautil/windowers.py
sbbrandt: This looks good and is much clearer for me than the mne implementation before.

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,356 @@
...
+        assert supercrop_stride_samples > 0, (
+            "supercrop stride has to be larger than 0")
+        self.stride = supercrop_stride_samples
+        self.drop_samples = drop_samples
+        self.mapping = mapping
+
+    # TODO: handle case we don't get a raw
+    def __call__(self, raw):
+        # EventWindower: if annotations are given in raw, use them to get onsets
+        assert hasattr(raw, "annotations") and hasattr(raw.annotations, "onset")
```
sbbrandt: I think every mne.Raw object has an annotations attribute with an onset attribute. This statement will thus always be true. If annotations is not set, raw.annotations.onset will be a list of length 0. This will probably lead to a later error, e.g. when adding the `stop_offset` in `_supercrop_starts` to an empty list, resulting in another empty list.

If annotations are mandatory, you could go for `assert len(raw.annotations.onset) != 0`.

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,356 @@
...
+        fs = raw.info["sfreq"]
+        onsets = round_list_to_int(raw.annotations.onset*fs)
+        i_trials, starts, stops = _supercrop_starts(
+            onsets, self.start_offset, self.stop_offset, self.size, self.stride,
+            self.drop_samples)
+        description = raw.annotations.description
+        if self.mapping is not None:
+            description = [self.mapping[d] for d in description]
```
sbbrandt: Here first, description is an array of str. If mapping is set, description will become a list of int.

If mapping is set by the user, e.g. to only have specific subset of events, the mapping will not contain every description from the annotations object. This will probably lead to an error.

On the other hand, this is good to transform the str descriptions to event codes, but then mapping should be mandatory I think.

braindecode/scoring.py
```diff
@@ -12,9 +12,79 @@
...
+    """
+    Assigning supercrop predictions to trials  while removing duplicate
+    predictions.
+
+    Parameters
+    ----------
+    preds: list of ndarrays (atleast 2darrays)
+        List of supercrop predictions, in each supercrop prediction
+         time is in axis=1
+    i_supercrop_in_trials: list
```
sliwy: maybe we can switch from `i` to `idx` for all those variables? for me it would be more clear

robintibor: Created an issue for such questions in general https://github.com/braindecode/braindecode/issues/59

braindecode/datasets/dataset.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, windows, target="target", transforms=None):
+        self.windows = windows
+        self.target = target
+        if self.target != "target":
+            assert self.windows.info["subject_info"] is not None and \
+                   self.target in self.windows.info["subject_info"].keys()
+        self.transforms = (
+            transforms
+            if isinstance(transforms, list) or (transforms is None)
```
agramfort: I expect people to try to pass tuples too.

braindecode/datasets/dataset.py
```diff
@@ -0,0 +1,78 @@
...
+        """Return one window and its label
+
+        Parameters
+        ----------
+        index : int
+            index of the window to return
+
+        Returns
+        -------
+        x : ndarray, shape (n_channels, window_size)
```
agramfort: window_size -> n_times ? at least n_xxx

braindecode/datautil/lazy_iterators.py
```diff
@@ -4,9 +4,7 @@
 import torch as th
 import numpy as np
 
-from braindecode.datautil.iterators import (
-    _compute_start_stop_block_inds,
-)
+from braindecode.datautil.iterators import _compute_start_stop_block_inds
 from braindecode.util import get_balanced_batches
```
agramfort: use relative imports

braindecode/datautil/transforms.py
```diff
@@ -0,0 +1,115 @@
...
+class FilterRaw(object):
+    """Apply mne filter on raw data
+
+    Parameters
+    ----------
+    mne_filter_kwargs : **kwargs
+        kwargs passed to mne.io.Raw.filter
+    """
+
+    def __init__(self, **mne_filter_kwargs):
```
agramfort: I would avoid kwargs as much as possible as it's impossible to document easily

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+
+
+class FixedLengthWindower(BaseWindower):
+    """Fixed onset windower
+
+    ToDo: samples or seconds
+
+    Parameters
+    ----------
+    window_size_samples : int | None
```
agramfort: n_times ?

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+        size of one window in samples
+    overlap_size_samples : int
```
agramfort: n_times_overlap ?

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+        size of overlap for window in samples
+    drop_last_samples : bool
```
agramfort: we should pay attention with the word samples as it can be understood as samples for the learning task. Maybe just drop_last or drop_last_times?

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+        size of one window in samples
+    stride_samples : int
+        size of overlap for window in samples
+    drop_last_samples : bool
+        See BaseWindower
+    """
+
+    def __init__(
+        self,
+        window_size_samples,
```
agramfort: n_times

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+        stride_samples,
```
agramfort: stride

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+        drop_last_samples,
```
agramfort: drop_last

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,392 @@
...
+        trial_start_offset_samples=0,
```
agramfort: n_times_offset ?

braindecode/util.py
```diff
@@ -226,9 +226,9 @@ def dict_equal(d1, d2):
     intersect_keys = d1_keys.intersection(d2_keys)
     modified = {o: (d1[o], d2[o]) for o in intersect_keys if d1[o] != d2[o]}
     return (
-        intersect_keys == d2_keys and
-        intersect_keys == d1_keys and
-        len(modified) == 0
+        intersect_keys == d2_keys
+        and intersect_keys == d1_keys
+        and len(modified) == 0
```
agramfort: flake8 suggests the previous way.

braindecode/datautil/windowers.py
hubertjb: @gemeinl Thanks! Please let me know if I understand correctly: you made a new class that reimplements what `EventWindower` was meant to do, but with the following differences: (1) it uses a new internal function (`_supercrop_starts`) instead of `mne.events_from_annotations` to identify events, (2) it uses 'crop'/'supercrop' language instead of 'windows' in the variable names and docstring, (3) it sets the default behaviour to `drop_samples=False` (instead of requiring the user to set this explicitly) and (4) it doesn't inherit from the `BaseWindower` class.

So, a couple of questions/points:

1. Before I forget, I think docstrings should start with a one-line short description on the same line as the first `"""`: see https://www.python.org/dev/peps/pep-0008/#documentation-strings
2. Do you see this new Windower as a replacement for `EventWindower` or is this meant as something more specific to the crop/supercrop case? If it is meant to replace `EventWindower`, why did you chose to go this route instead? Specifically, why did you chose to replace `mne.events_from_annotations`? Personally, I think the closer we stay to the MNE way of doing things, the simpler it will be to maintain the package.
3. If it is indeed meant to replace `EventWindower`, do you think we should keep the crop/supercrop-specific language? My worry is that since this class is central to the package, and these concepts are relatively niche, they might make the understanding of the library more difficult. As far as I understand, `supercrop` could just be replaced by `window` here as there is no subdivision into crops happening at this level.
4. I understand that you guys' use case is to extract a window that covers the last "missed" samples in each event. I still believe that this default behaviour can be problematic in some cases (as the opposite behaviour can be problematic in others). I liked our previous solution to this - what do you think?
5. I see there is commented code indicating that this class would also handle the use case for `FixedLengthWindower`. Are you planning on merging the two?

braindecode/datasets/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+class BaseDataset(Dataset):
+    """
+    A base dataset.
+
+    Parameters
+    ----------
+    raw: mne.Raw
+    info: pandas.DataFrame
+        holds additional information about the raw
```
hubertjb: What is the expected shape of the DataFrame? I don't see where this is being used right now and I'm not sure what it is supposed to be containing.

braindecode/datasets/base.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    def __init__(self, raw, info, target=None):
+        self.raw = raw
+        self.info = info
+        if target is not None:
+            assert target in self.info, f"'{target}' not in info"
+        self.target = target
+
+    def __getitem__(self, index):
+        return self.raw, self.target
```
hubertjb: Is there a use case for returning the entire Raw object as a Raw object? It could be left NotImplemented if there is not.

braindecode/datasets/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def __len__(self):
+        return len(self.raw)
+
+
+class WindowsDataset(BaseDataset):
+    """
+    Applies a windower to a base dataset.
```
hubertjb: Maybe "Dataset wrapper around an mne.Epochs object"? There is no windower here, the data is already passed as windows.

braindecode/datasets/base.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Parameters
+    ----------
+    windows: ConcatDataset
```
hubertjb: Should be `mne.Epochs` if I understand correctly.

braindecode/datasets/base.py
```diff
@@ -0,0 +1,62 @@
...
+        windows/supercrops obtained throiugh application of a Windower to a
```
hubertjb: Crop/Supercrop-specific language: do we want to keep this here?

braindecode/datasets/base.py
```diff
@@ -0,0 +1,62 @@
...
+        BaseDataset
+    info: pandas.DataFrame
+        hold additional info about the windows
+    """
+    def __init__(self, windows, info):
+        self.windows = windows
+        self.info = info
+
+    def __getitem__(self, index):
+        target = self.windows.events[:,-1]
```
hubertjb: Should be `target = self.windows.events[index, -1]` to avoid copying the entire last column every time.

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+    from moabb.datasets.utils import dataset_list
+    for dataset in dataset_list:
+        if dataset_name == dataset.__name__:
+            # return an instance of the found dataset class
+            return dataset()
+    raise ValueError("'dataset_name' not found in moabb datasets")
+
+
+def _fetch_and_unpack_moabb_data(dataset, subject_ids):
+        data = dataset.get_data(subject_ids)
```
hubertjb: Should be one indentation level lower.

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+        supercrops/windows do not equally devide the continuous signal
+    ignore_events: bool
+        when True, ignores events specified in mne.Raw and uses a
+        FixedLenthWindower to create supercrops/windows
+
+    """
+    # TODO: include preprocessing at different stages
+    def __init__(
+            self, dataset_name, subject_ids, trial_start_offset_samples,
+            trial_stop_offset_samples, supercrop_size_samples,
```
hubertjb: crop/supercrop-specific language: do we want to keep this?

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+            supercrop_stride_samples, drop_samples=False, ignore_events=False):
+        if ignore_events:
```
hubertjb: Do we want to create the windower internally? This could be problematic if the different windowers end up expecting different arguments. Also, this limits the capacity of a user to make changes to the way windows are extracted. Personally, I'd prefer if we explicitly asked the user for a `Windower` object. Each specific dataset (maybe not this one though, as it is for every MOABB dataset) could have a default Windower object that is initialized if no windower is defined, to extract the most usual windows for this dataset.

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,200 @@
...
+        supercrop size
+    supercrop_stride_samples: int
+        stride between supercrops
+    drop_samples: bool
+        whether or not have a last overlapping supercrop/window, when
+        supercrops/windows do not equally devide the continuous signal
+    mapping: dict(str: int)
+        mapping from event description to target value
+    """
+    def __call__(self, base_ds):
```
hubertjb: Why do we pass a Dataset object here instead of a mne.Epochs object? Wouldn't it be simpler to directly work with the Raw object?

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+            trial_start_offset_samples=trial_start_offset_samples,
+            trial_stop_offset_samples=trial_stop_offset_samples,
+            supercrop_size_samples=supercrop_size_samples,
+            supercrop_stride_samples=supercrop_stride_samples,
+            drop_samples=drop_samples)
+
+        raws, info = fetch_data_with_moabb(dataset_name, subject_ids)
+        all_windows_ds = []
+        for raw_i, raw in enumerate(raws):
+            base_ds = BaseDataset(raw, info.iloc[raw_i])
```
hubertjb: I left a comment in the windower script - Why do we first create a BaseDataset out of the Raw object? Wouldn't it be simpler to just pass the Raw object directly to the Windower object?

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+    Parameters
+    ----------
+    dataset_name: name of dataset included in moabb to be fetched
+    subject_ids: list(int) | int
+        (list of) int of subject(s) to be fetched
+    trial_start_offset_samples: int
+        start offset from original trial onsets in samples
+    trial_stop_offset_samples: int
+        stop offset from original trial onsets in samples
+    supercrop_size_samples: int
```
hubertjb: Again, supercrops vs. windows?

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+
+    mapping: dict{target_value: int}
+        maps target values to integers
+    """
+
+    def __init__(self, path, trial_start_offset_samples,
+                 trial_stop_offset_samples, supercrop_size_samples,
+                 supercrop_stride_samples, subject_ids=None,
+                 drop_samples=False, target="pathological", mapping=None):
+        windower = FixedLengthWindower(
```
hubertjb: I haven't used the TUH dataset yet, but could someone want to make windows based on Events instead of sliding windows? In either case, I think it would be simpler to ask for a Windower instead of passing all the arguments directly (again, there could be a default).

braindecode/datasets/datasets.py
```diff
@@ -0,0 +1,446 @@
...
+                session = "train"
+            else:
+                assert "eval" in path_splits
+                session = "eval"
+            age, gender = _parse_age_and_gender_from_edf_header(file_path)
+            info = pd.DataFrame(
+                [[age, pathological, gender, session, subject_id]],
+                columns=["age", "pathological", "gender",
+                "session", "subject"], index=[subject_id])
+            base_ds = BaseDataset(raw, info, target=target)
```
hubertjb: Need to go through a BaseDataset first?

braindecode/datautil/transforms.py
```diff
@@ -0,0 +1,115 @@
...
+        -------
+        raw : mne.io.Raw
+            normalized raw data
+
+        """
+        raw = raw.apply_function(lambda x: x - x.mean())
+        return raw.apply_function(lambda x: x / x.std())
+
+
+class FilterWindow(object):
```
hubertjb: Note: @agramfort had suggested merging the two Filter objects and the two Zscore objects in #32

braindecode/datautil/windowers.py
hubertjb: I just saw the updated code - will repost updated comment below!

braindecode/datautil/windowers.py
```diff
@@ -0,0 +1,200 @@
...
+#          David Sabbagh <dav.sabbagh@gmail.com>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+import mne
+import pandas as pd
+
+
+class Windower(object):
```
hubertjb: Updating my comments from above now that I've seen the latest version:

If I understand correctly, we have a new set of Windower classes which (1) use a new internal function (_supercrop_starts) instead of mne.events_from_annotations to identify events, (2) use 'crop'/'supercrop' language instead of 'windows' in the variable names and docstring and (3) set the default behaviour to drop_samples=False (instead of requiring the user to set this explicitly).

1. Before I forget, I think docstrings should start with a one-line short description on the same line as the first """: see https://www.python.org/dev/peps/pep-0008/#documentation-strings
2. Why did we chose to use a new internal function instead of the existing MNE functions? Personally, I think the closer we stay to the MNE way of doing things, the simpler it will be to maintain the package. Also, people already familiar with MNE don't have to learn a new way of extracting windows.
3.  Do you think we should keep the crop/supercrop-specific language? My worry is that since this class is central to the package, and these concepts are relatively niche, they might make the understanding of the library more difficult. As far as I understand, supercrop could just be replaced by window here as there is no subdivision into crops happening at this level.
4. I understand that you guys' use case is to extract a window that covers the last "missed" samples in each event. I still believe that this default behaviour can be problematic in some cases (as the opposite behaviour can be problematic in others). I liked our previous solution to this which was to force the user to provide drop_samples - what do you think?

hubertjb:
@robintibor @gemeinl @sliwy @sbbrandt I added some comments to the current version of the code.

One general remark that I repeated multiple times in the comments: we have to decide whether we want to keep the supercrop/crop language in base classes. If classes/functions are not supercrop/crop-specific, I would vote for using the term "windows" instead. Whatever we decide, we should stick to one term.

Also, I think we should try to limit the scope of this PR and merge it soon, so we can open up parallel PRs to work on the Dataset class (to keep working on this PR it looks like I would have to push to @robintibor's personal fork, which might actually be OK, but I'm not sure it's best practice).

braindecode/datasets/base.py
gemeinl: Shape of DataFrame is 1 (row) x n_infos (cols). I took a photo of the flipchart in Inria where you guys designed the dataset classes. There was a BaseDataset which contains information on a dataset level. (There was also a EpochsDataset that contains information on a sample level. Or as I know interpreted it, on a windows/supercrop level, since windows/supercrops can have the size of 1 sample in the extrema.) For example, for MOABB datasts the BaseDataset info can be subject_id, session_id, and run_id. For the Temple University dataset, this can be the age and gender of patients as well as the pathology status of EEG recordings. (The WindowsDataset info will then be i_supercrop_intrial, i_start_of_supercrop, i_stop_of_supercrop, which are required for correctly combining supercrop/window predictions as well as the supercrop/window target.)

braindecode/datasets/base.py
gemeinl: Right now, I removed the transformers from our MOABB and TUH datasets, to get a simple, first running version. In another comment Alex pointed out that we would like to have smart transformers that know whether they are working on raw, continuous data or on epoched data. If they should operate on raw data, this could be a reason to query the raw.

braindecode/datasets/base.py
gemeinl: I did not pay enough attention to this. Language needs some adjustment. Especially in the datasets and windowers. I have no strong preference whether we should call it "windows" or "supercrops".

braindecode/datasets/datasets.py
gemeinl: Yes, you are right. I talked about this to Robin on friday as well. We should move the windower out of the specific datasets again and leave it to the user. I also like your idea of a deafult windower.

braindecode/datautil/windowers.py
gemeinl: 1. I wasn't aware of this. Thanks for pointing this out. I will pay more attention to this in the future.
2. As @robintibor pointed out, the supercrop/window logic creation did not translate well to MNE and added even more complexity to the problem. Just think about tmin, tmax, chunk_duration (which changes its behavior), etc.. There is actually no need to use MNE at this point. To me, the new _supercrop_starts function is clearer than everything we had before.
3. @robintibor please decide on the language and we can then stick to it.
4. I am sorry, I forgot about this. I think, it will be best to not set a default and force the user to make a decision. Therefore, he will then also be responsible for the consequences.

braindecode/datasets/datasets.py
gemeinl: There are no events defined for the TUH data, so I wouldn't know how to do it. Again, I agree, there should be a default for the windower, leaving the possibility to the user to do whatever he wants.

braindecode/datautil/transforms.py
gemeinl: I think we also said that they do not have to (should not) follow scikit-lean API as it is unnecessary. Also, we said they should be "smart" in a sense that they know whether they work on continuous, raw or epoched data.

braindecode/datautil/windowers.py
gemeinl: In the case of the FixedLenthWidnower, we need additional information that is not contained in the Raw object. Since there are no events defined for the TUH data, the target is provided in the BaseDataset on a dataset level. We have to retrieve this information when cutting supercrops/windows, to assign a target to every supercrop.

braindecode/datasets/datasets.py
gemeinl: Thanks for pointing this out!

braindecode/datasets/base.py
gemeinl: Thanks!

braindecode/datasets/datasets.py
gemeinl: I am not sure if it is actually needed. But, I think we need something addition to the mne Raw object to hold subject_id, session_id, run_id, age, gender, pathology status, or else.

braindecode/datasets/base.py
hubertjb: Ah I get it, this makes sense! What do you think of using a simple dict instead of a DataFrame here?

braindecode/datasets/base.py
hubertjb: I think the idea behind having transformers that can work on Raw or Epochs is so that there is a single class per type of transform (instead of having one for preprocessing on continuous data and one for windowed data).

I realize that this might just be for getting a first running version as you say, but wasn't the WindowsDataset initially meant to apply a (windows) transformer just before returning a window? If so, transformers should be applied inside a Dataset class, and not outside. In that case, I would expect the same thing to happen in a BaseDataset, and so BaseDataset's __getitem__ would first apply a continuous transformer (if necessary) and then return a numpy array instead of a Raw object. What do you think?

braindecode/datasets/base.py
gemeinl: The WindowsDataset adopts (or concats) the info DataFrame from its base datasets (see https://github.com/robintibor/braindecode-1/blob/dataset-integration/braindecode/datasets/datasets.py#L236, https://github.com/robintibor/braindecode-1/blob/dataset-integration/braindecode/datasets/datasets.py#L360)
It can be used to conveniently create groups within the WindowsDataset based on pandas groupby function (which you also wrote down on the flipchart :)). See <code> plot_dataset_example.py</code> for an example!

braindecode/datautil/windowers.py
hubertjb: Can you give examples of targets that you would use on the TUH data (I guess normal/abnormal, age,sex, etc.)? I was thinking there might be a way to save that data inside the Raw.info structure - in that case, we could work directly with a Raw object an avoid another level of abstraction. The FixedLengthWindower could extract all the fields we care about from the info structure and put them in the `metadata` attribute of the Epochs object.

@agramfort what do you think about adding other fields to the Raw.info structure to pass other recording-level labels (e.g., normal/abnormal)?

gemeinl: The dataset was designed to decode normal/abnormal. However, we have also used age and sex of patients in some experiments. I guess, these are the obvious decoding choices. You are right, adding information to fields in mne Raw objects is another way of doing this. However, as I said above, it might be an advanage to have this info in a DataFrame for convenient data splits. For example, maybe you want to decode the pathology status only from recordings of patients that exceed a certain age.

robintibor:
added the point with supercrop naming to the existing issue about code style https://github.com/braindecode/braindecode/issues/59

Regarding scope of this PR: It should have running example of BCIC IV 2a cropped decoding with skorch that produces good results. Then we have a clear example that should run that we can improve on without going too far into different directions I think. This will help us be faster later.

braindecode/datasets/base.py
hubertjb: Yes, I remember that now. :P 

1- It is a small implementation detail then, but we might as well discuss it: since the info dataframe only contains one row in BaseDataset, it could be a simple dict (simplifies the code a bit). Is there a case where a DataFrame might be preferable, knowing that concatenating from dicts is as simple as if not simpler then it is with DataFrames?

2- As for the WindowsDataset, where do you see the groupby functionality happening? Would that be on the internal metadata attribute of the Epochs object? In that case, if I understand correctly, we would need to add columns to the metadata object with 'age', 'sex', 'pathological', etc.

braindecode/datasets/base.py
hubertjb: @gemeinl Reading another one of your answers below made me think: is it because we potentially want to concatenate BaseDatasets directly without passing through mne.Epochs? In that case I would understand the point for a DataFrame (although this might be covered already by the length-1 epoch case we tested during the sprint).

braindecode/datasets/base.py
hubertjb: @gemeinl Reading another one of your answers below made me think: is it because we potentially want to concatenate BaseDatasets directly without passing through mne.Epochs? In that case I would understand the point for a DataFrame (although this might be covered already by the length-1 epoch case we tested during the sprint).

braindecode/datasets/datasets.py
hubertjb: Got it! In that case, I think we could find a way for the windower to expect that information, and for its output to also contain that information, either as (1) a structure inside the mne.Epochs, (2) a separate output argument or (3) as part of the metadata DataFrame.

Maybe not essential for now, but it might be cleaner.

braindecode/datautil/windowers.py
hubertjb: Thanks for the answers! :)

2. Maybe we could discuss with @agramfort whether there's a modification to MNE's API that would make this simple here? I agree that the new function is clearer to read, but ultimately I think offloading whatever can be done by MNE to MNE would help in maintaining the codebase.

robintibor:
As promised, a hacky, yet mergeable version now @gemeinl @agramfort @sliwy @hubertjb @sbbrandt 

Thanks for all your efforts for this new dataset and trainer API!

A great news, we had possibly the first case since the remote working time that a test written by one person catches a big written by another person. Thanks @sliwy  for [test_scoring.py](https://github.com/robintibor/braindecode-1/blob/a733c99a1240d13c20120b89c4eb3e4aff832475/test/unit_tests/test_scoring.py) :+1: :100:   amazing

@hubertjb  could you start to collect issues out of unadressed comments of yours in here? Those that are not covered by existing issues. Maybe also connect with @gemeinl  via gitter to discuss

To reclarify one vision/use case I see for braindecode is to have simple high-performance library for deep learning EEG decoding. So one issue will be to drastically simplify basic examples using good defaults etc.

A next milestone could be to reach the decoding performance of the [X/y/manual BCIC IV 2a example ](https://github.com/robintibor/braindecode-1/blob/a733c99a1240d13c20120b89c4eb3e4aff832475/examples/bcic_iv_2a_cropped.py) with the new moabb dataset  and a clear simple code (current not well-performing, not simple version [here](https://github.com/robintibor/braindecode-1/blob/a733c99a1240d13c20120b89c4eb3e4aff832475/examples/plot_bcic_iv_2a_moabb.py)). 

I will then merge this PR tomorrow unless there are any objections? Then we could work in different parts, I see dataset/preprocessing more with @gemeinl  and @hubertjb  and trainer/skorch-based improvements with @sliwy @robintibor ? I am unfortunately a bit time-constrained for the next 10 days as we have to rescue an ICML submission after realizing our main idea was already published :)

examples/bcic_iv_2a_cropped.py
```diff
@@ -14,61 +14,25 @@
 
 import torch
 from torch import optim
-from torch.utils.data import Dataset
```
agramfort: @robintibor can you name the examples with plot_ so they get rendered in the doc.

See for https://194-232335424-gh.circle-artifacts.com/0/dev/auto_examples/bcic_iv_2a_cropped.html#sphx-glr-auto-examples-bcic-iv-2a-cropped-py

it contains a hard coded link to file you are the only one to have. Can you relay on some moabb dataset?

examples/plot_dataset_example.py
```diff
@@ -0,0 +1,58 @@
+"""
+Simple Moabb Dataset Example
+=========================
+
+Showcasing how to fetch and crop a moabb dataset.
+"""
```
agramfort: this one is rendered fine https://194-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_dataset_example.html#sphx-glr-auto-examples-plot-dataset-example-py

test/unit_tests/test_scoring.py
```diff
@@ -220,3 +237,45 @@ def on_epoch_end(self, net, *args, **kwargs):
     )
 
     clf.fit(train_set, y=None, epochs=4)
+
+
+def check_preds_supercrops_trials(preds, supercrop_inds, expected_trial_preds):
```
agramfort:
```suggestion
def _check_preds_supercrops_trials(preds, supercrop_inds, expected_trial_preds):
```

it makes it clear that this function is a private/aux function from otherones. NOt to be called directly

examples/bcic_iv_2a_cropped.py
robintibor: This was a thing we discussed briefly in Paris I think. There are examples (right now, the two not with `plot_` at the start) that are meant for reproducing an actual pipeline from a paper and may require GPU/longer computing times. As far as I remember, you suggested either change to having parameters that allow fast runtimes and some hint that they need to be changed to get good results or remove `plot_` to not have them rendered. I opted for second choice now. Still, of course would be better to fetch the dataset without local folder path etc. Would prefer to have this not block current PR merge, so created separate issue: https://github.com/braindecode/braindecode/issues/61

test/unit_tests/test_scoring.py
robintibor: Thanks, followed your suggestion in latest commit.

examples/bcic_iv_2a_cropped.py
agramfort: I would see if we can rather run it with very little number of epochs (maybe just 1) to allow checks it runs and tell clearly on the top what to change to get fully reproducible results.

PR #64, opened 2020-02-04
Improved Dataset classes and Windower interaction

gemeinl:
Created a BaseConcatDataset, which also implements a way to split into subsets.
Moved Windower out of MOABBDataset/TUHAbnoraml.
There are now functions to create windows, no more Windower classes.
See also https://github.com/braindecode/braindecode/issues/63

PR #65, opened 2020-02-05
[WIP] Improved Dataset classes and Windower interaction (continued)

hubertjb:
This PR includes the work of @gemeinl (#64) and adds a few fixes and tests. I'm making a new PR instead of working on the existing one since I wasn't sure how to contribute to @gemeinl's PR directly.

Original changes in #64:
* Created a BaseConcatDataset, which also implements a way to split into subsets
* Moved Windower out of MOABBDataset/TUHAbnormal.
* There are now functions to create windows, no more Windower classes.

Additional changes:
* Some internal changes to Datasets to simplify manipulation of descriptions
* Recomputing `BaseConcatDataset.cumulative_sizes` after applying transforms to account for potential change of length
* Updating examples and tests

### DIFF
--- a/braindecode/datautil/transforms.py
+++ b/braindecode/datautil/transforms.py
@@ -1,5 +1,5 @@
 """Transforms that work on Raw or Epochs objects.
-ToDo: should transformer also transform y (e.g. cutting continuous labelled 
+ToDo: should transformer also transform y (e.g. cutting continuous labelled
       data)?
 """
 
@@ -24,13 +24,14 @@ def transform_concat_ds(concat_ds, transforms):
     transforms: dict(str | callable: dict)
         dict with function names of mne.raw or a custom transform and function
         kwargs
-        
+
     Returns
     -------
     concat_ds:
     """
-    assert isinstance(transforms, OrderedDict), (
-        "Order of transforms matters! Please provide an OrderedDict.")
+    if not isinstance(transforms, OrderedDict):
+        raise TypeError(
+            "Order of transforms matters! Please provide an OrderedDict.")
     for ds in concat_ds.datasets:
         if hasattr(ds, "raw"):
             _transform_raw(ds.raw, transforms)
@@ -52,7 +53,8 @@ def _transform_raw(raw, transforms):
         if callable(transform):
             _custom_transform_raw(raw, transform, transform_kwargs)
         else:
-            assert hasattr(raw, transform), f"raw does not have {transform}"
+            if not hasattr(raw, transform):
+                raise AttributeError(f"Raw does not have {transform}.")
             _mne_transform(raw, transform, transform_kwargs)
 
 
@@ -69,8 +71,8 @@ def _transform_windows(windows, transforms):
         if callable(transform):
             _custom_transform_windows(windows, transform, transform_kwargs)
         else:
-            assert hasattr(windows, transform), (
-                f"epochs does not have {transform}")
+            if not hasattr(windows, transform):
+                raise AttributeError(f"Epochs does not have {transform}.")
             _mne_transform(windows, transform, transform_kwargs)
 
 
@@ -92,7 +94,7 @@ class FilterRaw(object):
 
     def __call__(self, raw):
         """Apply filter
-        
+
         Parameters
         ----------
         raw : mne.io.Raw
--- a/braindecode/datautil/transforms.py
+++ b/braindecode/datautil/transforms.py
@@ -10,7 +10,65 @@ ToDo: should transformer also transform y (e.g. cutting continuous labelled
 #
 # License: BSD (3-clause)
 
+from collections import OrderedDict
 import mne
+
+
+def transform_concat_ds(concat_ds, transforms):
+    """Apply a number of transformers to a concat dataset.
+
+    Parameters
+    ----------
+    concat_ds: A concat of BaseDataset or WindowsDataset
+        datasets to be transformed
+    transforms: dict(str | callable: dict)
+        dict with function names of mne.raw or a custom transform and function
+        kwargs
+    Returns
+    -------
+    concat_ds:
+    """
+    assert isinstance(transforms, OrderedDict), (
+        "Order of transforms matters! Please provide an OrderedDict.")
+    for ds in concat_ds.datasets:
+        if hasattr(ds, "raw"):
+            _transform_raw(ds.raw, transforms)
+        else:
+            assert hasattr(ds, "windows"), (
+                "Can only tranform concatenation of BaseDataset or "
+                "WindowsDataset")
+            _transform_windows(ds.windows, transforms)
+
+
+def _transform_raw(raw, transforms):
+    for transform, transform_kwargs in transforms.items():
+        if callable(transform):
+            _custom_transform_raw(raw, transform, transform_kwargs)
+        else:
+            assert hasattr(raw, transform), f"raw does not have {transform}"
+            _mne_transform(raw, transform, transform_kwargs)
+
+
+def _custom_transform_raw(raw, transform, transform_kwargs):
+    transform(raw, **transform_kwargs)
+
+
+def _mne_transform(raw_or_epochs, transform, transform_kwargs):
+    getattr(raw_or_epochs.load_data(), transform)(**transform_kwargs)
+
+
+def _transform_windows(windows, transforms):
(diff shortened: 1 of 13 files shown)


hubertjb:
When this is merged, I'll open a short PR based on the present PR to require the latest MNE dev version: https://github.com/hubertjb/braindecode/tree/require-latest-mne

braindecode/datautil/transforms.py
```diff
@@ -10,9 +10,76 @@
...
+        if callable(transform):
+            _custom_transform_windows(windows, transform, transform_kwargs)
+        else:
+            if not hasattr(windows, transform):
+                raise AttributeError(f"Epochs does not have {transform}.")
+            _mne_transform(windows, transform, transform_kwargs)
+
+
+def _custom_transform_windows(windows, transform, transform_kwargs):
+    transform(windows, **transform_kwargs)
```
sbbrandt: `_transform_raw` and `_transform_windows` as well as `_custom_transform_raw` and `_custom_transform_windows` could be merged into one function I think. The differentiation at line 35 is just used for the error messages and could be done inside the merged function.

examples/plot_dataset_example.py
```diff
@@ -1,58 +1,133 @@
...
+# Transforms can also be applied on supercrops/windows in the same way as shown
+# above on continuous data:
+
+def crop_windows(windows, start_offset_samples, stop_offset_samples):
+    fs = windows.info["sfreq"]
+    windows.crop(tmin=start_offset_samples / fs, tmax=stop_offset_samples / fs,
+                 include_tmax=False)
+
+epochs_transform_dict = OrderedDict({
+    "pick_types": {"eeg": True, "meg": False, "stim": False},
```
sbbrandt: Lukas and I thought about discarding the stimulus channel right after getting the events, to prevent eventual training on data still containing the labels in the data array. That way you wouldn't need to call pick_types here again.

braindecode/datautil/windowers.py
```diff
@@ -66,95 +45,145 @@ class EventWindower(Windower):
...
+            if starts[-1] < last_allowed_ind:
+                starts[-1] = last_allowed_ind
 
-        # TODO: handle multi-target case
-        assert len(base_ds.info[base_ds.target]) == 1, (
-            "multi-target not supported")
-        description = base_ds.info[base_ds.target].iloc[0]
+        # TODO: handle multi-target case / non-integer target case
+        if ds.target is None:
+            target = 1  # default target
```
gemeinl: What was the idea behind this default target here @hubertjb ? I think this is critical and should not be. It might cause running classification on wrong targets without the user even noticing. I think we should raise an error here, when the target is None.

braindecode/datautil/transforms.py
hubertjb: Good point, I'll merge them!

examples/plot_dataset_example.py
gemeinl: To elaborate a bit more. For the BCIC IV 2a dataset, the events are defined by the stimulus channel. We call mne.find_events inside the create_windows_from_events function. This means that we still need the stimulus channel after preprocessing the raw. Additionally, we have to make sure that the user does not use the stimulus channel for decoding, since it contaisn the targets.

hubertjb: I see, I think that makes sense. I suppose you would add that in the windower functions?

Also, I just realized `pick_types` doesn't work inplace (according to the documentation). If that's the case, or generally speaking for methods that don't work inplace, did you guys think about how that could be done? Would we have to define a wrapper around the method?

braindecode/datautil/windowers.py
hubertjb: I see your point, we don't want to create fake targets that a user could end up using by mistake. The reason I included a default target is I believe there are some situations where you don't actually want to specify a target here: 1) you're not going to use a target in a supervised classification sense (e.g., structured prediction tasks, autoencoding, etc.) or 2) the target will be created by the sampler itself later (e.g., in a contrastive learning task where the sampler picks two windows and provides a target of 0 if they are far away or 1 if they are close).

One solution would be to have the default target be a type that couldn't be used as a target later on, and so an unsuspecting user would get an error when training. What do you think?

examples/plot_dataset_example.py
hubertjb: > To elaborate a bit more. For the BCIC IV 2a dataset, the events are defined by the stimulus channel. We call mne.find_events inside the create_windows_from_events function. This means that we still need the stimulus channel after preprocessing the raw. Additionally, we have to make sure that the user does not use the stimulus channel for decoding, since it contaisn the targets.

One way around this would be to use MNE's annotations objects instead of stim channels. I know it adds some complexity to the code (especially since in BCI-like datasets it's convenient to work in samples instead of in seconds), but it's more natural for sleep recordings for instance.

gemeinl: Yes, dropping the stimulus channel right after extracting the events in the windowing functions could be one way to do it.

I wasn't aware that pick_types does not work in place... Is there a way around it using mne? Maybe defining bad channels instead of picking channels?

braindecode/datautil/windowers.py
gemeinl: If you want a type that cannot be used as target, None would work. However, in the create_windows_from_events we explicitly require an integer target right now: https://github.com/braindecode/braindecode/blob/1bbe5fe376664330a4dedbd58509f37c8d18c8d5/braindecode/datautil/windowers.py#L159-L163

We could do it like this:
```
target = ds.target
if mapping is not None, ("a mapping from '{target}' to int is required"):
    target = mapping[target]
```

examples/plot_dataset_example.py
gemeinl: > One way around this would be to use MNE's annotations objects instead of stim channels. I know it adds some complexity to the code (especially since in BCI-like datasets it's convenient to work in samples instead of in seconds), but it's more natural for sleep recordings for instance.

So instead of using the stimulus channel in the windowing functions, you suggest that we use annotations, is that correct? And this means that we extract events from the stimulus channel in a preprocessing step and add them to the annotations before dropping channels and before windowing? Did I get this right?

hubertjb: > So instead of using the stimulus channel in the windowing functions, you suggest that we use annotations, is that correct? And this means that we extract events from the stimulus channel in a preprocessing step and add them to the annotations before dropping channels and before windowing? Did I get this right?

Yes, that's it. The windowers would need to rely on mne.Annotations. I think the Dataset class itself (e.g., MOABBDataset) should create and populate an Annotations object to be placed in the Raw object (and then drop the stim channels).

braindecode/datautil/windowers.py
hubertjb: Ok cool, I think that would work!

hubertjb: @gemeinl actually that didn't work, because MNE requires an int in the third column of the event array. One hacky-ish solution would be to use an int (e.g., 0) in the third column of the event array when calling mne.Epochs(), but actually saving None in the metadata object. What do you think?

examples/plot_dataset_example.py
sbbrandt: Especially as some MOABB datasets already contain the annotations object instead of a stim channel, that's a good thing to do, I think. MOABBDataset could just check if the annotations object is empty and populate it from the stim channel if that's the case. I think we implemented something similar in an earlier version already.

braindecode/datautil/windowers.py
gemeinl: Ah, yes, I see. Maybe we could go with -1? I think you already did this in 4613b63. And yes, we should save None in the metadata. Eventually, the target is read from there: https://github.com/braindecode/braindecode/blob/4613b639976b6330c16ced433fd46a72c5cfdd08/braindecode/datasets/base.py#L65

examples/plot_dataset_example.py
gemeinl: Yes, you are right @sbbrandt. That's why we had (and still have but don't use) https://github.com/braindecode/braindecode/blob/4613b639976b6330c16ced433fd46a72c5cfdd08/braindecode/datasets/datasets.py#L21-L131.
It would be great to find a consistent way for all (MOABB) datasets, such that the windowing function will always get the events from mne.Annotations.

PR #66, opened 2020-02-07
[WIP] Adding requirement to download latest mne master

hubertjb:
This allows using recent improvements such as faster epoch loading with preload=False.

Also, this PR removes now unused copy-pasted mne functions in datasets.py and includes minor updates to setup.py and the readme.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,7 @@ setup(
     long_description=long_description,
 
     # The project's main homepage.
-    url='https://github.com/robintibor/braindecode',
+    url='https://github.com/braindecode/braindecode',
 
     # Author details
     author='Robin Tibor Schirrmeister',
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -53,8 +54,7 @@
 
         # Specify the Python versions you support here. In particular, ensure
         # that you indicate whether you support Python 2, Python 3 or both.
-        'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
```
gemeinl: Python version should be consistent everywhere. In environment.yml it is also 3.6, in travis.yml it is 3.7. We will go with 3.7 as it also introduces dataclasses (https://docs.python.org/3/library/dataclasses.html) which we might want to use.

hubertjb:
```suggestion
        'Programming Language :: Python :: 3.7',
```

PR #70, opened 2020-02-07
[WIP] Moving towards annotations object

sbbrandt:
-windower now expects annotations object to get events
-tests adapted to new annotation object
-removed custom mapping from bcic IV 2a example

The MOABB datasets will now be equipped with the annotations object. The trial onset and duration are set correctly based on information from MOABB.

The annotation object stores the string description of events. Thus, the original mapping is discarded and replaced by a custom one when applying the event windower. By default, the event mapping will start from 0, as requested by training for classification. The mapping is passed to mne.Epochs and stored on windows level.

### DIFF
--- a/braindecode/datasets/croppedxy.py
+++ b/braindecode/datasets/croppedxy.py
@@ -36,8 +36,9 @@ class CroppedXyDataset(Dataset):
         for i_trial, trial in enumerate(X):
             idx = _compute_supercrop_inds(
                 np.array([0]),
-                0,
                 trial.shape[1],
+                0,
+                0,
                 input_time_length,
                 n_preds_per_input,
                 drop_samples=False,
--- a/braindecode/datasets/croppedxy.py
+++ b/braindecode/datasets/croppedxy.py
@@ -35,12 +35,12 @@ class CroppedXyDataset(Dataset):
         i_supercrop_to_idx = []
         for i_trial, trial in enumerate(X):
             idx = _compute_supercrop_inds(
-                np.array([0]),
-                trial.shape[1],
-                0,
-                0,
-                input_time_length,
-                n_preds_per_input,
+                starts=np.array([0]),
+                stops=trial.shape[1],
+                start_offset=0,
+                stop_offset=0,
+                size=input_time_length,
+                stride=n_preds_per_input,
                 drop_samples=False,
             )
             for _, i_supercrop_in_trial, start, stop in zip(*idx):
--- a/braindecode/datasets/datasets.py
+++ b/braindecode/datasets/datasets.py
@@ -28,13 +28,16 @@ def _find_dataset_in_moabb(dataset_name):
             return dataset()
     raise ValueError("'dataset_name' not found in moabb datasets")
 
-
 def _fetch_and_unpack_moabb_data(dataset, subject_ids):
     data = dataset.get_data(subject_ids)
     raws, subject_ids, session_ids, run_ids = [], [], [], []
     for subj_id, subj_data in data.items():
         for sess_id, sess_data in subj_data.items():
             for run_id, raw in sess_data.items():
+                # set annotation if empty
+                if len(raw.annotations) == 0:
+                    annots = _annotations_from_raw(raw, dataset)
+                    raw.set_annotations(annots)
                 raws.append(raw)
                 subject_ids.append(subj_id)
                 session_ids.append(sess_id)
@@ -42,6 +45,22 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids):
     description = pd.DataFrame(zip(subject_ids, session_ids, run_ids),
                                columns=["subject", "session", "run"])
     return raws, description
+
+
+def _annotations_from_raw(raw, dataset):
+    # find events from stim channel
+    events = mne.find_events(raw)
+
+    # get annotations from events
+    event_desc = {k: v for v, k in dataset.event_id.items()}
+    annots = mne.annotations_from_events(events, raw.info['sfreq'], event_desc)
+
+    # set trial on and offset given by moabb
+    onset, offset = dataset.interval
+    annots.onset += onset
+    annots.duration += offset - onset
+
+    return annots
 
 
 def fetch_data_with_moabb(dataset_name, subject_ids):
--- a/braindecode/datasets/datasets.py
+++ b/braindecode/datasets/datasets.py
@@ -36,7 +36,9 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids):
             for run_id, raw in sess_data.items():
                 # set annotation if empty
                 if len(raw.annotations) == 0:
-                    annots = _annotations_from_raw(raw, dataset)
+                    annots = _annotations_from_moabb_stim_channel(
+                        raw, dataset
+                    )
                     raw.set_annotations(annots)
                 raws.append(raw)
                 subject_ids.append(subj_id)
@@ -47,7 +49,7 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids):
     return raws, description
 
 
-def _annotations_from_raw(raw, dataset):
+def _annotations_from_moabb_stim_channel(raw, dataset):
     # find events from stim channel
     events = mne.find_events(raw)
 
--- a/braindecode/datautil/windowers.py
+++ b/braindecode/datautil/windowers.py
@@ -38,7 +38,7 @@ def create_windows_from_events(
     trial_start_offset_samples: int
         start offset from original trial onsets in samples
     trial_stop_offset_samples: int
-        stop offset from original trial onsets in samples
+        stop offset from original trial stop in samples
     supercrop_size_samples: int
         supercrop size
     supercrop_stride_samples: int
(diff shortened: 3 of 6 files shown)


braindecode/datasets/datasets.py
```diff
@@ -44,6 +47,22 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids):
     return raws, description
 
 
+def _annotations_from_raw(raw, dataset):
```
hubertjb: I would rename this function to something like `_annotations_from_stim_channel` as it specifically requires stim channels to be available.

braindecode/datasets/datasets.py
```diff
@@ -44,6 +47,22 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids):
...
+    # find events from stim channel
+    events = mne.find_events(raw)
+
+    # get annotations from events
+    event_desc = {k: v for v, k in dataset.event_id.items()}
+    annots = mne.annotations_from_events(events, raw.info['sfreq'], event_desc)
+
+    # set trial on and offset given by moabb
```
hubertjb: If this function is specific to MOABB datasets, then maybe the function name should also reflect that.

test/unit_tests/datasets/test_dataset.py
```diff
@@ -51,8 +51,8 @@ def set_up():
 def concat_ds_targets():
     raws, description = fetch_data_with_moabb(
         dataset_name="BNCI2014001", subject_ids=4)
-    events = mne.find_events(raws[0])
-    targets = events[:, -1]
+    events, _ = mne.events_from_annotations(raws[0])
+    targets = events[:, -1] - 1
```
hubertjb: Is the -1 still necessary?

test/unit_tests/datautil/test_windowers.py
```diff
@@ -19,8 +19,8 @@
 def concat_ds_targets():
     raws, description = fetch_data_with_moabb(
         dataset_name="BNCI2014001", subject_ids=4)
-    events = mne.find_events(raws[0])
-    targets = events[:, -1]
+    events, _ = mne.events_from_annotations(raws[0])
+    targets = events[:, -1] - 1
```
hubertjb: -1 still necessary?

braindecode/datautil/windowers.py
```diff
@@ -60,9 +60,12 @@ def create_windows_from_events(
...
     list_of_windows_ds = []
     for ds in concat_ds.datasets:
+        if mapping is None:
+            # mapping event descriptions to integers from 0 on
+            mapping = {v: k for k, v in enumerate(
+                np.unique(ds.raw.annotations.description))}
 
-        # TODO: how to get events without 'stim' channel?
-        events = mne.find_events(ds.raw)
+        events, _ = mne.events_from_annotations(ds.raw, mapping)
```
hubertjb: I think this is a good improvement over the stim channel-only method!

I believe there is still another use case that this approach doesn't cover though. What if you have annotations for some clinical events (e.g., sleep stages, seizures, etc.) that are not necessarily of the same length? For example, you might have a sleep stage annotation that is 10-min long, followed by another one that is only 60-s long, etc. Let's say you are interested in extracting as many non-overlapping 30-s windows as possible inside each annotation. Currently, this function would only let you extract windows inside a fixed region following each event found (specifically, between `trial_start_offset_samples` and `trial_stop_offset_samples`), that is the same across all annotations. In this case, you would then be stuck with only extracting windows out of the first 60 s of each sleep stage annotation.

A possible solution would be to use the `chunk_duration` argument of `events_from_annotations` which already covers this scenario.  The function `_compute_supercrop_inds` could then be repurposed to find the `i_trials`, `i_supercrop_in_trials`, `starts` and `stops` for each event spit out by `events_from_annotations`. Another possibility would be to reimplement the `chunk_duration` functionality into `_compute_supercrop_inds` if that's simpler for now. Eventually though we could see whether there would be interest on MNE's side to add these kinds of outputs to `events_from_annotations`.

braindecode/datasets/datasets.py
sbbrandt: Yes, I agree.

braindecode/datasets/datasets.py
sbbrandt: Yes, I also agree. I used to have moabb in the name but somehow changed it.

test/unit_tests/datasets/test_dataset.py
sbbrandt: The annotations object only stores the string description for events. Thus, the mapping from string to integer is lost. When calling `create_windows_from_events`, I create a custom mapping from 0 on, as we need that for the classification.

`mne.events_from_annotations` internally creates a default mapping when no custom one is specified, that starts from 1. That's why there is a `-1` here.

Another option would be to use a specified mapping as done here:

https://github.com/sbbrandt/braindecode-1/blob/6877169232a5ffcad0ed5073eddb484ec917ad56/braindecode/datautil/windowers.py#L63-L68

braindecode/datautil/windowers.py
sbbrandt: As far as I get it, `_compute_supercrop_inds` uses `trial_stop_offset_samples` to determine the length of an event. The events object we had earlier did not include any information about the event length, so we had to specify this.

The annotations object now includes a duration from which we can recover the `trial_stop_offset_samples` for each event, as specified by the experiment. We could set this to optional, if you want the whole event to be included, and rely on the durations from the annotations object. If `chunk_duration` is given, this will be our future window length and the `events_from_annotations` will return the right events.

Do you think that will work?

hubertjb: Yes, I think that would work! The only thing I'd say is that `chunk_duration` is actually equivalent to `stride` rather than window `size` (unless we also use the `overlap` argument :P).

test/unit_tests/datasets/test_dataset.py
hubertjb: Got it, thanks!

braindecode/datautil/windowers.py
sbbrandt: True, it just gives you the onsets ;)

Anyway, If you try to get as many same length intervals into a trial, `chunk_duration` should be your `window size` as well as your `stride`, I guess.

There is one issue with mne though. When providing `chunk_duration` in `events_from_annotations`, the last samples are dropped, if the remaining part is smaller than `chunk_duration`. Our option whether to drop the last samples will become nasty to implement then.

hubertjb: > Anyway, If you try to get as many same length intervals into a trial, `chunk_duration` should be your `window size` as well as your `stride`, I guess.

Isn't that the case only if you want an overlap of 0? 

> There is one issue with mne though. When providing `chunk_duration` in `events_from_annotations`, the last samples are dropped, if the remaining part is smaller than `chunk_duration`. Our option whether to drop the last samples will become nasty to implement then.

If you remember I think we started implementing something like this during the sprint, it was indeed not super clean. But in the end I think the logic is not too complicated: if drop_last = False, we find the onset of a window that ends exactly at the end of the annotation; if that offset is not already in the list of onsets, we add it.

sbbrandt: > Isn't that the case only if you want an overlap of 0?

True

> If you remember I think we started implementing something like this during the sprint, it was indeed not super clean. But in the end I think the logic is not too complicated: if drop_last = False, we find the onset of a window that ends exactly at the end of the annotation; if that offset is not already in the list of onsets, we add it.

Yes, we already had that. I didn't remember ;)

gemeinl:
To me, the most urgent next step would be to have <code>plot_bcic_iv_2a_moabb.py</code> reach the same level of accuracies as before. We need at least one stable example with fetching and loading through moabb, preprocessing with mne, and training with skorch before implementing all kinds of possible use cases we might need in the future, and before we get lost in these details again. @sbbrandt I think you started with this already?

robintibor:
Yes, I agree, moving with this work directly towards currently unfinished `plot_bcic_iv_2a_moabb.py` to have a fully working realistic example and also see how the usage for the user is. If there is a specific use case one of us needs then best implement an example in the ideal code you wish you could have from the user side (not caring about how to do that internally). Then we can see very clearly what we need to do. I find this strategy often very efficient.

sbbrandt:
Yes, I had an example running, using the annotations version of the event windower and thus the correct trial times. I only converted the signal to muV in preprocessing and reached accuracy above 50%.

I can implement an example similar to the example in @robintibor old braindecode version, including the preprocessing.

braindecode/datautil/windowers.py
sbbrandt: As @robintibor mentioned, we discussed the chunk duration issue already a lot during the sprint. For now, we would like to leave that to our own function, as this is making it clear what happens.

I now changed the way, the event windower handles the `trial_offset_samples`, namely as a real offset from the annotation objects duration.

braindecode/datautil/windowers.py
```diff
@@ -177,7 +187,7 @@ def create_fixed_length_windows(
 
 
 def _compute_supercrop_inds(
-        onsets, start_offset, stop_offset, size, stride, drop_samples):
+        onsets, stops, start_offset, stop_offset, size, stride, drop_samples):
```
robintibor: Also add stops to documentation, rename onsets->starts

braindecode/datasets/croppedxy.py
```diff
@@ -36,8 +36,9 @@ def __init__(self, X, y, input_time_length, n_preds_per_input):
         for i_trial, trial in enumerate(X):
             idx = _compute_supercrop_inds(
                 np.array([0]),
-                0,
                 trial.shape[1],
+                0,
```
robintibor: maybe call by name, this is getting confusing

braindecode/datautil/windowers.py
```diff
@@ -211,37 +213,40 @@ def _compute_supercrop_inds(
...
+                supercrop_starts.append(stop - size)
                 i_supercrop_in_trials.append(i_supercrop_in_trials[-1] + 1)
-                i_trials.append(onset_i)
+                i_trials.append(start_i)
 
     # update stops to now be event stops instead of trial stops
-    stops = np.array(starts) + size
-    assert len(i_supercrop_in_trials) == len(starts) == len(stops)
-    return i_trials, i_supercrop_in_trials, starts, stops
+    stops = np.array(supercrop_starts) + size
```
robintibor: But then this also supercrop_stops, no?

robintibor:
Great. merged!

PR #71, opened 2020-02-07
removed dependecy on the resampy library

gemeinl:


hubertjb:
@gemeinl I just realized, don't forget to modify `setup.py` too.

robintibor:
Great!

PR #75, opened 2020-02-12
Lazy loading example

hubertjb:
This PR adds an example that compares lazy to eager loading, as well as varies a few other parameters (`num_workers`, batch size, using cuda devices or not, etc.). There are still things to figure out, but I wanted to submit to PR to get feedback.

I followed the format of an example, however maybe this should be a benchmark test instead.

## Current limitations
1. We don't have a fetcher for TUH, and so the example script will only run if the users already have a local copy. This also means that the CI won't pass. We could make a fetcher, however a username/password is required to download the TUH dataset; I'm not sure how to take that into account. Another option would be to include a "lazy-loadable" toy dataset in the repo just for testing purposes.
2. I used seaborn to make a summary figure at the end of the example. If we want to avoid this dependency I can make something in pure matplotlib/pandas.
3. The CI won't be able to test the case `cuda=True`. Is that a problem?
4. I wasn't able to control the number of workers (`num_workers` in `DataLoader`) like I should - whenever I set it to something above 0 **all** CPUs would get recruited. I tried different things (`torch.set_num_threads()` in main and subprocesses, setting env variable OMP_NUM_THREADS) but that didn't change anything. In the end, I had to run the script with 
`>> taskset -c 0-7 python plot_lazy_vs_eager_loading.py`


hubertjb:
Also, here's the end figure in the example (since the CI will fail), obtained on a single V100 GPU and 8 CPU cores:

image

Main observations are 1) lazy loading is competitive with eager loading when `num_workers=8`, and 2) for a currently unknown reason it looks like Deep4 and Shallow are very similar in terms of execution time.

robintibor:
Thanks for the plots, can you also write here table with mean numbers (not per subject).

robintibor:
And for me I don't think we need to have our performance tests run as an example on CI? So I would just remove plot_ from beginning of file maybe.

braindecode/datasets/base.py
```diff
@@ -57,16 +57,15 @@ class WindowsDataset(BaseDataset):
     description: pandas.Series
         holds additional info about the windows
     """
+    md_keys = ['i_supercrop_in_trial', 'i_start_in_trial', 'i_stop_in_trial']
```
robintibor: This seems a bit confusing, why defined at that place? Why not as before? Also md_keys seems confusing as this is keys to get indices, but tehre are appartenly other metadatas as well, like `target`

test/unit_tests/datautil/test_windowers.py
```diff
@@ -126,6 +126,29 @@ def test_overlapping_trial_offsets(concat_ds_targets):
             drop_samples=False)
 
 
+def test_windows_from_events_preload_false(concat_ds_targets):
```
robintibor: If you want to have this test, then you could create dummy data and store in temp-folder. We should not have unfinished test in master I think

robintibor:
Do you have time to proceed here @hubertjb ? Or are you quite busy atm then we could also have a look

hubertjb:
@robintibor I'm almost done, I made a few improvements and ran a more exhaustive benchmark since your last commented. I've been a bit busy, but will get to it by Tuesday if that's OK with you guys.

robintibor:
ya sounds great no problem @hubertjb

test/unit_tests/datautil/test_windowers.py
hubertjb: Updated with newest commit!

braindecode/datasets/base.py
hubertjb: I've changed this with the newest commit so that the list is initialized in the init instead. Although this is probably not an issue, having it there means the get method doesn't have to create the same list every time it's called. Also it could eventually let us add an argument to let the user specify the columns to return instead of using the default ones.

hubertjb:
@robintibor I've realized that the window dimensionality also has an impact on training time (from a data loading perspective):

image

Again, this looks like having a sufficiently high `num_workers` mostly takes care of this issue.

hubertjb:
> Thanks for the plots, can you also write here table with mean numbers (not per subject).

image

Sorry, it's a bit dense. I can send you the csv if you'd like too.

hubertjb:
> Thanks for the plots, can you also write here table with mean numbers (not per subject).

Here's the full table (I can send you the CSV too if you'd like):
image

braindecode/util.py
```diff
@@ -346,3 +348,69 @@ def round_list_to_int(a):
     :return: a with values rounded to integer
     """
     return np.round(a).astype(np.int)
+
+
+def create_mne_raw(n_channels, n_times, sfreq, include_anns=True, savedir=None,
```
robintibor: If this only creates fake data, rename to create_mne_dummy_raw

braindecode/datasets/base.py
```diff
@@ -60,13 +60,13 @@ class WindowsDataset(BaseDataset):
     def __init__(self, windows, description):
         self.windows = windows
         self.description = description
+        self.index_info_cols = [
```
robintibor: Please move the variable back into the function. Reason is that variables should in general be in smallest scope, see  for example https://softwareengineering.stackexchange.com/a/388055 for soem reasoning on this. I expect recreating list in method to not have any noticeable impact on performance. If that is wrong, e.g. performance tests show otherwise, we can move it out then. If we need the variable outside the function for other reasons, as you mentioned as a future possibility, we will move it outside at that point.

braindecode/datautil/windowers.py
```diff
@@ -159,7 +177,7 @@ def create_fixed_length_windows(
             # if last supercrop does not end at trial stop, make it stop
             # there
             if starts[-1] < last_allowed_ind:
-                starts[-1] = last_allowed_ind
+                starts = np.append(starts, last_allowed_ind)
```
robintibor: this is now different to before, was it wrong before?

hubertjb: Good point, I forgot about this change. Can you tell me whether it does what you are expecting or not?

If my understanding is correct, what this line was doing before was **moving** the last valid window to end at the end of the recording (or at `stop_offset_samples` if specified). What the new version does is **add** a new window that ends at the end of the recording (or at `stop_offset_samples`). This way we don't loose the last valid window.

robintibor: Hm ya I went through the code as you have written it now and propose following:
Change
```python

starts = np.arange(
     start_offset_samples, stop, supercrop_stride_samples)

last_allowed_ind = stop - supercrop_size_samples
```
to
```python
last_allowed_ind = stop - supercrop_size_samples
starts = np.arange(
     start_offset_samples, last_allowed_ind+1, supercrop_stride_samples)
```
Remove
```python
if starts[-1] > last_allowed_ind:
    starts = starts[:-1]
```

Keep rest as you wrote it (so with append etc).

robintibor: With this change for me I think PR would be mergeable, maybe would try to also run some performance test myself, but could also do it post-merge

hubertjb: Awesome, that's much better!

Yes, it'd be great if you could run the benchmark on your setup too, either before or after the merge.

gemeinl:
Can you maybe also change line https://github.com/braindecode/braindecode/blob/94010bf5983e831869b2ee59348910b40ecf61f9/braindecode/datautil/windowers.py#L184 to add an offset <code>int(ds.raw.first_time * ds.raw.info["sfreq"]) to <code>start</code> to account for potential cropping of the raw prior to extracting windows. Or should we do this in an separate PR?

robintibor:
Great merged! Yes @gemeinl  let's do seperate PR!

robintibor:
Will run own comparison post-merge

Issue #73, opened 2020-02-10
Improve braindecode.signalproc


gemeinl:
- Functions <code>exponential_running_standardize</code> and <code>exponential_running_demean</code> should be adapted to accept <code>data</code> input as <code>n_channels x n_times</code>.
- <code>highpass_cnt</code>, <code>low_pass_cnt</code>, <code>bandpass_cnt</code>, and <code>filter_is_stable</code> should be removed and usages be replaced with mne functions and follwoing transform API (https://github.com/braindecode/braindecode/blob/master/braindecode/datautil/transforms.py#L18).

gemeinl:
I believe something went wrong here. Neither #77 nor #79 implement the changes suggested in this issue.

robintibor:
closed by #80

PR #80, opened 2020-02-14
Improve braindecode.signalproc

gemeinl:
Implementing the changes suggested in #73.


examples/bcic_iv_2a_cropped.py
```diff
@@ -69,37 +66,24 @@
 assert len(raw_train.ch_names) == 22
 # lets convert to millvolt for numerical stability of next operations
 raw_train = mne_apply(lambda a: a * 1e6, raw_train)
```
agramfort: could we use raw. apply_function here instead of a custom mne_apply?

gemeinl: @agramfort mne_ext submodule with function mne_apply is removed in PR #82. Function calls to mne_apply are replaced by raw.apply_function()

Issue #48, opened 2020-01-17
Make EEGClassifier have very nice interface


robintibor:
Put very good defaults so calling without anything will usually give you a good result.

sliwy:
I just take a short look at current examples to see what can be improved in the EEGClassifier interface. I think that the most unfriendly part is the callbacks part which takes a lot of lines and does really similar thing for `CroppedDecoding` and `TrialDecoding` but differs in details. My idea is to create two list of standard callbacks in the `callbacks.py` file so we can just import `cropped_callbacks` or `trial_callbacks`. This way we will remove callback definition from examples. Do you think that we should hide more from the user? @robintibor I think that other things should be easily accessible by a user (like `batch_size`, `optimizer`, `lr`, `train_test_split`).

robintibor:
I feel this may be a little too opaque for the user, how about we allow to supply a list of scikit-learn scoring functions like ['accuracy', 'f1_score'] either as `cropped_scorings` or as `trial_scorings` and rest of code is created internally? Then you can still decide which scorings you want but they are automatically correctly created and by default applied to both train and valid... ? How you think @sliwy

sliwy:
Seems legit to me and won't need a lot of modification. I'll try to implement.

sliwy:
@robintibor Do you know any way to infer if the callbacks should be `Cropped` or `Trial` in the `EEGClassifier`?

sliwy:
@robintibor Take a look at it: https://github.com/sliwy/braindecode/commit/82f212698dc6ff003840c01ff1465d1e9a3faad7 it does not support `CroppedDecoding` for now. First version for `TrialDecoding`, should be better tested, now just to show the conception.

robintibor:
Ya this seems really nice https://github.com/sliwy/braindecode/commit/82f212698dc6ff003840c01ff1465d1e9a3faad7. regarding knowing cropped or not, I think this can be a constructor_parameter `cropped=True/False`. I think it's not bad if this is written explicitly by user

PR #83, opened 2020-02-16
New base model class

sliwy:
I looked a little bit at the example code (I wanted to do #48 but I started looking at the example and I've seen output shape inferring and then `to_dense_prediction_model` function which modifies model in place. Both things to me look strange, so I started with removing those two parts).

In my opinion, the only way to remove it is to create one more class from which all models should inherit. This class inherits from `torch.nn.Sequential`, so it should not add too much complexity. Everything in the models stays pretty the same. Now `Model` class implements `to_dense_prediciton`, so in the example instead of calling `to_dense_prediction_model(model)` we call `model.to_dense_prediction()`, which is more an aesthetic change. In the `Model` there is also `output_shape` property that computes nn output shape. I moved also squeezing and time/spatial transposition to this class, as almost all models used it and it was defined many times in different places.

One more thing I changed is reading the parameters in models `__init__` as it was done by hacky deleting of `self`. Now it is done in a standard way. 

Let me know what do you think about adding the `Model` class. Do you think that in the future there should be more stuff to put it inside this class?

### DIFF
--- a/braindecode/models/eegresnet.py
+++ b/braindecode/models/eegresnet.py
@@ -10,6 +10,7 @@ from torch import nn
 from torch.nn import init
 from torch.nn.functional import elu
 
+from .functions import transpose_time_to_spat, squeeze_final_output
 from ..util import np_to_var
 from .modules import Expression, AvgPool2dWithConv
 
@@ -18,7 +19,8 @@ class EEGResNet(nn.Sequential):
     """
     Residual Network for EEG.
     """
-    def __init__(self, in_chans,
+    def __init__(self,
+                 in_chans,
                  n_classes,
                  input_time_length,
                  final_pool_length,
@@ -31,14 +33,24 @@ class EEGResNet(nn.Sequential):
                  batch_norm_epsilon=1e-4,
                  conv_weight_init_fn=lambda w: init.kaiming_normal_(w, a=0)):
         super().__init__()
+        self.in_chans = in_chans
+        self.n_classes = n_classes
+        self.input_time_length = input_time_length
         if final_pool_length == 'auto':
             assert input_time_length is not None
         assert first_filter_length % 2 == 1
-        self.__dict__.update(locals())
-        del self.self
+        self.final_pool_length = final_pool_length
+        self.n_first_filters = n_first_filters
+        self.n_layers_per_block = n_layers_per_block
+        self.first_filter_length = first_filter_length
+        self.nonlinearity = nonlinearity
+        self.split_first_layer = split_first_layer
+        self.batch_norm_alpha = batch_norm_alpha
+        self.batch_norm_epsilon = batch_norm_epsilon
+        self.conv_weight_init_fn = conv_weight_init_fn
 
         if self.split_first_layer:
-            self.add_module('dimshuffle', Expression(_transpose_time_to_spat))
+            self.add_module('dimshuffle', Expression(transpose_time_to_spat))
             self.add_module('conv_time', nn.Conv2d(1, self.n_first_filters,
                                                    (self.first_filter_length, 1),
                                                    stride=1,
@@ -146,7 +158,7 @@ class EEGResNet(nn.Sequential):
                         nn.Conv2d(n_cur_filters, self.n_classes,
                                   (1, 1), bias=True))
         self.add_module('softmax', nn.LogSoftmax(dim=1))
-        self.add_module('squeeze', Expression(_squeeze_final_output))
+        self.add_module('squeeze', Expression(squeeze_final_output))
 
         # Initialize all weights
         self.apply(lambda module: _weights_init(module, self.conv_weight_init_fn))
@@ -167,22 +179,6 @@ def _weights_init(module, conv_weight_init_fn):
     elif 'BatchNorm' in classname:
         init.constant_(module.weight, 1)
         init.constant_(module.bias, 0)
-
-
-def _squeeze_final_output(x):
-    """
-    remove empty dim at end and potentially remove empty time dim
-    do not just use squeeze as we never want to remove first dim
-    """
-    assert x.size()[3] == 1
-    x = x[:, :, :, 0]
-    if x.size()[2] == 1:
-        x = x[:, :, 0]
-    return x
-
-
-def _transpose_time_to_spat(x):
-    return x.permute(0, 3, 2, 1)
 
 
 class _ResidualBlock(nn.Module):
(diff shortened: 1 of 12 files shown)


agramfort:
I would at least call this class BaseSequentialModel

I think you need to improve test to check for strides and dilations

Also I would make sure that you don’t now enforce that all models to be used in a training loop in braindecode derive from this class. We have recurrent nets or custom
Model that is not sequential

sliwy:
Good point, I didn't see the problem with other models. One possible solution is to make this class inherit from `torch.nn.Module` so it should make easy to work with other non-`Sequential` models. In case of `Sequentials`, we can create a class inheriting from this `Model` and `Sequential`, but it makes structure of the project more complicated. What do you think?

agramfort:
I would use a mixin class to add these methods to certain models but keep direct inheritance from torch classes

robintibor:
Yeah I think we should be aware there are quite strong assumptions that are hard to verify for the to_dense_prediction_model ... at the time of writing, it was meant for me not to duplicate all the code for shallow/deep. So another option may be to have two new model classes CausalDensePredShallow and CausalDensePredDeep (which are subclasses of shallow/deep) and call the to_dense_prediction_model on init. Instead of subclasses, might also be another flag to constructor.. 
And then it would be more onto us, when we integrate new models to see if there are also a non-causal/densepred one and a causal densepred one... To clarify, what these terms mean, I explained once here: https://github.com/braindecode/braindecode/pull/32#issuecomment-576441387 

> This is possible in the following way for a "dense-prediction" "timestep-causal" network. Dense-prediction here means from a certain timestep on, you have predictions for every timestep until the last timestep in the supercrop, without gaps inbetween. Timestep-causal means a prediction for a specific timestep only uses data up until that timestep, for example the second-last prediction in the supercrop is only computed from the data at the timesteps until the second last timestep, and is not affected by the data at the last timestep.

I think in general, even for convolutional models, I see at least these things that could happen:
* the model fits our automatic conversion code to a dense-prediction timestep-causal network.
* the model is already a dense-prediction timestep-causal network.
* the model is automatically transformable into a dense-prediction timestep-causal network, but would need different code
* the model is not easily automatically transformable into a dense-prediction timestep-causal network

I think, in the end what may be necessary is to have some property in each model which says if that model is a dense-prediction timestep-causal network, so that subsequent functions can give warnings if we rely on that assumption.

robintibor:
Maybe easiest would be to add flag to constructor of shallow/deep and call the to dense on init? It is probably necessary for any other model to check how to transform it/if this function does something correct.  How do you think @sliwy ?

robintibor:
and the output shape, I think maybe just make it into a function as you did, but not add to some base class? I think it's quite nice atm that models are just regular pytorch modules atm?

sliwy:
@robintibor Do you think that all models implemented in braindecode should be convertible to dense-prediction causal-timestep model? If not maybe this `to_dense` procedure should be specific for those two models and then in the init we should specify whether the model should be dense or not as you mentioned in your comment. There will be no need to call `to_dense` method from outside the model. Am I right? If yes it sounds good to me. One thing I don't like (but maybe I am too sensitive) is modifying the class using a function - it looks weird to me and I would opt for creating a method for this which make it clear that this is a part of model implementation.

I agree that adding a new class from which models need to inherit will introduce more complexity and maybe it is not the best option but I like the idea of mixin as it fits there quite well and create a self-contained object of the model without using additional functions placed in other files. The biggest question is if the `to_dense`, `output_shape`, `transpose_time_to_spat`, and `squeeze_output` (two last are more generic not created only for cropped decoding) are methods that should be easily accessible by any model? This depends on whether all models should be convertible to dense-prediction.

Including everything connected to a model inside a class makes the code cleaner in my opinion as you can easily see what methods I can use with an object. One more thing I see, there will be no problem in not using this class as skorch is PyTorch compatible. Our class (this is my vision) would make life easier in case of handling EEG signal, for example, a method `transpose_time_to_spat`. Will we have more things like this in the future?

robintibor:
>If not maybe this to_dense procedure should be specific for those two models and then in the init we should specify whether the model should be dense or not as you mentioned in your comment. There will be no need to call to_dense method from outside the model. Am I right? If yes it sounds good to me
Yes, exactly. As written above we don't assume this function will work on all models.

I think why I am a bit hesistant to introduce some base class for all models is that I want users to easily include their own models. And introducing some base class that new models have to inherit from seems to me to make it look as if there is some additional constraints the models have to follow or some additional logic, while there isn't.

agramfort:
>
> I think why I am a bit hesistant to introduce some base class for all
> models is that I want users to easily include their own models. And
> introducing some base class that new models have to inherit from seems to
> me to make it look as if there is some additional constraints the models
> have to follow or some additional logic, while there isn't.
>

I agree

sliwy:
I removed all parts connected to base model class. Now it introduces only changes that put common code used by more than one model in the same file. I created also `get_output_shape` function.

braindecode/models/eegresnet.py
```diff
@@ -31,14 +33,24 @@ def __init__(self, in_chans,
...
-        del self.self
+        self.final_pool_length = final_pool_length
+        self.n_first_filters = n_first_filters
+        self.n_layers_per_block = n_layers_per_block
+        self.first_filter_length = first_filter_length
+        self.nonlinearity = nonlinearity
+        self.split_first_layer = split_first_layer
+        self.batch_norm_alpha = batch_norm_alpha
+        self.batch_norm_epsilon = batch_norm_epsilon
+        self.conv_weight_init_fn = conv_weight_init_fn
```
robintibor: Makes me cry but ok ;)

agramfort: using locals() is really brutal and it could pull some crap in the self.__dict__
one option is to inspect the __init__ and set all parameters passed that we could detect with inspect module.

something like:

```
import inspect
init_signature = inspect.signature(self.__init__)
```

robintibor: Yeah that maybe possible like as separate function. But maybe outside this PR then.
What is your view on data classes? https://docs.python.org/3/library/dataclasses.html

robintibor: Btw some small discussion on this also here: https://stackoverflow.com/a/12191200/1469195

robintibor: Created new issue for this https://github.com/braindecode/braindecode/issues/93

agramfort:
>
> Yeah that maybe possible like as separate function. But maybe outside this PR then.
>
> What is your view on data classes?
> https://docs.python.org/3/library/dataclasses.html
>

why not but it would bump the dependency to python >= 3.7

robintibor:
yes, but don't feel this is a problem to me.  Feel more it is strange to call a model a "dataclass".

agramfort:
I am always reluctant to require blending edge versions (3.8 is not
functional for me yet due to vtk for example)

>

PR #85, opened 2020-02-17
Removed deprecated transformers

gemeinl:
-removed deprecated tranformers following scikit-learn API
-added new zscore transform following current API
-added tests for zscore transform on continuous data
-minor improvements to windowers.py

### DIFF
--- a/braindecode/datautil/transforms.py
+++ b/braindecode/datautil/transforms.py
@@ -12,8 +12,6 @@ ToDo: should transformer also transform y (e.g. cutting continuous labelled
 
 from collections import OrderedDict
 
-import mne
-
 
 def transform_concat_ds(concat_ds, transforms):
     """Apply a number of transformers to a concat dataset.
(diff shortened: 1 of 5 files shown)


braindecode/datautil/transforms.py
```diff
@@ -77,103 +75,20 @@ def _transform(raw_or_epochs, transforms):
...
-            None, sfreq, l_freq=l_freq, h_freq=h_freq, method="fir", **kwargs
-        )
-        self.overlap_kwargs = overlap_kwargs
-
-    def __call__(self, windows):
-        return mne.filter._overlap_add_filter(
-            windows, self.filter, **self.overlap_kwargs
-        )
-
+    continuous_data: np.ndarray
```
hubertjb: For completeness, should we add the expected dimensions (n_channels, n_times)?

braindecode/datautil/transforms.py
```diff
@@ -77,103 +75,20 @@ def _transform(raw_or_epochs, transforms):
...
-            normalized raw data
-
-        """
-        raw = raw.apply_function(lambda x: x - x.mean())
-        return raw.apply_function(lambda x: x / x.std())
-
-
-class FilterWindow(object):
-    """FIR filter for windowed data.
+def zscore(continuous_data):
```
hubertjb: I think this function should also work with windowed data, we might just need to use `keepdims=True`. If that's desirable, then the argument could be renamed to `data`.

test/unit_tests/datautil/test_transforms.py
```diff
@@ -71,3 +70,19 @@ def test_transform_windows_method(windows_concat_ds):
     raw_window = windows_concat_ds[0][0]
     transform_concat_ds(windows_concat_ds, transforms)
     assert not np.array_equal(raw_window, windows_concat_ds[0][0])
+
+
+def test_zscore_base_concat_ds(base_concat_ds):
```
hubertjb: If `zscore` is changed above to also accept windowed data, we should add another test here for the windowed case.

braindecode/datautil/transforms.py
gemeinl: Yes, good idea. So transforms should work on numpy ndarrays. For application to mne.Raw we pass them to <code>raw.apply_function()</code>. For application to mne.Epochs there is no such function. (Could mne implement such a function?) We would then implement a way to get the ndarrays from the Epochs object, transform it, and put it pack. This will likely take place inside the Dataset class as in torchvision.

braindecode/datautil/transforms.py
gemeinl: With the changes discussed above it could then also ne (n_windows x n_channels x n_times). I will add this.

PR #86, opened 2020-02-17
WIP aligning README.rst to index.rst and vice versa

gemeinl:
@agramfort What would be the best way to keep both README.rst and index.rst up to date?


agramfort:
I would avoid duplication of content that is likely to change and point to
the one you want to keep so
readme pointing to web doc or the opposite.

>

robintibor:
for now we stick with keeping in sync manually.

PR #90, opened 2020-02-28
Full examples bcic iv 2a

robintibor:
Examples for BCIC IV 2a with good accuracies.

So, `plot_bcic_iv_2a_moabb_trial.py` and `plot_bcic_iv_2a_moabb_cropped.py`
should replace:
* plot_bcic_iv_2a_moabb.py
* bcic_iv_2a_cropped.py
* bcic_iv_2a_trial.py

Here are results from 3 seeds, 800 epochs for cropped, 1600 epochs for trial, averaged over the 9 subjects. This cosine-annealing-train-on-full-train setup outperforms our previous early-stopping setup as is expected, previous setup was also fairly bad results in that run, don't know why unfortunately. The accuracies for a cropped deep on 0 Hz are a bit suspiciously high ? Or is this in the range you see @sbbrandt ? Trialwise is very bad, indicating the hyperparameters (learning rate, weight decay, n_epochs) are not well-suited for trialwise decoding, at least on this dataset.

skorch_braindecode_bcic_iv_2a

PR #91, opened 2020-03-01
Callbacks improvements

sliwy:
Improves callbacks definiton, so now a user don't need to specify all callbacks using skorch. It is enough to pass scoring functions names from sklearn and `EEGClassifier` will create callbacks internally, as described in #48 
I also added `EEGClassifier` to be displayed in our docs. Docstring needs improvement definitely, (I will create an issue) as for now it does not show docstring of base class.

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -46,31 +46,31 @@ class EEGClassifier(NeuralNetClassifier):
     def _parse_callbacks(self, callbacks):
         callbacks_list = []
         if callbacks is not None:
-            for c in callbacks:
-                if isinstance(c, tuple):
-                    callbacks_list.append(c)
+            for callback in callbacks:
+                if isinstance(callback, tuple):
+                    callbacks_list.append(callback)
                 else:
-                    assert isinstance(c, str)
-                    scoring = get_scorer(c)
+                    assert isinstance(callback, str)
+                    scoring = get_scorer(callback)
                     scoring_name = scoring._score_func.__name__
                     lower_is_better = False if scoring_name.endswith('_score') else True
-                    train_name = f'train_{c}'
-                    valid_name = f'valid_{c}'
+                    train_name = f'train_{callback}'
+                    valid_name = f'valid_{callback}'
                     if self.cropped:
                         # In case of cropped decoding we are using braindecode
                         # specific scoring created for cropped decoding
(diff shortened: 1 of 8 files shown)


docs/api.rst
```diff
@@ -20,6 +20,18 @@ This is the reference for classes (``CamelCase`` names) and functions
    :no-members:
    :no-inherited-members:
 
+Classifier
+======
```
agramfort:
```suggestion
==========
```

we should see if CIs can fail if there are sphinx formatting errors

agramfort:
are all the examples in https://288-232335424-gh.circle-artifacts.com/0/dev/auto_examples/index.html still relevant?
It would be nice to have matplotlib plot something so the thumbnails are appealing in this page.

braindecode/classifier.py
```diff
@@ -1,16 +1,83 @@
...
+        super().__init__(*args, callbacks=callbacks, **kwargs)
+
+    def _parse_callbacks(self, callbacks):
+        callbacks_list = []
+        if callbacks is not None:
+            for callback in callbacks:
+                if isinstance(callback, tuple):
+                    callbacks_list.append(callback)
+                else:
+                    assert isinstance(callback, str)
```
robintibor: Can you change to 
```suggestion
                    assert isinstance(callback, str) or callable(callback)
```
This is also what get_scorer expects from scikit learn https://scikit-learn.org/stable/modules/generated/sklearn.metrics.get_scorer.html

robintibor: Ah ok I see you might then have to also change line below with _score_func.name

robintibor: Hm Ok I see you use assumption you have a string quite often below... Then ignore my comment, maybe for a future time I leave it here

braindecode/scoring.py
```diff
@@ -124,9 +124,7 @@ def __init__(
         name=None,
         target_extractor=to_numpy,
         use_caching=True,
-        input_time_length=None,
```
robintibor: ah great catch

robintibor:
> are all the examples in https://288-232335424-gh.circle-artifacts.com/0/dev/auto_examples/index.html still relevant?
> It would be nice to have matplotlib plot something so the thumbnails are appealing in this page.

Some will go, as written in #90 
But ya then maybe worthwhile to plot something when cleaning up the example files

PR #95, opened 2020-03-18
predefined split for experiments

robintibor:
A small separate PR just to remove the local splitting class for splits, instead using a function from skorch. How do you feel about this change @sliwy

### DIFF
--- a/examples/plot_bcic_iv_2a_moabb_cropped.py
+++ b/examples/plot_bcic_iv_2a_moabb_cropped.py
@@ -18,6 +18,7 @@ import pandas as pd
 import matplotlib.pyplot as plt
 from matplotlib.lines import Line2D
 from skorch.callbacks import LRScheduler
+from skorch.helper import predefined_split
 mne.set_log_level('ERROR')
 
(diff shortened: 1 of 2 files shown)


examples/plot_bcic_iv_2a_moabb_cropped.py
```diff
@@ -107,19 +108,16 @@
...
 
-
-class TrainTestBCICIV2aSplit(object):
-    def __call__(self, dataset, y, **kwargs):
-        splitted = dataset.split('session')
-        return splitted['session_T'], splitted['session_E']
-
+splitted = windows_dataset.split('session')
+train_set = splitted['session_T']
+test_set = splitted['session_E']
```
sliwy: maybe better would be `valid_set` as it is used by skorch for validation? For testing, we can think about different dataset?

sliwy:
The only thing I would change is the `test_set` name (I've made a comment). I like the change! :)

robintibor:
Super thanks @sliwy  changed as requested, ok merging

PR #96, opened 2020-03-18
removed monitors.py

robintibor:
Cleanup, removing unused code, adding tests, improving naming


test/acceptance_tests/test_cropped_decoding.py
```diff
@@ -14,12 +14,111 @@
 
 from braindecode.datautil.iterators import CropsFromTrialsIterator
 from braindecode.models.util import to_dense_prediction_model
-from braindecode.monitors import compute_preds_per_trial_from_crops
 from braindecode.models import ShallowFBCSPNet
 from braindecode.util import set_random_seeds
 from braindecode.util import var_to_np, np_to_var
 
 
+def compute_preds_per_trial_from_crops(all_preds, input_time_length, X):
```
robintibor: Actually whole test should be refactored to new dataset etc. classes and then use existing functions from scoring, if not possible, refactor scoring to make the functionality accessible (for the functions now copied inside from old monitor.py). Wanna have a try @sliwy ?

agramfort: this assumes a classification task right? what this code does is basically a voting classifier?
also make sure any function that is not meant to be used directly and public is preceded with a _

robintibor: This is classification task, the code sums up log probabilities per timestep for each class, so same as multiplication of probabilities per timestep for each to compute probability of trial for each class.

robintibor: Made the underscore change

agramfort:
we should not overfit classif problems or if we need code only for classif
it should be explicit in the function or module name

>

robintibor:
The functions now inside are expected to be removed, that was what I meant with my comment to @sliwy  regarding refactoring this. They are only used inside this test atm.
For further clarification, the per-class-sum described in my comment refers to a code inside the example: 
```python
 np.array(
                [np.mean(p, axis=1) for p in preds_per_trial]
            )
```
And the current to-be-removed `_compute_preds_per_trial_from_crops` is more generic, it collects all the predictions that belong to each trial and removes duplicate predictions. Similar functionality already exists in `scoring.py` now and should be used here inside this test.

agramfort:
ok great !

>

robintibor:
Concretely, function that should be used now (also inside this test): 
https://github.com/braindecode/braindecode/blob/c9df8f987da3241631f4a31a7e2d1f8fee29e1c1/braindecode/scoring.py#L19-L83
Example usage of this function here:
https://github.com/braindecode/braindecode/blob/c9df8f987da3241631f4a31a7e2d1f8fee29e1c1/braindecode/scoring.py#L156-L184

This function `trial_preds_from_supercrop_preds` also prime candidate to be the next to have a proper test for :)

test/acceptance_tests/test_cropped_decoding.py
sliwy: I've opened an issue #97 for this.

robintibor: siüer!

Issue #97, opened 2020-03-22
Tests refactor/cleanup


sliwy:
`test_cropped_decoding.py`, `test_trialwise_decoding.py` refactor to use new dataset classes

robintibor:
If this is done can you close @sliwy ?

PR #98, opened 2020-03-30
trialwise and cropped test decoding rewritten to use new model API, EpochsDataset…

sliwy:
#97 
I've tried to rewrite the `test/acceptance_tests/test_trialwise_decoding.py`. It does not reproduce the same results as the old test. I couldn't make skorch give the same results (even if I took exactly the same net and made batches sampling the same) as the previous code. Below there is a comparison of results, it is probably not really significant as the dataset contains only 60 observations, so I don't know if we should care a lot.
image
Epochs on x-axis.
 @robintibor let me know if you agree that current flow does not have to reproduce previous results.

One more thing is that I wanted to use the same dataset as in the previous test, so I had to read raw files with mne, then create epochs and pass it to the fit method. To do this I created a small dataset class implementing sampling from `mne.Epochs` object. `mne.Epochs` object as it is the simplest way to integrate with any different dataset that is now not implemented in braindecode. Code of the class is below (If you think it may be helpful we can add it to braindecode Dataset):
```python
import numpy as np
from torch.utils.data import Dataset
class EpochsDataset(Dataset):
    def __init__(self, windows):
        self.windows = windows
        self.y = np.array(self.windows.events[:, -1])
        self.y = self.y - self.y.min()

    def __getitem__(self, index):
        X = self.windows.get_data(item=index)[0].astype('float32')[:, :, None]
        y = self.y[index]
        return X, y

    def __len__(self):
        return len(self.windows.events)
````
Next thing will be to change reading data to use braindecode datasets but for now I can't do this - I need to spend more time on it as I get some errors. When I confirm that it's not my fault I will let you know where the problem is.

### DIFF
--- a/test/acceptance_tests/test_cropped_decoding.py
+++ b/test/acceptance_tests/test_cropped_decoding.py
@@ -5,13 +5,14 @@
 
 import mne
 import numpy as np
+import torch
 from mne.io import concatenate_raws
 from skorch.helper import predefined_split
 from torch import optim
 
 from braindecode import EEGClassifier
 from braindecode.datasets.croppedxy import CroppedXyDataset
-from braindecode.losses import CroppedNLLLoss
+from braindecode.losses import CroppedLoss
 from braindecode.models import ShallowFBCSPNet
 from braindecode.models.util import to_dense_prediction_model, get_output_shape
 from braindecode.util import set_random_seeds
@@ -98,7 +99,8 @@ def test_cropped_decoding():
     clf = EEGClassifier(
         model,
         cropped=True,
-        criterion=CroppedNLLLoss,
+        criterion=CroppedLoss,
+        criterion__loss_function=torch.nn.functional.nll_loss,
         optimizer=optim.Adam,
         train_split=train_split,
         batch_size=32,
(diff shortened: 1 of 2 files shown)


robintibor:
I will have a closer look later, but to answer your question: Yes, new code does not have to reproduce numbers of old code, that's fine. No need to make them reproduce exactly from my view here. Too many random generators used differently to somehow set to same state etc.

Iw as also wondering about EpochsDataset, probably would be worthwhile to also have some function (outside of test), that takes X and y, wraps into mne and produces this kind of EpochsDataset.. so that people using other non-mne loading code have a clear way to enter their data. Such functionality could then be moved outside test, directly into main source

sliwy:
It should be also possible to not use mne directly and `EpochsDataset`, instead we can use `MOABBDataset` class but I have some issues with this class, so I need to spend more time on it.

robintibor:
hm ya also true, this is maybe a bit different points, what we want to use here in the test and the second task of having a converter x,y -> windowsdataset... could be a separate PR also... for now also fine if you just change the expected values in the test so that tests pass

sliwy:
I changed also `test_cropped_decoding.py` to use `EEGClassifier`. Below the same plots as for `test_trialwise_decoding.py`. @robintibor do you think that I should modify something more?

image

robintibor:
so now this is also mergeable right?

sliwy:
rebased and ready to merge I think

robintibor:
great

PR #99, opened 2020-04-06
WIP: Support for regression tasks 

gemeinl:
Braindecode can now be used to perform regression tasks!
Concretely, regression here refers to the same setting as in classification (with a single target per trial/window/crop), however, instead of integer targets, we have real valued targets.


gemeinl:
Some questions came to my mind during working on this:

- EEGClassifier and EEGRegressor now have a lot of duplicate code. How do we handle this?
- For my application of regression, I need to preprocess the targets (scaling). One could handle this by accessing the <code>dataset.description</code> DataFrame and overwriting the targets. One could also think of a logic using <code>transform_concat_ds</code> for target scaling.

Any thoughts on this?

gemeinl:
@sliwy please have a look specifically at line https://github.com/braindecode/braindecode/pull/99/commits/b06fad6e3d96804ccec8b224cbf7899e61c64d4a#diff-66bbe0ace97b4f253147b1d0d9fa5783R54, where I set <code>lower_is_better=False</code> for all regression tasks. This is due to the fact, that scoring functions in sklearn are given as "neg_mean_squared_error" or "neg_mean_absolute_error"... (https://scikit-learn.org/stable/modules/model_evaluation.html)

gemeinl:
To be able to perform regression, the (log)softmax layer is manually removed (https://github.com/braindecode/braindecode/pull/99/commits/72c9d66a0643651cbc8beb734a0a16f7712cdc97#diff-32f3bf89c00c0a90a326f9a2e6fab880R65-R70). Should this be added as a utility function?

braindecode/__init__.py
```diff
@@ -1,3 +1,4 @@
 from .version import __version__
 
 from .classifier import EEGClassifier
+from .regressor import EEGRegressor
```
sliwy, line 4: You should add also an entry to `docs/api.rst` file to generate documentation for this new class.

gemeinl: Thanks for catching this. Will add to api.rst.

braindecode/regressor.py
```diff
@@ -0,0 +1,194 @@
...
+                x,y,i = batch
+                # Store for scoring callbacks
+                self.net._last_supercrop_inds = i
+            else:
+                x,y = batch
+
+            # TODO: should be on dataset side
+            if hasattr(x, 'type'):
+                x = x.type(torch.float32)
+                y = y.type(torch.float32)
```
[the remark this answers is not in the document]
gemeinl: Yes, it could definitely be merged. As also the 'TODO' message in the code says, it would be favorable that the datasets already give data of the appropriate type. This would remove the necessity of casting in the <code>ThrowAwayIndexLoader</code>.  This would mean that the datasets need to know which task ithey are being used for.

examples/plot_tuh_age_regression.py
```diff
@@ -0,0 +1,155 @@
...
+        n_filters_4=int(n_chans * (2 ** 3.0)),
+        final_conv_length=1,
+    )
+    optimizer_lr = 0.01
+    optimizer_weight_decay = 0.0005
+else:
+    raise ValueError(f'{model_name} unknown')
+
+new_model = torch.nn.Sequential()
+for name, module_ in model.named_children():
```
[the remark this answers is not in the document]
gemeinl: I guess this is something we need to discuss. They way it is written here is how I did it so far. It is not nice, I agree. I also thought about a function coverting EEGClassifier to EEGRegressor and vice versa. However, this conversion might not work on custom models used with braindeocde.. What do you think @robintibor ?

examples/plot_tuh_age_regression.py
```diff
@@ -0,0 +1,155 @@
...
+    ('pick_channels', dict(ch_names=ch_names)),
+    ('reorder_channels', dict(ch_names=ch_names)),
+    ('apply_function', dict(fun=lambda x: x * factor, channel_wise=False)),
+    ('crop', dict(tmin=tmin, tmax=tmax, include_tmax=False)),
+    ('resample', dict(sfreq=sfreq)),
+    ('apply_function', dict(fun=clip, clipping_value=clipping_value))
+]
+transform_concat_ds(dataset, raw_transform_dict)
+
+sfreqs = [ds.raw.info['sfreq'] for ds in dataset.datasets]
```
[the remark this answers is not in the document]
gemeinl: It can be removed from the example, especially since a resampling is performed in the line directly above. However, the TUH Abnormal EEG Dataset contains raw recordings with varying sampling frequencies. To me, it is not obvious where to enforce / assure same sampling frequency for all data being used...

braindecode/classifier.py
```diff
@@ -53,7 +53,11 @@ def _parse_callbacks(self, callbacks):
                     assert isinstance(callback, str)
                     scoring = get_scorer(callback)
                     scoring_name = scoring._score_func.__name__
-                    lower_is_better = False if scoring_name.endswith('_score') else True
+                    if (scoring_name.endswith('_score') or
+                        callback.startswith('neg_')):
```
sliwy, line 57: @gemeinl great you added this `neg_` check. I missed it.

examples/plot_tuh_age_regression.py
```diff
@@ -0,0 +1,152 @@
...
+model = new_model
+
+if cuda:
+    model.cuda()
+
+to_dense_prediction_model(model)
+n_preds_per_input = get_output_shape(model, n_chans, input_time_length)[2]
+
+dataset = TUHAbnormal(
+    path="/data/schirrmr/gemeinl/tuh-abnormal-eeg/raw/v2.0.0/edf/train/",
```
sliwy, line 84: I see that this path does not exist in the CI, so that's why your PR is failing, any idea how to fix this?

gemeinl: Well, a solution would be to not use the TUH dataset. For the purpose of this example I could fake data using <code>braindecode.util.create_mne_dummy_raw</code>.

braindecode/losses.py
```diff
@@ -1,10 +1,14 @@
...
+    """Compute Loss after averaging predictions across time.
     Assumes predictions are in shape:
     n_batch size x n_classes x n_predictions (in time)"""
 
+    def __init__(self, loss_function):
+        self.loss_function = loss_function
+
     def __call__(self, preds, targets):
-        return torch.nn.functional.nll_loss(torch.mean(preds, dim=2), targets)
+        avg_preds = torch.mean(preds, dim=2).squeeze()
```
robintibor, line 13: what is purpose of squeeze here? torch squeeze by default anyways dim 2 right? and if you have batch size 1 this will lead to error also right? I think in general squeeze should always be called with dimension argument to avoid errors in any case

braindecode/regressor.py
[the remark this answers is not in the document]
robintibor: For now would agree with @sliwy  solution to introduce a flag and just have one class

examples/plot_tuh_age_regression.py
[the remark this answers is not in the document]
robintibor: atm I am more in favour of keeping as is, as it is more flexible, does not require new models to like have missing last layer or sth. For our own models we could create base class (deep4) and subclasses (deep4regressor/deep4classifier) if we feel this is nice... another option would be to remove softmax layer altogether, and make it part of loss function ... but to me not so crucial atm

braindecode/losses.py
gemeinl: So far, code was written to perform multi-class classification. Predictions here are given as <code>batch_size x n_classes</code>. For regression, <code>n_classes = 1</code> and I get a UserWarning of incompatible sahpes when computing <code>mse_loss</code>. Therefore, I want to squeeze empty first dimension. Is there a better way to do this?

braindecode/regressor.py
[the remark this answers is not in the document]
gemeinl: I'll introduce a flag <code>is_regression</code> to the <code>ThrowAwayIndexLoader</code>. I will then move this class to <code>braindecode.util</code> and import it in <code>classifier.py</code> and <code>regressor.py</code>, OK?

braindecode/losses.py
sliwy: How about solving this issue in a way that multi-target regression works too? In my case it may be useful as I have to predict x, y, z coordinates. Do you think it is possible?

gemeinl: If I understand correctly, in your setting predictions would be in shape <code>batch_size x 3 x n_times</code> and targets in shape <code>batch_size x 3</code>. For <code>mse_loss</code> predictions and targets should have the same dimensionality. So this should work with <code>squeeze(dim=1)</code>, right?

sliwy: @gemeinl yeah, it should work. I checked and in this setup `torch.mean` already returns tensor of shape `batch_size x 3`, so the `squeeze` operation is not needed. Is it added for other cases?

examples/plot_tuh_age_regression.py
sliwy: Don't we want to have nice plots and make an example of real data usage? If we use dummy input it is no longer a real example. I don't know what is the best solution, but possible from my perspective:
1. We use another smaller dataset on which we can perform regression (TUH dataset is too big to use it on CI? Am I right?)
2. We use dummy input and treat this as a not real-life example (other examples uses only small datasets, so they are not real-life too?).
3. We add functionality to TUHDataset to use only a subset of data for TUH dataset, so it works on CI.

gemeinl: Yes you are right. It would then not be a real example anymore, just an example that shows how to generally do it.
1. I guess the main problem is, that although the TUH dataset is publicly available, it requires for registration. I think, we cannot reupload files without asking for permission. Maybe moabb could ask for integration in their data fetching.
2. Yes, that would be easiest for now.
3. We have this subset reading functionality already (see <code>subject_ids</code> https://github.com/braindecode/braindecode/blob/master/braindecode/datasets/datasets.py#L125-L126).

braindecode/losses.py
gemeinl: Yes, we only need squeezing when <code>n_classes=1</code>.

robintibor: yeah so for sure add dim to squeeze. then should be fine and extensible i guess

examples/plot_tuh_age_regression.py
robintibor: yeah to me dummy for now seems best. I think making a real example out of it is sth that can be done at any point independently of other work

robintibor:
@sliwy me and lukas don't understand why we now have this long list of commits in this PR, lukas tried the rebase logic. Is it fine to just merge this PR now or would be there be future problems?

sliwy:
@robintibor I do not know if it can cause problems in the future but I investigated it a little (I do not know why it happened, I think that you might have pushed from different machines/versions of the branch and merged it in the end? almost every commit occurs in the branch 3 times). 
Just to be sure everything is ok, I have cleaned the situation on my branch [age_regression](https://github.com/sliwy/braindecode/tree/age_regression), so when creating the PR it only adds new changes. @gemeinl please take a look at #107 and check if everything is in the PR and I haven't removed any part of your work. Are you ok with merging from my PR?

Just small info, this is how `age_regression` branch looked like in `gitk`, maybe you will figure out why it happened:
image

PR #100, opened 2020-04-06
Added pillow to docs requirements andchange mne to install from pypi

sliwy:
I think it should solve failing CI as it shows lack of Pillow package

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -2,3 +2,4 @@ sphinx_gallery
 sphinx_rtd_theme
 numpydoc
 memory_profiler
+pillow
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,5 +3,5 @@ pandas
 scipy
 matplotlib
 h5py
-git+https://github.com/mne-tools/mne-python.git
+mne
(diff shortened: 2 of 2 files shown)


sliwy:
One more thing I had to change to make it green was the version of mne. As I see mne version on pypi was bumped and now 0.20.0 is the default version, so we do not need to install mne from github. There were some changes in mne on the newest version that resulted in moabb crash.

agramfort:
thx @sliwy

robintibor:
> One more thing I had to change to make it green was the version of mne. As I see mne version on pypi was bumped and now 0.20.0 is the default version, so we do not need to install mne from github. There were some changes in mne on the newest version that resulted in moabb crash.

@sylvchev can you help with this?

agramfort:
It's an issue that mne killed deprecated features from 0.20 already in master.

moabb should now be adjusted to avoid using deprecated features from 0.20

mne master is now 0.21.dev

braindecode should stick to 0.20 stable for now

sylvchev:
I'll check this on MOABB and keep you posted.

PR #102, opened 2020-04-20
Windowers events bug fix

sliwy:
I wanted to use `PhysionetMI` dataset and I found a bug in `create_windows_from_events` I tried to fix it, let me know if it's ok for you. 

1.  `create_windows_from_events` function assumed that all events appear in all datasets which may be not true. If `mapping` parameter was `None` and in the first dataset, there was only event for example `T0`, in the next datasets the only event `T0` will be used. I expected that `mapping==None` reads all the events. Am I right?
2. In case any event is filtered out with the `mapping` parameter, the code crashed because it may happen that in some datasets there are not events of all types. [Line121](https://github.com/braindecode/braindecode/compare/master...sliwy:windowers_bug?expand=1#diff-c031004527cfc6287fbb1741b7f3e1ddR121) Now I am using `events_id` returned by `mne.events_from_annotations(ds.raw, mapping)` which contains only events that can be found in the dataset.
3. In case `mapping` is filtering out events it also crashed in the line [Line89](https://github.com/braindecode/braindecode/compare/master...sliwy:windowers_bug?expand=1#diff-c031004527cfc6287fbb1741b7f3e1ddR89) as `annotations` have all events and `onsets` only ones that were in the `mapping`. That is why I am introducing `filtered_durations` in [Line86-Line88](https://github.com/braindecode/braindecode/compare/master...sliwy:windowers_bug?expand=1#diff-c031004527cfc6287fbb1741b7f3e1ddR86-R88).

Example code that did not work:
```python
from braindecode.datautil import create_fixed_length_windows, create_windows_from_events
ds = MOABBDataset('PhysionetMI', [1])
windows = create_windows_from_events(ds, 160, 70, 1, 10, False, mapping={'T0': 0, 'T1': 1})
windows = create_windows_from_events(ds, 160, 70, 1, 10, False)
```

### DIFF
--- a/braindecode/datautil/windowers.py
+++ b/braindecode/datautil/windowers.py
@@ -52,7 +52,7 @@ def create_windows_from_events(
         if True, preload the data of the Epochs objects.
     drop_bad_windows: bool
         If True, call `.drop_bad()` on the resulting mne.Epochs object. This
-        step allows identifiying e.g., windows that fall outside of the
+        step allows identifying e.g., windows that fall outside of the
         continuous recording. It is suggested to run this step here as otherwise
         the BaseConcatDataset has to be updated as well.
 
@@ -65,17 +65,29 @@ def create_windows_from_events(
         trial_start_offset_samples, trial_stop_offset_samples,
         supercrop_size_samples, supercrop_stride_samples)
 
+    # If user did not specify mapping, we extract all events from all datasets
+    # and map them to increasing integers starting from 0
+    infer_mapping = mapping is None
+    if infer_mapping:
+        mapping = {}
+
     list_of_windows_ds = []
     for ds in concat_ds.datasets:
-        if mapping is None:
+        if infer_mapping:
+            unique_events = np.unique(ds.raw.annotations.description)
+            new_unique_events = [x for x in unique_events if x not in mapping]
             # mapping event descriptions to integers from 0 on
-            mapping = {v: k for k, v in enumerate(
-                np.unique(ds.raw.annotations.description))}
+            max_id_mapping = len(mapping)
+            mapping.update(
+                {v: k + max_id_mapping for k, v in enumerate(new_unique_events)}
+            )
 
-        events, _ = mne.events_from_annotations(ds.raw, mapping)
+        events, events_id = mne.events_from_annotations(ds.raw, mapping)
         onsets = events[:, 0]
-        stops = onsets + (ds.raw.annotations.duration
-                          * ds.raw.info['sfreq']).astype(int)
+        filtered_durations = np.array(
+            [a['duration'] for a in ds.raw.annotations if a['description'] in events_id]
+        )
+        stops = onsets + (filtered_durations * ds.raw.info['sfreq']).astype(int)
 
         if stops[-1] + trial_stop_offset_samples > len(ds):
             raise ValueError('"trial_stop_offset_samples" too large. Stop of '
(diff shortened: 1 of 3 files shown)


braindecode/datautil/windowers.py
```diff
@@ -65,17 +65,28 @@ def create_windows_from_events(
...
+            if mapping is None:
+                mapping = {v: k for k, v in enumerate(unique_events)}
+            else:
+                max_id_mapping = max(list(mapping.values())) + 1
+                unique_events = [x for x in unique_events if x not in mapping]
+                mapping.update(
+                    {v: k + max_id_mapping for k, v in enumerate(unique_events)}
+                )
+
+        events, events_id = mne.events_from_annotations(ds.raw, mapping)
```
robintibor, line 84: I would refactor overall to:
```python
# If user did not specify mapping,
# we extract all events from all datasets
# and map them to increasing integers
# starting form 0
infer_mapping = mapping is None
if infer_mapping:
	mapping = {}

list_of_windows_ds = []
for ds in concat_ds.datasets:
    if infer_mapping:
	unique_events = np.unique(ds.raw.annotations.description)
        new_unique_events = [x for x in unique_events if x not in mapping]
        # mapping event descriptions to integers from 0 on
	if len(mapping) == 0:
		max_id_mapping = 0
	else:
	        max_id_mapping = max(list(mapping.values())) + 1
        mapping.update(
            {v: k + max_id_mapping for k, v in enumerate(new_unique_events)}
```
Found code a bit hard to read quickly before, like did not understand mapping will change for example

robintibor, line 84: I would refactor overall to:
```python
# If user did not specify mapping,
# we extract all events from all datasets
# and map them to increasing integers
# starting form 0
infer_mapping = mapping is None
if infer_mapping:
	mapping = {}

list_of_windows_ds = []
for ds in concat_ds.datasets:
    if infer_mapping:
	unique_events = np.unique(ds.raw.annotations.description)
        new_unique_events = [x for x in unique_events if x not in mapping]
        # mapping event descriptions to integers from 0 on
	if len(mapping) == 0:
		max_id_mapping = 0
	else:
	        max_id_mapping = max(list(mapping.values())) + 1
        mapping.update(
            {v: k + max_id_mapping for k, v in enumerate(new_unique_events)}
```
Found code a bit hard to read quickly before, like did not understand mapping will change for example

robintibor, line 84: I would refactor overall to:
```python
# If user did not specify mapping,
# we extract all events from all datasets
# and map them to increasing integers
# starting form 0
infer_mapping = mapping is None
if infer_mapping:
	mapping = {}

list_of_windows_ds = []
for ds in concat_ds.datasets:
    if infer_mapping:
	unique_events = np.unique(ds.raw.annotations.description)
        new_unique_events = [x for x in unique_events if x not in mapping]
        # mapping event descriptions to integers from 0 on
	if len(mapping) == 0:
		max_id_mapping = 0
	else:
	        max_id_mapping = max(list(mapping.values())) + 1
        mapping.update(
            {v: k + max_id_mapping for k, v in enumerate(new_unique_events)}
```
Found code a bit hard to read quickly before, like did not understand mapping will change for example

robintibor:
great catches, made some small suggestions

braindecode/datautil/windowers.py
```diff
@@ -65,17 +65,28 @@ def create_windows_from_events(
         trial_start_offset_samples, trial_stop_offset_samples,
         supercrop_size_samples, supercrop_stride_samples)
 
+    infer_mapping = True if mapping is None else False
```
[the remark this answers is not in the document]
sliwy: Yeah, thanks!

braindecode/datautil/windowers.py
sliwy: Great, thanks, I did those changes and I found one more simplification, I replaced
```python
	if len(mapping) == 0:
		max_id_mapping = 0
	else:
	        max_id_mapping = max(list(mapping.values())) + 1
```
with just simple `len(mapping)`

sliwy:
@robintibor I'll try to add something for this

robintibor:
super

Issue #57, opened 2020-01-21
Shuffle train iterator by default


robintibor:
Skorch does not do that, see https://github.com/skorch-dev/skorch/issues/419 , we should do it by default, much more intuitive.

PR #103, opened 2020-04-20
EEGClassifier default train dataset shuffle

sliwy:
Introduces shuffling the training dataset by default in `EEGClassifier`. It solves #57.
In case we merge it we have to modify accuracy and loss values in tests modified in #98.

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -33,15 +33,22 @@ class EEGClassifier(NeuralNetClassifier):
         callbacks (e.g., for the callback with name ``'print_log'``, use
         ``net.set_params(callbacks__print_log__keys_ignored=['epoch',
         'train_loss'])``).
+    iterator_train__shuffle: bool (default=True)
+        Defines whether train dataset will be shuffled. As skorch does not
+        shuffle the train dataset by default this one overwrites this option.
     """
     # TODO: Update docstring to use NeuralNetClassifier docstring with some
     #  imporvements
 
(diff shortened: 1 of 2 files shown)


braindecode/classifier.py
```diff
@@ -33,15 +33,21 @@ class name. Name conflicts are resolved by appending a count suffix
...
     """
     # TODO: Update docstring to use NeuralNetClassifier docstring with some
     #  imporvements
 
     def __init__(self, *args, cropped=False, callbacks=None, **kwargs):
         self.cropped = cropped
         callbacks = self._parse_callbacks(callbacks)
-
-        super().__init__(*args, callbacks=callbacks, **kwargs)
+        iterator_train__shuffle = kwargs.pop('iterator_train__shuffle', True)
```
robintibor, line 46: why not just put it into __init__ function signature? then also shows up in IDE etc. and then no need to do this kwargs.pop stuffs

robintibor, line 46: why not just put it into __init__ function signature? then also shows up in IDE etc. and then no need to do this kwargs.pop stuffs

sliwy: Yeah, sure, thanks for catching it, definitely it is a better solution. My mistake :)

robintibor:
so, should I merge https://github.com/braindecode/braindecode/pull/98 first? @sliwy  how it's convenient for you?

sliwy:
@robintibor I fixed tests, let's merge this first, then I will fix #98

robintibor:
ok merged

PR #105, opened 2020-04-20
Removed mne installation using newest github version

sliwy:
This PR fixes tests that are crashing in travis build (I changed `requiremnts.txt` mne version in one of the previous PR but it was defined also in conda `environment.yml`). I think that we have also a problem with travis build. Travis build is not done for each PRs, so we can merge PRs with failing tests. @agramfort do you have any idea why for a PR we have only circle CI done (it does not run tests)? Is it a problem with travis configuration?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -15,6 +15,6 @@ dependencies:
 - pytest
 - pip
 - skorch
+- mne
 - pip:
-  - https://github.com/mne-tools/mne-python/zipball/master
   - https://github.com/braindecode/braindecode/zipball/master


sliwy:
Example of PR with only 2 circle CI checks #101

sliwy:
Ok, I see, that this PR has triggered the travis build, so at this point, I do not understand why there are some PRs without this check.

agramfort:
these days I have seen travis fail to start on certain PRs. 

do you have a failing test using mne stable in pypi?

sliwy:
Is the problem with travis permanent? Any idea how to fix that? In this PR it works well.
All tests work when using mne stable in pypi, this PR modifies mne version in the `environment.yml` to mne from pypi. I missed it in the previous PR #100. I was surprised that our CI was still green - travis should not pass but it was never triggered.

robintibor:
great, merged!

PR #107, opened 2020-04-23
Support for regression task git cleaned

sliwy:
This is PR with a cleaned git tree based on changes from #99. I added also train dataset shuffling from default for `EEGRegressor`.
@gemeinl please check if it contains all your changes, I might miss something.
Closes #99

### DIFF
--- a/braindecode/__init__.py
+++ b/braindecode/__init__.py
@@ -1,3 +1,4 @@
 from .version import __version__
 
 from .classifier import EEGClassifier
+from .regressor import EEGRegressor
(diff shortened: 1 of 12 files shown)


robintibor:
Great @sliwy .

This is diff to lukas version: https://github.com/gemeinl/braindecode-1/compare/age_regression...sliwy:age_regression
Looks fine to me, only contains some additional stuff from sliwy for default argument shuffle etc.

So let's merge this and close other @gemeinl ? You can do it! :D

PR #114, opened 2020-05-07
[WIP] Dataset examples

gemeinl:
Creating examples to demonstrate usage of mne datasets and custom datasets with braindecode.


gemeinl:
@robintibor Should I also add a function `create_from_mne`?

robintibor:
ja warum nicht!

braindecode/datautil/windowers.py
```diff
@@ -176,7 +176,7 @@ def create_fixed_length_windows(
     for ds in concat_ds.datasets:
         # already includes last incomplete supercrop start
         stop = (ds.raw.n_times
-                if stop_offset_samples is None
+                if (stop_offset_samples is None or stop_offset_samples == 0)
```
robintibor, line 179: why this == 0 here makes sense? can you add a small comment maybe?

braindecode/datautil/windowers.py
```diff
@@ -193,8 +193,7 @@ def create_fixed_length_windows(
         if mapping is not None:
             target = mapping[target]
 
-        fake_events = [[start, supercrop_size_samples, -1]
-                        for i_start, start in enumerate(starts)]
+        fake_events = [[start, supercrop_size_samples, -1] for start in starts]
```
robintibor, line 196: nice catch

robintibor:
Should I add tests for the create_ functions tomorrow?

braindecode/datautil/windowers.py
gemeinl: Comment here or in the code you mean?
`stop_offset_samples` can be unset (`None`) or `int` (including 0). Both, `None` and 0 have the same meaning: there is no stop offset.

gemeinl:
> Should I add tests for the create_ functions tomorrow?

Yes, if someone wants to help, creating tests would be great!

braindecode/datautil/windowers.py
gemeinl: Alternatively, instead of adding this here, one could adjust the checks (https://github.com/braindecode/braindecode/blob/baa86663b8765207b4f4a744661ce6fdc4a78da0/braindecode/datautil/windowers.py#L294-L295) of the windowing arguments to only allow integer values > 0 for `stop_offset_samples` which probably makes more sense. What do you think?

robintibor:
just first attempt at test, more to come in future

PR #116, opened 2020-05-11
EEGClassifier and EEGRegressor docstring update

### LINKED ISSUE #92: EEGClassifier docstrings
Make `EEGClassifier` docstring display all the parameteres (including base class). Probably we should use a way similar to skorch which uses `__doc__` attribute of base class and modifies it (see here https://github.com/skorch-dev/skorch/blob/560e72149914b2d99ce6477226dd91835db8c37f/skorch/classifier.py#L61).

sliwy:
I made EEGClassifier and EEGRegressor use skorch docstrings as all the parameters from skorch may be used also in our classes. 
At the beginning of `Parameters` section, there are parameters from skorch and in the end there are parameters added by us, specific for our classes. The order may be changed but I think, that this is the most reasonable way.

Closes #92

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -6,11 +6,11 @@ from skorch.classifier import NeuralNetClassifier
 from skorch.utils import train_loss_score, valid_loss_score, noop
 
 from .scoring import PostEpochTrainScoring, CroppedTrialEpochScoring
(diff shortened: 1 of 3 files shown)


robintibor:
cool, ya we can see if we should switch order later

PR #117, opened 2020-05-11
Refactorings and Renamings

robintibor:


agramfort:
don't forget to updated https://github.com/braindecode/braindecode/blob/master/docs/api.rst with new public functions and classes. thx

robintibor:
after finishing of refactoring I will make a whole pass/restructure doc starting page etc.

robintibor:
> mikrovolt -> microvolt

ja nagut, aber eigentlich ist es mikrovolt ;p ;) but changed, yes.

braindecode/classifier.py
```diff
@@ -5,7 +5,7 @@
 from skorch.classifier import NeuralNetClassifier
 from skorch.utils import train_loss_score, valid_loss_score, noop
 
-from .scoring import PostEpochTrainScoring, CroppedTrialEpochScoring
+from braindecode.training.scoring import PostEpochTrainScoring, CroppedTrialEpochScoring
```
agramfort, line 8: I think we only use full imports for tests. Here it should be a relative import

braindecode/datasets/base.py
```diff
@@ -24,7 +24,7 @@ class BaseDataset(Dataset):
     Parameters
     ----------
     raw: mne.io.Raw
-    description: dict | pandas.Series | None
+    description: dict | pandas.Series | Nonew
```
agramfort, line 27: Nonew
->
None

agramfort:
@robintibor I would advise to get at least one approval from someone else before merging.
It's a necessary step to have more people know the history of the code base and also spot typos I saw by glancing over the first files.

robintibor:
yes now everything a bit rushed for presentation today, will try to keep to it again afterwards, and do things earlier next time. Thanks for catches

PR #121, opened 2020-05-13
autopep8 run

robintibor:


robintibor:
not happy with what it is doing to examples, let's do another time.

PR #126, opened 2020-05-13
Small details added in installation in readme and index

sliwy:
Added MOABB installation as a step in installation process. Moved "special" case of installing newest version of braindecode from github using pip to be a "normal" alternative way.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -12,15 +12,30 @@ Installation
 
 1. Install pytorch from http://pytorch.org/ (you don't need to install torchvision).
 
-2. Install braindecode via pip:
+2. Install MOABB from https://github.com/NeuroTechX/moabb, needed if you want to use MOABB datasets utilities:
+
+.. code-block:: bash
+
+  git clone https://github.com/NeuroTechX/moabb.git
+  pip install -r moabb/requirements.txt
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -13,21 +13,30 @@ Installation
...
-2. Install numpy (necessary for resamply installation to work), e.g.:
+2. Install MOABB from https://github.com/NeuroTechX/moabb, needed if you want to use MOABB datasets utilities:
 
 .. code-block:: bash
 
-  pip install numpy
+  git clone https://github.com/NeuroTechX/moabb.git
+  pip install -r moabb/requirements.txt
+  pip install moabb/
+  rm -r moabb/
```
agramfort, line 23:
```
pip install download
python -c "from download import download; download('https://raw.githubusercontent.com/NeuroTechX/moabb/master/requirements.txt', 'requirements.txt', replace=True)"
pip install -r requirements.txt
pip install -U https://github.com/NeuroTechX/moabb/archive/master.zip
```
seems better as it will work on any machine even without git installed.

sliwy: thanks!

robintibor:
great

PR #129, opened 2020-06-12
Fix exp standardize

robintibor:
Init block size was not used correctly as noticed in https://github.com/TNTLFreiburg/braindecode/issues/56
Test was too unspecific to catch it. Now fixed it and made test more specific (would have failed on old code).

### DIFF
--- a/braindecode/datautil/preprocess.py
+++ b/braindecode/datautil/preprocess.py
@@ -148,12 +148,12 @@ def exponential_moving_standardize(
     standardized = demeaned / np.maximum(eps, np.sqrt(np.array(square_ewmed)))
     standardized = np.array(standardized)
     if init_block_size is not None:
-        other_axis = tuple(range(1, len(data.shape)))
+        i_time_axis = 0
         init_mean = np.mean(
-            data[0:init_block_size], axis=other_axis, keepdims=True
+            data[0:init_block_size], axis=i_time_axis, keepdims=True
         )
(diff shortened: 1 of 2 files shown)


robintibor:
superstrangeto me that test_trialwise_decoding now fails . It doesn't use the changed functions at all. also still runs fine on my machine, so I maybe something changed in an underlying library? like mne/skorch? errors seem a bit big to be due too hardware fluctuations I would say? very strange.

sliwy:
I just compared versions between this log and our master build and the most suspected one is skorch 0.7.0 -> 0.8.0.
If we want to test it, mybe we can add skorch version 0.7.0 to the requirements file and in the environment.yml?
There is also a change in numpy and mne, that's what I checked so far, but I do not think that it may be responsible for that.

robintibor:
I confirmed also on my machine it fails with skorch 0.8. 
These are change in skorch 0.7->0.8: https://github.com/skorch-dev/skorch/compare/v0.7.0...v0.8.0#diff-7cbd7999ce1895528c101345c52cbae4 https://github.com/skorch-dev/skorch/releases

From the tests: 
1) Seems that first values is always correct (like at start of training), then from second on, values are different
2) test_cropped_decoding.py still everything correct

To me still not clear how this is possible. If it is change in randomization, would expect also test_cropped_decoding.py to fail. If change in scoring logic, why are the first values always ok?

Issue #130, opened 2020-06-23
Circular import ?


Simon-Free:
Hello everyone,

I tried to add some new files to braindecode.datautil, and weren't able to load them properly. During the process, I found one error : 
Input : `from braindecode.datautil import *`
Output : 
```
ImportError                               Traceback (most recent call last)
<ipython-input-5-81dbcb36302e> in <module>
----> 1 from braindecode.datautil import *

~/miniconda3/lib/python3.7/site-packages/braindecode/datautil/__init__.py in <module>
      3 """
      4 
----> 5 from .windowers import create_windows_from_events, create_fixed_length_windows
      6 from .preprocess import zscore, scale, exponential_moving_demean, exponential_moving_standardize

~/miniconda3/lib/python3.7/site-packages/braindecode/datautil/windowers.py in <module>
     13 import pandas as pd
     14 
---> 15 from ..datasets.base import WindowsDataset, BaseConcatDataset
     16 
     17 

~/miniconda3/lib/python3.7/site-packages/braindecode/datasets/__init__.py in <module>
      5 from .moabb import MOABBDataset
      6 from .tuh import TUHAbnormal
----> 7 from .mne import create_from_mne_raw, create_from_mne_epochs
      8 from .xy import create_from_X_y

~/miniconda3/lib/python3.7/site-packages/braindecode/datasets/mne.py in <module>
      4 
      5 from .base import BaseDataset, BaseConcatDataset, WindowsDataset
----> 6 from ..datautil.windowers import (
      7     _check_windowing_arguments, create_windows_from_events)
      8 

ImportError: cannot import name '_check_windowing_arguments' from 'braindecode.datautil.windowers' (/home/parietal/sfreybur/miniconda3/lib/python3.7/site-packages/braindecode/datautil/windowers.py)
```

Cause seems to be a circular import, but I am not confident enough to change the code organisation by myself.

robintibor:
Thanks for the info! Can you try moving the imports  

https://github.com/braindecode/braindecode/blob/41a216572f08ef23fd0af47e10e89bc9179cce5e/braindecode/datasets/mne.py#L6-L7

inside the start of these functions (after docstring of course):

https://github.com/braindecode/braindecode/blob/41a216572f08ef23fd0af47e10e89bc9179cce5e/braindecode/datasets/mne.py#L10-L13

[...]

PR #131, opened 2020-06-23
Moved files using datasets to datautil to fix circular imports

sliwy:
I took a quick look at #130 and I moved two files which uses things from datautil (`xy.py` and `mne.py`) to the datautil module. Now I can do `from braindecode.datautil import *`. @robintibor @agramfort let me know what you think, is it a good place for those files? They are just functions creating datasets.

@Simon-Free can you check if it works now?

One more thing is that I hardcoded skorch version as our tests does not work with the newest one. #129

### DIFF
--- a/examples/plot_custom_dataset_example.py
+++ b/examples/plot_custom_dataset_example.py
@@ -11,7 +11,7 @@
 
 import mne
 
-from braindecode.datasets.xy import create_from_X_y
+from braindecode.datautil.xy import create_from_X_y
 
 ###############################################################################
 # To set up the example, we first fetch some data using mne:
--- a/examples/plot_custom_dataset_example.py
+++ b/examples/plot_custom_dataset_example.py
@@ -11,7 +11,7 @@
 
 import mne
 
-from braindecode.datautil.xy import create_from_X_y
+from braindecode.datautil import create_from_X_y
 
 ###############################################################################
 # To set up the example, we first fetch some data using mne:
--- a/examples/plot_mne_dataset_example.py
+++ b/examples/plot_mne_dataset_example.py
@@ -11,7 +11,7 @@
 
 import mne
 
-from braindecode.datasets.mne import (
+from braindecode.datautil.mne import (
     create_from_mne_raw, create_from_mne_epochs)
 
 ###############################################################################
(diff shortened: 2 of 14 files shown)


Simon-Free:
How may I download your code ? I tried `git pull sliwy circular_import_fix`, but it does not work, any idea ?

robintibor:
did you add sliwy as a remote,e.g.,
`git remote add sliwy git@github.com:sliwy/braindecode.git`
(or https, depending on what you are using)

then you could do for example:
```
git fetch sliwy
git checkout sliwy/circular_imports_fix
```

robintibor:
Hmhm I wonder if it is better the way you did now or to say we move datasets/base.py together with what was before in datautil. then what is left in datasets would be "specific" sets like moabb, tuh, xy,mne,.... and everything generic in datautil. but then ofc naming may be confusing (basedataset in datautil not dataset...)
I don't know what you think? We can also discuss with @gemeinl  he returns from vacation next week...

sliwy:
If you look at those functions `create_from_mne_raw`, `create_from_mne_epochs`, `create_from_X_y` they all just create a dataset based on some form of data, in this case mne raw or mne epochs or just X and y. For me it is a kind of helper to create dataset but I am not sure if it should be in datasets module. For me it sounds like datautil to create a dataset. 
I agree that we can wait for @gemeinl to have better overview in datasets concept!

examples/plot_custom_dataset_example.py
```diff
@@ -11,7 +11,7 @@
 
 import mne
 
-from braindecode.datasets.xy import create_from_X_y
+from braindecode.datautil.xy import create_from_X_y
```
agramfort, line 14: could you make it so that:

`from braindecode.datautil import create_from_X_y`

works and is the recommended way @sliwy ?

examples/plot_mne_dataset_example.py
```diff
@@ -11,7 +11,7 @@
 
 import mne
 
-from braindecode.datasets.mne import (
+from braindecode.datautil.mne import (
```
agramfort, line 14:
```suggestion
from braindecode.datautil import (
```

? 

to avoid too deep imports...

robintibor:
@agramfort do you mean the import stuff in general so that in general there should be no imports deeper than two levels anywhere within the library? then maybe in separate PR?

robintibor:
great @sliwy @gemeinl  also agrees alles gut

PR #133, opened 2020-07-02
Resolving the confusion about subject ids and recordings ids in TUH

gemeinl:
Fixed naming from subject_ids to recording_ids.
Added the actual subject ids to the description pandas.Series.

### DIFF
--- a/braindecode/datasets/tuh.py
+++ b/braindecode/datasets/tuh.py
@@ -16,23 +16,23 @@ class TUHAbnormal(BaseConcatDataset):
     ----------
     path: str
         parent directory of the dataset
-    subject_ids: list(int) | int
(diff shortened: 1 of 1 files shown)


robintibor:
do you consider it reasonable/sensible to have some test for this? like create artificial data? or maybe test will be a lot of effort but still not test relevant part (that you read out data correctly  according to original specification TUH?)

gemeinl:
What exactly do you mean? Loading is done through mne, I think they should and do cover testing correct loading of data.
What could be tested is the whole parsing of labels, subjects, and sessions from file pathas and probably of ages and genders from the edf header.

robintibor:
ok let's just merge it then

PR #138, opened 2020-07-07
Adding code of Temporal Convolutional Network

gemeinl:


braindecode/models/tcn.py
```diff
@@ -0,0 +1,142 @@
...
+        (batch_size, _, time_size, _) = x.size()
+        assert time_size >= self.min_len
+        # remove empty trailing dimension
+        x = x.squeeze(3)
+        x = self.temporal_blocks(x)
+        # Convert to: B x T x C
+        x = x.transpose(1, 2).contiguous()
+
+        fc_out = self.fc(x.view(batch_size * time_size, x.size(2)))
+        if hasattr(self, "log_softmax"):
```
gemeinl, line 83: Should we remove the log_softmax layer in our models?
It would make the architectures more flexible in cases where we want a single output.
There exist loss functions in pytorch that directly include the log_softmax.

braindecode/models/tcn.py
```diff
@@ -0,0 +1,142 @@
...
+            dilation = 2 ** i
+            self.min_len += 2 * (kernel_size - 1) * dilation
+
+        # start in eval mode
+        self.eval()
+
+    def forward(self, x):
+        # x is in format: B x C x T x 1
+        (batch_size, _, time_size, _) = x.size()
+        assert time_size >= self.min_len
```
gemeinl, line 75: This will be called quite often. Is this critical and slows things down too much?
Is there a better way to do it?
Due to the padding, the code never crashes, although input time dimension might be too small.

braindecode/models/tcn.py
```diff
@@ -0,0 +1,142 @@
...
+        out = self.dropout1(out)
+        out = self.conv2(out)
+        out = self.chomp2(out)
+        out = self.relu2(out)
+        out = self.dropout2(out)
+        res = x if self.downsample is None else self.downsample(x)
+        return self.relu(out + res)
+
+
+class Chomp1d(nn.Module):
```
gemeinl, line 133: Is there a better name for this module? "to chomp" means "mampfen" or "kauen" in german which is really a precise description.

braindecode/models/tcn.py
robintibor: should be fine

braindecode/models/tcn.py
robintibor: well since it is from original code let's keep it

PR #139, opened 2020-07-08
Add tests for TUHAbnormal dataset

gemeinl:
Added tests for parsing the label, the subject_id and "train" or "eval" from file paths.
Replaced sorting through custom functions by sorting using pandas.
Improved documentation.

### DIFF
--- a/braindecode/datasets/tuh.py
+++ b/braindecode/datasets/tuh.py
@@ -15,16 +15,16 @@ class TUHAbnormal(BaseConcatDataset):
     ----------
     path: str
         parent directory of the dataset
-    recording_ids: list(int) | int
-        (list of) int of recording(s) to be read
+    recording_ids: list(int)
+        list of int of recording(s) to be read
     target_name: str
-        can be 'pathological', 'gender', or 'age'
+        can be "pathological", "gender", or "age"
     preload: bool
-        if True, preload the data of the Raw objects.
+        if True, preload the data of the Raw objects
     """
     def __init__(self, path, recording_ids=None, target_name="pathological",
                  preload=False):
-        all_file_paths = read_all_file_names(path, extension='.edf')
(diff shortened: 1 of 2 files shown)


braindecode/datasets/tuh.py
```diff
@@ -16,100 +15,115 @@ class TUHAbnormal(BaseConcatDataset):
...
+            pathological, train_or_eval, subject_id = (
+                self._parse_properties_from_file_path(file_path))
             age, gender = _parse_age_and_gender_from_edf_header(file_path)
-            # see https://www.isip.piconepress.com/projects/tuh_eeg/downloads/tuh_eeg_abnormal/v2.0.0/_AAREADME.txt
-            subject_id = path_splits[-3]
             description = pd.Series(
-                {'age': age, 'pathological': pathological, 'gender': gender,
-                'session': session, 'subject': subject_id}, name=recording_id)
+                {"age": age, "pathological": pathological, "gender": gender,
+                "session": train_or_eval, "subject": subject_id},
```
gemeinl, line 41: "Session" is strictly speaking incorrect here. There exist sessions in the TUH dataset, however, we do not use them. What is meant here is whether a recordings belongs to the training or final evaluation set. How to properly name this?

robintibor: maybe train_or_eval is the most direct? just wonder about consistency with other datasets, in case one has more than train or eval... on the other hand, i have come across a lot of datasets with predefined train-eval/train-test splits, but cannot remember predefined train-valid-test splits.. so maybe train_or_eval is the most clear?

gemeinl: I am afraid there will not be consistency.. For example, in BCIC IV 2a fetched through MOABB there is a column "session" with entries "session_T" and "session_E".

PR #147, opened 2020-07-24
Improve splitting of datasets

gemeinl:
```split()``` now has a single argument ```by``` which ca be a string, a list of integers, or a list of list of integers.
splitting now always returns a dictionary with string keys.

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -106,13 +106,13 @@ class BaseConcatDataset(ConcatDataset):
     """
     def __init__(self, list_of_ds):
         # if we get a list of BaseConcatDataset, get all the individual datasets
-        if isinstance(list_of_ds[0], BaseConcatDataset):
+        if list_of_ds and isinstance(list_of_ds[0], BaseConcatDataset):
             list_of_ds = [d for ds in list_of_ds for d in ds.datasets]
         super().__init__(list_of_ds)
         self.description = pd.DataFrame([ds.description for ds in list_of_ds])
         self.description.reset_index(inplace=True, drop=True)
 
-    def split(self, by=None, some_property=None, split_ids=None):
+    def split(self, by=None, property=None, split_ids=None):
         """Split the dataset based on information listed in its description
         DataFrame or based on indices.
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by agramfort** [comment]
would you ming add some test @gemeinl ? can this be illustrated in some documentation / example?

**Comment by gemeinl** [comment]
> would you ming add some test @gemeinl ? can this be illustrated in some documentation / example?

Not at all. I added an example as well as some tests.

File: braindecode/datasets/base.py [hunk 1]
```diff
@@ -109,34 +109,37 @@ def __init__(self, list_of_ds):
         super().__init__(list_of_ds)
         self.description = pd.DataFrame([ds.description for ds in list_of_ds])
 
-    def split(self, some_property=None, split_ids=None):
-        """Split the dataset based on some property listed in its description
+    def split(self, by):
```
agramfort, line 112: technically you are breaking public API here again. Maybe we want to reach a proper version
before stabilizing API. I feel there are still many moving parts.

examples/plot_split_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+========================
+
+In this example, we show multiple ways of how to split datasets.
+"""
+
+# Authors: Lukas Gemein <l.gemein@gmail.com>
+#
+# License: BSD (3-clause)
+
+from IPython.display import display
```
agramfort, line 11: why do you need this?
adding a dependency on ipython to do a print seems an overkill to me.

braindecode/datasets/base.py
gemeinl: My bad, I am not yet used to deprecating and not breaking the API. And yes, a lot of stuff is moving.

examples/plot_split_dataset.py
gemeinl: I just copied it from examples.plot_dataset_example.py, so the dependency was already there. Could of course remove it everywhere.

agramfort:
> I just copied it from examples.plot_dataset_example.py, so the dependency was already there. Could of course remove it everywhere.

I would do this. Maybe in another PR if you want

braindecode/datasets/base.py
agramfort: @gemeinl if you care about the design of the API it's great. Don't worry for now about breaking API as long as it's for the best.

Issue #136, opened 2020-07-06
Import Error when importing create_from_X_y


Khalizo:
Hi there, 

I have installed braindecode but get this error when trying to import create_from_X_y. Please how can I fix this?

![image](https://user-images.githubusercontent.com/22616237/86639239-84c7a280-bfd0-11ea-8579-3ea0494e0f08.png)

robintibor:
Hi this due to a change in the code that has not been pushed to PyPI yet.
You can either update to current master or import via 

```python
from braindecode.datasets.xy import create_from_X_y
```

See https://github.com/braindecode/braindecode/blob/41a216572f08ef23fd0af47e10e89bc9179cce5e/examples/plot_custom_dataset_example.py also

Khalizo:
Dear Robintibor,

Thanks for your swift response, I really appreciate all the work you
have done with this library!

I am also struggling with another issue. I am trying to predict levels of
attention using EEG signals from 8 channels, however I am not sure how to
convert it into a brain decode format.

My "X" is a dataframe with shape (2400, 8):
[image: image.png]

While my "y" has shape (2400,):
[image: image.png]

How can convert it into a format that is usable?

Thanks,
Babs

robintibor:
So, as far as I see you want to predict attention level at each point? can you share your data? Then I could have a try as well to see what is best implementation for your use case and document it better.

Khalizo:
Hi Robintibor,

Here is a sample file of 1 test from a single user. The user was given 5
different paragraphs to read and then gave a score between 1-5 of the level
of attention. 8 channels were used to record the EEG readings.

Please let me know if you have any questions.

Thanks,
Babs

robintibor:
There seems to be no sample file can you attach it again or link to it somewhere?

Khalizo:
Hi Robintibor,

Can you see it via this link below?
  user1_test10_attention.csv
<https://drive.google.com/file/d/1Y13SwJmUDYTmPsZd1DN-EH2FT6PGN2hx/view?usp=drivesdk>

Thanks,
Babs

robintibor:
@sliwy seems to me this might be a use case for continuous classification, would you be interested to have a look how to do that? Maybe it's already possible using custom loss function, maybe sth extra has to be implemented, not sure

Khalizo:
Hi Both,

My X_train is a 3D array with the shape (number of samples, timepoints,
channels) and my labels are in a 1D array (number of samples).

[...]

PR #148, opened 2020-07-24
Improve create_from_X_y

gemeinl:
split create_from_X_y into two functions, one returning BaseDatasets for further preprocessing on trial level, the second directly returning WindowsDatasets.

### DIFF
--- a/braindecode/datautil/xy.py
+++ b/braindecode/datautil/xy.py
@@ -10,7 +10,7 @@ from braindecode.datautil.windowers import (
 log = logging.getLogger(__name__)
 
 
-def create_raws_from_X_y(X, y, sfreq, ch_names=None):
+def create_trials_from_X_y(X, y, sfreq, ch_names=None):
     """Create a BaseConcatDataset of BaseDatasets from X and y to be used for
     decoding with skorch and braindecode, where X is a list of pre-cut trials
     and y are corresponding targets.
@@ -77,7 +77,7 @@ def create_windows_from_X_y(X, y, drop_last_window, sfreq=None, ch_names=None,
         X and y transformed to a dataset format that is compatible with skorch
         and braindecode
     """
-    base_datasets = create_raws_from_X_y(
+    base_datasets = create_trials_from_X_y(
         X=X, y=y, sfreq=sfreq, ch_names=ch_names)
     n_samples_per_x = [x.shape[1] for x in X]
     if window_size_samples is None and window_stride_samples is None:
--- a/braindecode/datautil/xy.py
+++ b/braindecode/datautil/xy.py
@@ -10,9 +10,46 @@ from braindecode.datautil.windowers import (
 log = logging.getLogger(__name__)
 
 
-def create_from_X_y(
-        X, y, drop_last_window, sfreq=None, ch_names=None,  window_size_samples=None,
-        window_stride_samples=None):
+def create_raws_from_X_y(X, y, sfreq, ch_names=None):
+    """Create a BaseConcatDataset of BaseDatasets from X and y to be used for
+    decoding with skorch and braindecode, where X is a list of pre-cut trials
+    and y are corresponding targets.
+
+    Parameters
+    ----------
+    X: array-like
+        list of pre-cut trials as n_trials x n_channels x n_times
+    y: array-like
+        targets corresponding to the trials
+    sfreq: common sampling frequency of all trials
+    ch_names: array-like
+        channel names of the trials
+
+    Returns
+    -------
+    base_datasets: BaseConcatDataset
+        X and y transformed to a dataset format that is compatible with skorch
+        and braindecode
+    """
+    base_datasets = []
+    if sfreq is None:
+        log.error("No sampling frequency given.")
+    if ch_names is None:
+        ch_names = [str(i) for i in range(X.shape[1])]
+        log.info(f"No channel names given, set to 0-{X.shape[1]}).")
+
+    for x, target in zip(X, y):
+        info = mne.create_info(ch_names=ch_names, sfreq=sfreq)
+        raw = mne.io.RawArray(x, info)
+        base_dataset = BaseDataset(raw, pd.Series({"target": target}),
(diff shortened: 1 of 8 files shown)


braindecode/datautil/xy.py
```diff
@@ -26,39 +24,62 @@ def create_from_X_y(
...
+    decoding with skorch and braindecode, where X is a list of pre-cut trials
+    and y are corresponding targets.
+
+    Parameters
+    ----------
+    X: array-like
+        list of pre-cut trials as n_trials x n_channels x n_times
+    y: array-like
+        targets corresponding to the trials
+    sfreq: common sampling frequency of all trials
```
agramfort, line 63:
```suggestion
    sfreq: float
        Common sampling frequency of all trials.
```

braindecode/datautil/xy.py
```diff
@@ -26,39 +24,62 @@ def create_from_X_y(
...
+    ch_names: array-like
```
agramfort, line 64:
```suggestion
    ch_names: list of str
```

braindecode/datautil/xy.py
```diff
@@ -26,39 +24,62 @@ def create_from_X_y(
...
+        whether or not have a last overlapping window, when
+        windows/windows do not equally divide the continuous signal
+    window_size_samples: int
+        window size
+    window_stride_samples: int
+        stride between windows
 
+    Returns
+    -------
+    windows_datasets: BaseConcatDataset
```
agramfort, line 76: why windows_datasets of type BaseConcatDataset and not WindowsDataset?

docs/api.rst
```diff
@@ -102,7 +102,8 @@ Data Utils
 .. autosummary::
    :toctree: generated/
 
-    create_from_X_y
```
agramfort, line 105: that's an API breakage without deprecation

gemeinl: @agramfort Is it fine to use mne.utils.deprecated to add a DeprecationWarning? Or is this confusing, since the function will then appear to be a mne and not a braindecode function?

braindecode/datautil/xy.py
gemeinl: So probably naming is the problem here. In braindecode, a ```BaseDataset``` is a wrapper around a single ```mne.Raw```. Similarly, a ```WindowsDataset``` is a wrapper around a single ```mne.Epochs```. Both can be concatenated using the ```BaseConcatDataset``` class. So, since here multiple ```WindowsDatasets``` are returned, it is actually a ```BaseConcatDataset``` of ```WindowsDataset```.

docs/api.rst
agramfort: i would copy such a util function in the braindecode somewhere.

braindecode/datautil/xy.py
agramfort: thanks for the explanation. In mne when you concat raw you still get raw and if you concat epochs you still  get epochs. Is it really necessary to have another type of dataset in the public API?

gemeinl: For the dataset part we tried to stick to the PyTorch API as closely as possible. Our ```BaseDataset``` inherits from ```torch.utils.data.dataset.Dataset``` and our ```BaseConcatDataset``` inherits from ```torch.utils.data.dataset.ConcatDataset```, see https://pytorch.org/docs/stable/data.html#torch.utils.data.ConcatDataset. But we can always discuss with @robintibor.

docs/api.rst
gemeinl: Ok, thanks! I will discuss with @robintibor about where to put it and how to do it.

braindecode/datautil/xy.py
agramfort: ok then I would just propose to update the docstring:

```suggestion
    concat_dataset: BaseConcatDataset
```

docs/api.rst
```diff
@@ -103,6 +103,8 @@ Data Utils
    :toctree: generated/
 
     create_from_X_y
+    create_trials_from_X_y
+    create_windows_from_X_y
```
agramfort, line 107: would it make sense to keep a single `create_from_X_y` and then internally, dependings with items of X have the same length use windows or base? Basically as a user do I need to care of I can just pass the chunks of EEG data I have an the models will just deal with them? if you expect some breakage as some models only work with WindowsDataset it's another story but if you expect models to 'just work' maybe we can avoid some complexity on the user side.

gemeinl: Ok, I will try to explain what's going on. We just had a usecase (see https://github.com/braindecode/braindecode/issues/136) where a user had a dataset in Xy format, however, it was not preprocessed. So he wanted to convert it into braindecode format to be able to use the available braindecode/mne preprocessing. So far, the ```create_from_X_y``` function assumed data was already preprocessed and directly transformed it into compute windows for decoding. So it did not allow for preprocessing. I suggested something similar to you (see https://github.com/braindecode/braindecode/issues/136#issuecomment-662299096), @robintibor did not agree. It is still undecided / open for discussion.

agramfort:
to be honest rather than making complex the API of brain decode I would
have written an example
or referred the user to mne documentation to obtain raw instances and do
his preprocessing
properly. My worry is how are users going to discover the use cases for
these functions?
the more you have different ways to do the same thing the harder it is to
document it.
If there is only one prefered way it may force you to do an extra step but
at least all
scripts will look similar among users.

>

gemeinl:
I totally see your points. @robintibor will be back on monday, we will then hopefully discuss all the PRs and issues.
Thank you very much for your valuable input @agramfort! I appreaciate it!

robintibor:
Ok yes in the end maybe we settle towards our main example:
* starting from X,y
* explicitly going through mne (like creating raws etc.) in the example (noting this can be skipped if one already has mne data)
*  then starting all braindecode from here

robintibor:
closing this as it has diverged a bit much from current code, still very useful for addressing #544

PR #152, opened 2020-08-08
Improvements to the windowers

hubertjb:
This PR introduces some minor improvements to the windowers:

- fixing windower functions when `first_samp` is not 0 (i.e., the Raw object has been cropped)
- adding `epochs_kwargs` to windowing functions
- modifying default value of `stop_offset_samples` to None for consistency
- improving docstrings
- adding random state argument to `create_mne_dummy_raw`
- adding tests


braindecode/datautil/windowers.py
```diff
@@ -154,7 +168,7 @@ def create_windows_from_events(
 def create_fixed_length_windows(
         concat_ds, start_offset_samples, stop_offset_samples,
         window_size_samples, window_stride_samples, drop_last_window,
-        mapping=None, preload=False, drop_bad_windows=True):
+        mapping=None, preload=False, drop_bad_windows=True, **epochs_kwargs):
```
agramfort, line 171: i don't like kwargs. I would expose and document here the Epochs parameters you think can be useful.

braindecode/datautil/xy.py
```diff
@@ -69,7 +69,7 @@ def create_from_X_y(
     windows_datasets = create_fixed_length_windows(
         base_datasets,
         start_offset_samples=0,
-        stop_offset_samples=0,
+        stop_offset_samples=None,
```
agramfort, line 72: API change?

braindecode/util.py
```diff
@@ -250,7 +252,10 @@ def create_mne_dummy_raw(n_channels, n_times, sfreq, include_anns=True,
     save_fname : dict | None
         Dictionary containing the name the raw data was saved to.
     """
-    data = np.random.rand(n_channels, n_times)
+    if random_state is None or isinstance(random_state, int):
+        random_state = np.random.RandomState(random_state)
```
agramfort, line 256: use sklearn.utils.check_random_state

test/unit_tests/datautil/test_windowers.py
```diff
@@ -14,13 +16,15 @@
 from braindecode.datasets.moabb import fetch_data_with_moabb
 from braindecode.datautil import (
     create_windows_from_events, create_fixed_length_windows)
+from braindecode.datautil.preprocess import MNEPreproc, preprocess
```
agramfort, line 19: from xxx.preprocess import preprocess 

I don't like that as it's then confusing what is a function and what is a module

braindecode/datautil/windowers.py
hubertjb: Fixed in new commit.

braindecode/datautil/xy.py
hubertjb: That's correct. Is there a recommended way to flag API changes? I added a warning here: https://github.com/braindecode/braindecode/pull/152/files#diff-c031004527cfc6287fbb1741b7f3e1ddR231

test/unit_tests/datautil/test_windowers.py
hubertjb: What about we rename `braindecode.datautil.preprocess` to `braindecode.datautil.preprocessing`? It's a bit longer, but at least we avoid this ambiguity. @robintibor @gemeinl

agramfort:
>
> What about we rename braindecode.datautil.preprocess to
> braindecode.datautil.preprocessing? It's a bit longer, but at least we
> avoid this ambiguity. @robintibor <https://github.com/robintibor> @gemeinl
> <https://github.com/gemeinl>
>

3 levels of modules seem too much for a small package as it is now. I would
expose it
in  braindecode.datautil or consider a new
submodule braindecode.preprocessing

test/unit_tests/datautil/test_windowers.py
gemeinl: Personally, I don't like the name "datautil" at all and feel like we could use some major refactoring here. My thoughts:
Currently, ``braindecode.datautil`` for me contains a mess of 
- Loading and saving datasets (``serialization.py``) 
- Creating datasets (``xy.py``, ``mne.py``)
- Preprocessing (``preprocess.py``)
- Creating windows ((``windowers.py``) which is also a bad name, since we don't have windower classes anymore). Also, creating windows is a kind of preprocessing

We could probably rename ``braindecode.datautil`` to ``braindecode.preprocessing`` with ``preprocessing.py`` which contains a merge of ``preprocess.py`` and ``windowers.py``.

We could add ``braindecode.serialization`` with ``serialization.py``

I am not so sure what to do about ``xy.py`` and ``mne.py``, though.

braindecode/datautil/windowers.py
```diff
@@ -86,23 +104,28 @@ def create_windows_from_events(
...
+            [a['duration'] for a in ds.raw.annotations
+             if a['description'] in events_id]
         )
         stops = onsets + (filtered_durations * ds.raw.info['sfreq']).astype(int)
+        # XXX This could probably be simplified by using chunk_duration in
+        #     `events_from_annotations`
 
-        if stops[-1] + trial_stop_offset_samples > len(ds):
+        if (stops[-1] + trial_stop_offset_samples >
+                ds.raw.n_times + ds.raw.first_samp):
```
robintibor, line 117: This seems potentially quite errorprone code, do we have a test for this?

braindecode/datautil/windowers.py
```diff
@@ -86,23 +104,28 @@ def create_windows_from_events(
...
                              f'({trial_stop_offset_samples}) must be smaller '
-                             f'then length of recording {len(ds)}.')
+                             f'then length of recording ({len(ds)}).')
 
         if infer_window_size_stride:
             # window size is trial size
             if window_size_samples is None:
-                window_size_samples = stops[0] - (onsets[0] + trial_start_offset_samples)
+                window_size_samples = stops[0] - (
+                    onsets[0] + trial_start_offset_samples)
```
robintibor, line 128: here does not seem to account for trial stop samples?

braindecode/datautil/windowers.py
```diff
@@ -18,15 +20,17 @@
...
+
+    This function extracts windows of size window_size_samples in the interval
+    [trial_start_offset_samples, trial_stop_offset_samples] around each event,
+    with a separation of window_stride_samples between consecutive windows. If
+    the last window around an event does not end at trial_stop_offset_samples
+    and drop_last_window is set to False, an additional overlapping window that
+    ends at trial_stop_offset_samples is created.
 
     in mne: tmin (s)                    trial onset        onset + duration (s)
     trial:  |--------------------------------|--------------------------------|
```
robintibor, line 36: is doc here maybe wrong/inconsistent? 
Below it says:
` trial_stop_offset_samples -       Stop offset from original trial stop, in samples.`
This is also what happens in code in _compute_window_inds:
https://github.com/braindecode/braindecode/blob/fac6cea5d6ea0af49dcb39c85d3be33dfa791c22/braindecode/datautil/windowers.py#L290-L294

braindecode/datautil/windowers.py
```diff
@@ -190,15 +208,20 @@ def create_fixed_length_windows(
...
+            'to indicate end of trial/recording. Using `None`.')
+        stop_offset_samples = None
 
     list_of_windows_ds = []
     for ds in concat_ds.datasets:
-        stop = ds.raw.n_times if stop_offset_samples == 0 else stop_offset_samples
-        stop = stop - window_size_samples
+        stop = ds.raw.n_times \
+            if stop_offset_samples is None else stop_offset_samples
+        stop = stop - window_size_samples + ds.raw.first_samp
```
robintibor, line 241: yeah great, not even sure what was logic behind previous decision. seems quite strange. I wonder actually if we had intended stop offset samples to refer to offset to end of recording? as it is in from events thingy and end of trial?

robintibor:
Did you see the comments @hubertjb ?

hubertjb:
> Did you see the comments @hubertjb ?

@robintibor yes, sorry for the delay! I'm getting back to it today or tomorrow.

braindecode/datautil/windowers.py
hubertjb: Yes, `test_windowers.py::test_windows_cropped` used to test this indirectly - I'll add something more specific to this catch.

hubertjb: Actually I'm not 100% sure the behaviour of `mne.Epochs` is correct when the raw file has been cropped. Maybe there's something I'm missing that would make the handling of `raw.first_samp` simpler... I opened an issue here https://github.com/mne-tools/mne-python/issues/8201 just in case.

braindecode/datautil/windowers.py
hubertjb: Good point, I'll fix the docstring.

braindecode/datautil/windowers.py
hubertjb: Aah you have a good point, that's probably what was going on. I think the new approach might be more useful: let's say you want to extract windows from 10 minutes of a recording - you can provide the onset and offset as 2 and 12 mins, which is more natural than 2 and *calculate length of recording* - (2 + 10). What do you think?

robintibor: I agree that extracting a particular length if you don't have events might be what you want in many cases and would also keep like this for this PR. Only question long-term is if we want to have both options (stop given from beginning of recording and from end of recording), one could also do it in every dirty way: if negative, assume it is from end of recording, if positive, assume from start of recording... but as said let's keep in this way (from start of recording) for this PR

braindecode/datautil/windowers.py
robintibor: So now this will be correct, but need newest mne from master? just for clarification

robintibor (approved):

robintibor:
so seems ready to merge from what I see @hubertjb ?

hubertjb:
> so seems ready to merge from what I see @hubertjb ?

Almost! I just need to test with the latest MNE as you said - we might indeed need to ask for the latest version until the next version is out.

hubertjb:
@robintibor the latest commit takes the fix in MNE into account.

I guess we should update the requirements to ask for the latest MNE master (or should we wait for an MNE release)? @agramfort

hubertjb:
(The travis fail is because of the MNE version.)

robintibor:
When will this be in pip-installable mne @agramfort ?

agramfort:
@robintibor @hubertjb you can change the CI here to use:

pip install -U https://github.com/mne-tools/mne-python/archive/master.zip

new MNE version should be released on sept 15

robintibor:
ah great then we can also make new braindecode version on pip quickly after. And for now just put as alex said, if you don't know how to do it, i can do it for you as well, just let me know @hubertjb

hubertjb:
Tests are passing with the latest MNE master on Travis! Should we wait for the upcoming MNE release before merging?

agramfort (approved):

robintibor:
well how is ETA now for MNE release? last was "new MNE version should be released on sept 15" @agramfort ? should be soon then I guess?

agramfort:
it's imminent. Next monday at the latest

robintibor:
so do you want to try go back to MNE from pip since new version is there? @hubertjb then I can merge

hubertjb:
> so do you want to try go back to MNE from pip since new version is there? @hubertjb then I can merge

That'd be great @robintibor, thanks! I just reverted the change to the Travis config file, if tests pass we should be good to go.

robintibor:
hm so seems travis fails with old error, any reason for that @hubertjb @agramfort ? Also CircleCI seems quite a bit slower than all previous builds :/ Don't know why that might be

hubertjb:
Weird... could it be a cache thing? https://docs.travis-ci.com/user/caching/#configuration

robintibor:
It seems reason is mne is still at 0.20.8 at https://anaconda.org/conda-forge/mne . When will this change @agramfort ?

agramfort:
I don't maintain mne in conda. Please use pip

>

robintibor:
ok can you move mne to pip in https://github.com/braindecode/braindecode/blob/master/environment.yml ? I guess that should work then?

robintibor:
great  work @hubertjb ! merged!

PR #155, opened 2020-08-14
freeze torch rng for scoring functions; fix tests for skorch 0.8

robintibor:


robintibor:
So, I found:

* calling `__iter__()` of a PyTorch DataLoader causes a change in the random generator state of pytorch (even with shuffle=False etc.)
* Therefore, scoring functions that create a DataLoader can influence the later training, e.g. by influencing which data is in which training batch
* Removing the 'accuracy' metric scoring actually made results (loss) consistent between 0.7 and 0.8 skorch
* Also, just removing creation of EpochScoring for valid set made results (loss) consistent already
* Prints revealed training batches change with this valid EpochScoring added
* Therefore, strongly suggests something in random generator state changes, still not 100% clear what difference between 0.7 and 0.8 skorch

Still, overall feel it is likely enough that changes in metrics are due to random generator calls so I ensured that our scoring classes retrain previous random generator state on exit and updated numbers.

braindecode/training/scoring.py
```diff
@@ -273,8 +283,9 @@ def on_epoch_end(self, net, dataset_train, dataset_valid, **kwargs):
             for cb in epoch_cbs:
                 cb.y_preds_ = y_preds
                 cb.y_trues_ = y_test
-
         # y pred should be same as self.y_preds_
+        # robintibor@gmail.com: Unclear if this also leads to any
```
agramfort, line 287: you have git blame for this. I would not clutter the code with email adresses

test/acceptance_tests/test_cropped_decoding.py
```diff
@@ -119,23 +119,22 @@ def test_cropped_decoding():
         np.array(
             [
                 1.455306,
-                1.455934,
-                1.210563,
-                1.065806
+                1.784507,
+                1.421611,
+                1.057717
```
agramfort, line 124: why do you still need to change this? Do we test on CIs using skorch 0.7?

test/acceptance_tests/test_eeg_classifier.py
```diff
@@ -188,82 +188,66 @@ def test_eeg_classifier():
...
+         'train_batch_count': 3,
+         'train_loss': 1.0526101191838582,
+         'train_loss_best': False,
+         'train_trial_accuracy': 0.625,
+         'train_trial_accuracy_best': True,
+         'valid_batch_count': 1,
+         'valid_loss': 1.0293445587158203,
+         'valid_loss_best': True,
+         'valid_trial_accuracy': 0.25,
+         'valid_trial_accuracy_best': False}
```
agramfort, line 250: you bascially updated all the tests @robintibor so we cannot easily assess that your PR is a bug fix.

robintibor: The reason is that I now changed behavior to enforce that adding metrics will not change random state. Before, the results would actually change depending on if you had any extra metric added or none, because this would affect the random state during training.

braindecode/training/scoring.py
robintibor: ok nagut, removed

agramfort (approved):

PR #158, opened 2020-08-31
Added filterbanking

gemeinl:
Filterbanking can now be used similarly to the other preprocessing steps.
It modifies a ``mne.io.Raw`` in-place and adds new channels with the filtered signals.

### DIFF
--- a/braindecode/datautil/preprocess.py
+++ b/braindecode/datautil/preprocess.py
@@ -283,7 +283,10 @@ def filterbank(raw, frequency_bands, drop_original_signals=True, n_jobs=1,
     all_filtered = []
     for (l_freq, h_freq) in frequency_bands:
         filtered = raw.copy()
-        filtered.filter(l_freq=l_freq, h_freq=h_freq, **filter_kwargs)
+        filtered.filter(
+            l_freq=l_freq, h_freq=h_freq, picks="all", n_jobs=n_jobs,
+            method=method, iir_params=iir_params,  phase=phase,
+            fir_window=fir_window, fir_design=fir_design, **filter_kwargs)
         # mne automatically changes the highpass/lowpass info values
         # when applying filters and channels cant be added if they have
         # different such parameters. Not needed when making picks as
--- a/braindecode/datautil/preprocess.py
+++ b/braindecode/datautil/preprocess.py
@@ -12,6 +12,7 @@ ToDo: should transformer also transform y (e.g. cutting continuous labelled
 
 from collections.abc import Iterable
 from functools import partial
+from warnings import warn
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


braindecode/datautil/preprocess.py
```diff
@@ -251,3 +252,54 @@ def scale(data, factor):
...
+               fir_design='firwin', **filter_kwargs):
+    """Applies multiple bandpass filters to the signals in raw. Raw will be
+    modified in-place and number of channels in raw will be updated to
+    len(frequency_bands) * len(raw.ch_names) (-len(raw.ch_names) if
+    drop_original_signals). For filtering arguments please refer to
+    mne.filter.filter_data().
+
+    Params
+    ------
+    raw: mne.Raw
```
agramfort, line 268:
```suggestion
    raw: instance of mne.io.Raw
```

braindecode/datautil/preprocess.py
```diff
@@ -251,3 +252,54 @@ def scale(data, factor):
...
+        The raw singnals to be filtered
```
agramfort, line 269:
```suggestion
        The raw signals to be filtered
```

braindecode/datautil/preprocess.py
```diff
@@ -251,3 +252,54 @@ def scale(data, factor):
...
+def filterbank(raw, frequency_bands, drop_original_signals=True, n_jobs=1,
+               method='fir', iir_params=None, phase='zero', fir_window='hamming',
+               fir_design='firwin', **filter_kwargs):
+    """Applies multiple bandpass filters to the signals in raw. Raw will be
+    modified in-place and number of channels in raw will be updated to
+    len(frequency_bands) * len(raw.ch_names) (-len(raw.ch_names) if
+    drop_original_signals). For filtering arguments please refer to
+    mne.filter.filter_data().
+
+    Params
```
agramfort, line 266: Parameters

is the standard for numpydoc

braindecode/datautil/preprocess.py
```diff
@@ -251,3 +252,54 @@ def scale(data, factor):
...
+    ------
+    raw: mne.Raw
+        The raw singnals to be filtered
+    frequency_bands: list(tuple)
+        The frequency bands to be fitlered for (e.g. [(4, 8), (8, 13)])
+    drop_original_signals: bool
+        Whether to drop the original unfiltered signals
```
agramfort, line 273: missing many params

gemeinl: In the docstring above I stated to refer to ```mne.filter.filter_data()``` for a description of the parameters. Should I copy the docstring from mne instead and paste it here?

agramfort:
I would say yes. I expect to learn about parameters in the Parameters
section. Even **filter_kwargs should be explained there.

>

braindecode/datautil/preprocess.py
```diff
@@ -251,3 +253,63 @@ def scale(data, factor):
...
+        for ch in raw.ch_names[i::len(original_ch_names)]]
+    raw.reorder_channels(chs_by_freq_band)
+    if drop_original_signals:
+        raw.drop_channels(original_ch_names)
+
+
+# copy docstring from mne to here
+filterbank.__doc__ += mne.io.Raw.filter.__doc__[
+        mne.io.Raw.filter.__doc__.find("picks : "):
+        mne.io.Raw.filter.__doc__.find("Returns")].replace(8 * " ", 4 * " ")
```
agramfort, line 315: Honestly this is black magic. This code in mne barely changes so I would copy the docstring.
YOu can make a test that asserts that docs are in sync. It's much simpler.

robintibor: I feel previous solution to explicitly refer to MNE also has advantages:
* Less clutter in our code
* It is more explicit that we just forward these arguments, one does not have to check whether there is any difference between MNE and us

And we anyways build so much around MNE now, pretty much encouraging the user to learn some MNE.

braindecode/datautil/preprocess.py
```diff
@@ -251,3 +253,63 @@ def scale(data, factor):
     if hasattr(data, '_data'):
         data._data = scaled
     return scaled
+
+
+def filterbank(raw, frequency_bands, drop_original_signals=True, n_jobs=1,
+               method='fir', iir_params=None, phase='zero', fir_window='hamming',
+               fir_design='firwin', **filter_kwargs):
```
agramfort, line 260: why not simply having **filter_kwargs and say it supports params of filter function from mne?

if feels you explicit some params and some not.

gemeinl: In another PR you mentioned disliking kwargs and to make arguments explicit. I therefore took the arguments that are relevant for my use case and made them explicit, while at the same time keeping full compatibility with mne through filter_kwargs. I will now update it again to be less explicit and only have the filter_kwargs. The docstring for this parameter will then contain a referral to mne.

agramfort: arfff to be it's one option or the other. If it was me and I would copy all
the mne parameters and copy the docstring. After updating the docstring
5 times due to mne changes I would reconsider. Note that mne release
are every 6 months so...

PR #159, opened 2020-09-01
removed code duplicate in BaseDataset and WindowsDataset

gemeinl:


braindecode/datasets/base.py
```diff
@@ -56,6 +49,16 @@ def __getitem__(self, index):
...
 
+    def _create_description(self, description):
+        if description is not None:
+            if (not isinstance(description, pd.Series)
+                    and not isinstance(description, dict)):
+                raise ValueError(f"'{description}' has to be either a "
+                                 f"pandas.Series or a dict")
+            if isinstance(description, dict):
+                description = pd.Series(description)
+        return description
```
agramfort, line 60: this function does not do anything with self so it should not be a class method
but rather a private function in the file

robintibor:
always great to remove code duplication :+1:

agramfort:
thx @gemeinl !

PR #161, opened 2020-09-03
MRG: Sleep staging example

hubertjb:
I have started making a sleep staging example based on [this paper](https://ieeexplore.ieee.org/abstract/document/8307462/?casa_token=FVbhoYtuaIEAAAAA:PqILZbwMgZB7q_jVUHEWmUkdFS-AYnnpUM5-tysyb7qwr8lodey0kZH7GLTnezTo1K4hPsLbRFUQwg) and the [Sleep Physionet dataset](https://physionet.org/content/sleep-edfx/1.0.0/).

It's still work in progress, but feedback would already be useful!

Next steps:
* Add z-scoring of windows
* Use `crop_wake_mins` to trim Wake regions in Sleep Physionet
* Adapt hyperparameters to this low-data setting
* Plot learning curve/confusion matrix
* Add a short discussion at the end of the example on other models, refs, etc.

### DIFF
--- a/braindecode/datasets/sleep_physionet.py
+++ b/braindecode/datasets/sleep_physionet.py
@@ -11,8 +11,8 @@ from .base import BaseDataset, BaseConcatDataset
 class SleepPhysionet(BaseConcatDataset):
     """Sleep Physionet dataset.
 
-    Sleep dataset from https://physionet.org/content/sleep-edfx/1.0.0/. Contains
-    overnight recordings from 20 healthy subjects.
+    Sleep dataset from https://physionet.org/content/sleep-edfx/1.0.0/.
+    Contains overnight recordings from 78 healthy subjects.
 
     See [MNE example](https://mne.tools/stable/auto_tutorials/sample-datasets/plot_sleep.html).
 
@@ -28,15 +28,15 @@ class SleepPhysionet(BaseConcatDataset):
     load_eeg_only: bool
         If True, only load the EEG channels and discard the others (EOG, EMG,
         temperature, respiration) to avoid resampling the other signals.
-
-    TODO:
-    - Include all recordings from extended version (PR to MNE?)
-    - Crop before and after each file
+    crop_wake_mins: float
+        Number of minutes of wake time to keep before the first sleep event
+        and after the last sleep event. Used to reduce the imbalance in this
+        dataset. Ignore if
     """
     def __init__(self, subject_ids=None, recording_ids=None, preload=False,
-                 load_eeg_only=True):
+                 load_eeg_only=True, crop_wake_mins=0):
         if subject_ids is None:
-            subject_ids = range(20)
+            subject_ids = range(83)
         if recording_ids is None:
             recording_ids = [1, 2]
 
@@ -45,13 +45,15 @@ class SleepPhysionet(BaseConcatDataset):
         all_base_ds = list()
         for p in paths:
             raw, desc = self._load_raw(
-                p[0], p[1], preload=preload, load_eeg_only=load_eeg_only)
+                p[0], p[1], preload=preload, load_eeg_only=load_eeg_only,
+                crop_wake_mins=crop_wake_mins)
             base_ds = BaseDataset(raw, desc)
             all_base_ds.append(base_ds)
         super().__init__(all_base_ds)
 
     @staticmethod
-    def _load_raw(raw_fname, ann_fname, preload, load_eeg_only=True):
+    def _load_raw(raw_fname, ann_fname, preload, load_eeg_only=True,
+                  crop_wake_mins=False):
         ch_mapping = {
             'EOG horizontal': 'eog',
             'Resp oro-nasal': 'misc',
@@ -64,6 +66,19 @@ class SleepPhysionet(BaseConcatDataset):
         raw = mne.io.read_raw_edf(raw_fname, preload=preload, exclude=exclude)
         annots = mne.read_annotations(ann_fname)
         raw.set_annotations(annots, emit_warning=False)
+
+        if crop_wake_mins > 0:
+            # Find first and last sleep stages
+            mask = [
+                x[-1] in ['1', '2', '3', '4', 'R'] for x in annots.description]
+            sleep_event_inds = np.where(mask)[0]
+
+            # Crop raw
+            tmin = annots[int(sleep_event_inds[0])]['onset'] - \
+                   crop_wake_mins * 60
+            tmax = annots[int(sleep_event_inds[-1])]['onset'] + \
+                   crop_wake_mins * 60
+            raw.crop(tmin=tmin, tmax=tmax)
 
         # Rename EEG channels
         ch_names = {
--- a/braindecode/datasets/sleep_physionet.py
+++ b/braindecode/datasets/sleep_physionet.py
@@ -1,3 +1,8 @@
+# Authors: Hubert Banville <hubert.jbanville@gmail.com>
+#
+# License: BSD (3-clause)
+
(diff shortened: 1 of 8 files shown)


hubertjb:
Also, this PR currently contains some changes from https://github.com/braindecode/braindecode/pull/152, so it should not be merged before the other is merged.

braindecode/datasets/sleep_physionet.py
```diff
@@ -0,0 +1,96 @@
+import os
```
agramfort, line 1: author and licence missing

braindecode/datasets/sleep_physionet.py
```diff
@@ -0,0 +1,96 @@
...
+        recordings). Can be [1], [2] or [1, 2].
+    preload: bool
+        if True, preload the data of the Raw objects.
+    load_eeg_only: bool
+        If True, only load the EEG channels and discard the others (EOG, EMG,
+        temperature, respiration) to avoid resampling the other signals.
+    crop_wake_mins: float
+        Number of minutes of wake time to keep before the first sleep event
+        and after the last sleep event. Used to reduce the imbalance in this
+        dataset. Ignore if
```
agramfort, line 34: if?

braindecode/datautil/windowers.py
```diff
@@ -190,15 +208,20 @@ def create_fixed_length_windows(
...
+            'to indicate end of trial/recording. Using `None`.')
+        stop_offset_samples = None
 
     list_of_windows_ds = []
     for ds in concat_ds.datasets:
-        stop = ds.raw.n_times if stop_offset_samples == 0 else stop_offset_samples
-        stop = stop - window_size_samples
+        stop = ds.raw.n_times \
+            if stop_offset_samples is None else stop_offset_samples
+        stop = stop - window_size_samples + ds.raw.first_samp
```
agramfort, line 221:
```suggestion
        stop -= window_size_samples + ds.raw.first_samp
```
```suggestion
        stop = stop - window_size_samples + ds.raw.first_samp
```

braindecode/models/sleep_stager.py
```diff
@@ -0,0 +1,90 @@
+import torch
+from torch import nn
+
+
+class ChambonSleepStager(nn.Module):
```
agramfort, line 5: not sure about the naming here :)

braindecode/models/sleep_stager.py
```diff
@@ -0,0 +1,90 @@
+import torch
```
agramfort, line 1: author + license

test/unit_tests/datautil/test_windowers.py
```diff
@@ -14,13 +16,15 @@
 from braindecode.datasets.moabb import fetch_data_with_moabb
 from braindecode.datautil import (
     create_windows_from_events, create_fixed_length_windows)
+from braindecode.datautil.preprocess import MNEPreproc, preprocess
```
agramfort, line 19:
```suggestion
from braindecode.preprocessing import MNEPreproc, preprocess
```

would be so much nicer

braindecode/models/sleep_stager.py
agramfort: to be honest I would put the model in the example to limit the library size / scope in the future.
we can do a caffe zoo like repo later with many EEG models including pretrained ones.

what do others think?

braindecode/datasets/sleep_physionet.py
agramfort: I think you should default to the common thing ie 5 mins before and 5 mins after sleeping.

braindecode/models/sleep_stager.py
hubertjb: I guess the current `models` directory is pretty much a zoo-like collection of EEG models (7 models so far). I'm not against moving them to a separate repo, as long as it doesn't create too much overhead. What do @robintibor @gemeinl think?

hubertjb:
This new version is pretty complete I think.

braindecode/models/sleep_stager.py
agramfort: I am not suggesting to put it another repo but in the example file. To be reconsidered when used
across files.

with all the machinery of braindecode now having 20 to 40 lines in an example to write up the module
seems an ok solution. As soon as it's in the library code changing it is harder (deprecations etc..)

robintibor: I agree with @hubertjb that models is supposed to be already like zoo-like collection of EEG models. So it depends on if the model is expected to change a lot in the near future, then ok maybe keep it in example. but if it is fairly settled I think it is nice to have it in models dir

robintibor: Similar to how torchvision has quite some models as well: https://pytorch.org/docs/stable/torchvision/models.html

agramfort:
then we need to find a good name for the model ;)

>

hubertjb:
> then we need to find a good name for the model ;)
> […](#)

In most papers where this architecture is used as a baseline, it's referred to as "Chambon et al. 2018". Here are some other options:
1. SleepStager (current name)
2. StagerNet (how we've called it in https://arxiv.org/pdf/2007.16104.pdf)
3. ChambonSleepStager (used in a previous version of the PR)
4. SleepStagerIEEETransNeurSys (mmh - or some variation with the journal title)
5. SleepStageClassifier

I think 1 is good, however maybe not specific enough if we end up including other sleep staging architectures in braindecode. I'd say 3 makes sense too seeing how the model is referred to in the literature.

agramfort:
I would go with :

SleepStageClassifierChambon2018

or eventually:

SleepStagerChambon2018

hubertjb:
I change my vote to SleepStagerChambon2018!

robintibor:
Fine for me :)

agramfort:
deal

>

hubertjb:
@robintibor I think this PR is ready to be merged too!

agramfort (approved):
Looks good !

https://543-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_sleep_staging.html#sphx-glr-auto-examples-plot-sleep-staging-py

robintibor:
Thanks for the link @agramfort I saw two things:
1) 

> ### Split dataset into train and valid
> We can easily split the dataset using additional info stored in the description attribute, in this case session column. We select session_T for training and session_E for validation.

This seems from other example, please correct

2) I think for plotting it would be nicer to plot 1- balanced accuracy= balanced misclassification rate, as then curve shape is more easily comparable between misclassification rate and loss.

hubertjb:
Here's the new version: https://545-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_sleep_staging.html#sphx-glr-auto-examples-plot-sleep-staging-py

examples/plot_sleep_staging.py
```diff
@@ -0,0 +1,328 @@
...
+
+######################################################################
+# Split dataset into train and valid
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+
+######################################################################
+# We can easily split the dataset using additional info stored in the
+# `description` attribute of :class:`braindecode.datasets.BaseDataset`,
+# in this case ``session`` column. Here, we split the examples per subject.
```
robintibor, line 147:
```suggestion
# in this case using the ``subject`` column. Here, we split the examples per subject.
```

there is no session used here...

hubertjb: arrh thanks for spotting that @robintibor

robintibor:
Super! Merged!

PR #164, opened 2020-09-03
allow lazy preproc

robintibor:
Only load data if preprocessor needs it

### DIFF
--- a/braindecode/datautil/preprocess.py
+++ b/braindecode/datautil/preprocess.py
@@ -33,13 +33,21 @@ class MNEPreproc():
         self.kwargs = kwargs
 
     def apply(self, raw_or_epochs):
+        try:
+            self._try_apply(raw_or_epochs)
+        except RuntimeError:
+            # maybe data need to be loaded
+            raw_or_epochs.load_data()
+            self._try_apply(raw_or_epochs)
+
+    def _try_apply(self, raw_or_epochs):
         if callable(self.fn):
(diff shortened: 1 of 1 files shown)


robintibor:
one main use case here is that for crop function, otherwise all data is loaded before cropping, clearly undesirable

braindecode/datautil/preprocess.py
```diff
@@ -33,13 +33,21 @@ def __init__(self, fn, **kwargs):
         self.kwargs = kwargs
 
     def apply(self, raw_or_epochs):
+        try:
+            self._try_apply(raw_or_epochs)
```
agramfort, line 37: there is something easier. You can check the preload attribute of the instance.

if note raw_or_epochs.preload:
     raw_or_epochs.load_data()

robintibor: Maybe I should make the comment clearer:
Some mne functions do *not need* the data to be loaded (e.g., mne crop), and in this case, we want to apply the function without loading the data. I am not aware of any way to know beforehand if the function will need the data to be loaded, hence the try/except.
Before what happened was:
* I wanted to crop out only a small part of the recording using mne crop, and then apply other preprocessings next
* Before mne crop was applied, in our code, the whole data was loaded unnecessarily
* This makes it substantially slower than in the code now proposed (I did perf-comparisons)

Will clarify the comment in the code

PR #166, opened 2020-09-14
do not assume 4d in get_output_shape

robintibor:
Small fix


agramfort:
a test would have been much appreciated ;)

PR #167, opened 2020-09-18
Tests for output shape and gradients visualization + bugfix

robintibor:


robintibor:
and even found a bug! @agramfort must be so happy :)
Before, amplitude gradients were accidentally summed over filters, so gradient for filter 3 would be sum of grads filter 0,1,2,3 instead of being just gradient of filter 3 as intended.

agramfort:
I am happy but I would be even happier if you could update changelog to document the bug fix :)

robintibor:
Now I am sure you are so happy you want to invite me me for a Pizza Pesto from https://www.la-guichetta.fr/notre-carte/pizzas-au-feu-de-bois/ , correct? :)

agramfort:
I won't pay the full pizza yet :) 

what I have in mind is something like: https://mne.tools/stable/whats_new.html
or https://scikit-learn.org/stable/whats_new/v0.23.html

so it appears on the website so contributors get public credit for their efforts!

robintibor:
hmhm ok for now I will merge this so there is at least anything. then go further in separate pr

PR #177, opened 2020-11-03
Experimentation data aug

Simon-Free:
Hello,

I'm proposing some modification of the dataset structure to include a way to easily augment data. The core structure is based on [this article covering the canonic way to deal with data augmentation in Pytorch](https://pytorch.org/tutorials/beginner/data_loading_tutorial.html. I added to this a few useful functions. Here is my detailed proposal :

New classes:

* `Datum`: First, a unitary chunk of data is represented by a `Datum`, containing not only the signal X, but some metadata useful for transforms, like the label (used later to merge several signals with the same label) or the timestamp index (used to delay data)

* `TransformDataset, TransformConcatDataset`: Those classes derived from `WindowDataset, WindowConcatDataset` were created in the same spirit as the one presented in the article linked above. Suppose that there are `n` data in your raw dataset, having 2 available transformations will double its size thanks to the the `__getitem__` adaptation : with 2 transforms, the `i`-th element will be the `i//2` img to which the `i%2` transform will be applied. Theses classes also offer a way to retrieve the raw data, without transformation.

* `Transform`: This class take as argument a `probability`, a `magnitude`, and an operation. As described in most image recognition data augmentation framework, a transform applies on a data chunk with probability `p` an operation `t` of magnitude `m`.

New workflow:

* An augmented training manager takes a configured `ÈEGClassifier/EEGRegressor`, a train dataset, a list of transforms (transforms, which may actually be a composition of transforms using torchvision `Compose` class). It computes useful global variables potentially used by transforms, augments the dataset, and train the model on the new dataset, and return the trained model (it could, though, integrate also the testing and return only the final accuracy). But still, this manager is the minimal required feature. It should be discussed whether other features like best transform automatic search are or are not out of the scope or braindecode. Note that for the moment, only one operation is implemented (actually two, but they are following the same principle (okay, actually three with identity)), other will come if the PR is accepted.

* A new acceptance test is available, testing the augmented training on a dummy dataset, so are some unit test for transforms and other tools

A few other minor updates are available, like a function to tiny a dataset without using a Subset (which is memory-speaking far from efficient), or some tools to create the dummy dataset used for acceptance and transform test.

Feel free to ask questions or for more informations 😃 !


gemeinl:
I did not yet go through it in detail but I already have some questions. Maybe you can clarify.

- Is there no other way for data augmentation but to introduce 3  new dataset classes? I have the impression that this heavily complicates the API and I would prefer not to do this.

- Why do we need  ```tinying_dataset```?

- Did the PR somehow get messed up with the one by @hubertjb ?

Simon-Free:
* Ok, actually, I'm only including two new dataset classes (TransformDataset & TransformConcatDataset). They are conceived to be able to replace WindowDataset and WindowConcatDataset everywhere in the code, and expected to behave exactly as those old classes when no augmentation are provided. In practice, WindowDataset and WindowConcatDataset would then just never be used outside of the definition of TransformDataset & TransformConcatDataset. I could have enrich WindowDataset and WindowConcatDataset, but this superclass enable to see which function concerns transforms and which not. But we could consider this enrichment, or a renaming of TransformDataset, though.

* `tinying_dataset` has nothing to do with transforms : it was conceived while facing the following issue : I was trying to build a dummy dataset to test my functions/classes, but the "Subset" canonical way to take a subset does not reduce the actual size of the dataset, and I build my own function which really delete all non-useful elements. Plus, I wanted to have some control on which elements I was keeping, as I specifically needed (at least) one data chunk for each label present in my test set. If you have another efficient and clearer way to do this, let me know :)

* I was using the SleepPhysionet Extended dataset for my experiments, like Hubert, so I partly merged his branch and mine. Did I make a mistake 😕 ?

gemeinl:
>     * Ok, actually, I'm only including two new dataset classes (TransformDataset & TransformConcatDataset). They are conceived to be able to replace WindowDataset and WindowConcatDataset everywhere in the code, and expected to behave exactly as those old classes when no augmentation are provided. In practice, WindowDataset and WindowConcatDataset would then just never be used outside of the definition of TransformDataset & TransformConcatDataset. I could have enrich WindowDataset and WindowConcatDataset, but this superclass enable to see which function concerns transforms and which not. But we could consider this enrichment, or a renaming of TransformDataset, though.

In dataset.py there is also this Datum class (https://github.com/braindecode/braindecode/pull/177/files#diff-0e3c9da7298e0ba7355437c41fcfd2d4ef72d0dddb5eb06e857a25499c5d3e64R24) which does not look like it is doing much, right? I thought one idea was to add a list / compose of transforms (implementing data augmentation) as an initialization argument to our existing dataset classes. They would then be applied on-the-fly when ```__getitem__``` gets called. Is there a limitation to this idea which is solved by your Transform*Dataset classes? Or would it be possible to implement it like this?

>     * `tinying_dataset` has nothing to do with transforms : it was conceived while facing the following issue : I was trying to build a dummy dataset to test my functions/classes, but the "Subset" canonical way to take a subset does not reduce the actual size of the dataset, and I build my own function which really delete all non-useful elements. Plus, I wanted to have some control on which elements I was keeping, as I specifically needed (at least) one data chunk for each label present in my test set. If you have another efficient and clearer way to do this, let me know :)

I am not sure if I understand correctly, what exactly do you mean with "reduce the actual size of the dataset"? The dataset classes also implement a ```split``` method which you can use to select some datasets based on indices or properties listed in the ```description``` data frame. If you would like to shorten / crop the signals you could do this with a preprocessing function using ```mne.Epochs.crop```. If you would like to only select some compute windows, you could use ```mne.Epochs.drop```. If you want to only pick a subset of channels, you could do this with ```mne.Epochs.pick``` or ```mne.Epochs.pick_channels```

>     * I was using the SleepPhysionet Extended dataset for my experiments, like Hubert, so I partly merged his branch and mine. Did I make a mistake confused ?

It looks like something undesired happened. For example, you seem to re-add an old version of the SleepStager (braindecode/models/sleep_stager.py), which was apparently already added and renamed (https://github.com/braindecode/braindecode/blob/master/braindecode/models/sleep_stager_chambon_2018.py).

Simon-Free:
TransformDataset: Modifying the WindowsDataset is ok I guess, but the two functions `get_raw_data` & `update_augmentation_policy` also need to be added in. `get_raw_data` is not particularly useful for the masking operation I integrated as example, but e.g for signals merging, this is essential. `update_augmentation_policy` permit to simply update transformations after the dataset has been initialized, useful if a lot of experiences have to be done.

Datum: Ok, Datum is maybe not at the best place for the moment, and maybe it seems like it does not much, but it is very useful to harmonize transforms. Operations (transform = operation w.p `p`) sometimes needs more informations than just `X`. Example : signal merging needs `X` plus `y` the label of `X`, in order to merge `X` with a signal `X'` associated to the same label. So, we have `masking_along_time(X, magnitude)` and `signal_merging(X, y, magnitude)`, and we would like to have an uniformized syntax for all of the transforms. So, we put all potentential useful informations in a Datum object, we always have the syntax `operation(datum, magnitude)`, and operation have the possibility to use the metadata if needed. Is this clearer ? 🙂 

Tinying: One dataset is still too big for me : I have still ~1000 epochs in it, and I would like 5. From what you describe, `mne.Epochs.drop` seems to do the job, I will try it !

Sleepstager: Ok, that was probably a merge error, I'll take a walk and will patch it right after

Simon-Free:
One small question : I can integrate the `TransformDataset` class into the `WindowsDataset` class, but I had not seen that the `WindowsConcatDataset` was not existing. However, in `TransformConcatDataset`, I used to propagate a modification of the transforms into the `TransformDataset` objects. Since the notion of transform is only defined at the WindowsDataset level, I seemingly cannot simply reuse a `BaseConcatDataset`, and need to create at least a `WindowsConcatDataset` class (I though this one was already existing). Would it be ok ?

braindecode/augmentation/transform_class.py
```diff
@@ -0,0 +1,26 @@
...
+import random
+
+
+class Transform:
+
+    def __init__(self, operation, probability=None, magnitude=0):
+        self.operation = operation
+        self.probability = probability
+        self.magnitude = magnitude
+        self.transform = partial(operation, magnitude=magnitude)
```
robintibor, line 11: I would not make the assumption that any operation must have some magnitude.

robintibor: Rather just don't have magnitude and if your operation needs it you can still supply it directly to your transform object or via partial to your transform function if needed

robintibor: Rather just don't have magnitude and if your operation needs it you can still supply it directly to your transform object or via partial to your transform function if needed

robintibor:
Thanks for your work, I think it's really important for us to have some functionality for data augmentations.

I think we need to ensure that the overall code/modules have a simple-to-understand structure and are very well decoupled to allow easy extensiblity with new transformations etc.

There are a couple of things I still need to understand here.

First, I thought what we would need for data augmentation, also to be similar to torchvision:
1. the existing datasets accept a possibly empty list of transformations
2. transforms by itself are simply classes implementing __call__ and meeting certain expectations regarding input/output
3. datasets calling these transforms on __getitem__ supplying any needed data

This would be very minimal and should not require introducing new classes like Datum... tricky part would be to determine what data to supply/forward to the transformation, such that all needed data is there, it remains fast, etc.

I think such things like augmentation training manager first should be just in an example, not inside codebase until we have a clearly implemented simple augmentation part.

Also changing  transforms should be possible in the beginning without adding a dataset class, even if user has to write a little extra code at the start.

I think "Transform" class to me I would prefer to give it a different name, general transforms should just follow the to-be-determined interface in their call method and that's it. If you want to have a augmentation that is only applied with certain probability it could be called "StochasticTransformer" or something like that.  


Also atm the PR is a bit mixed with unrelated formatting changes in other files (bbci.py, tuh.py etc.), could be nice if they are removed. 


Do you think you can distill something a little bit simpler/more minimal out of your code? I mean the general direction seems close to me already.

braindecode/augmentation/transforms/masking_along_axis.py
```diff
@@ -0,0 +1,128 @@
...
+    mask_start = params["mask_start"]
+    mask_end = params["mask_end"]
+
+    if params["axis"] == 1:
+        X[:, mask_start:mask_end, :, :] = params["mask_value"]
+    elif params["axis"] == 2:
+        X[:, :, mask_start:mask_end, :] = params["mask_value"]
+    else:
+        raise ValueError('Only Frequency and Time masking are supported')
+    X = X.reshape(X.shape[:-2] + X.shape[-2:])
```
tomMoral, line 31: Maybe I am missing something here but this seems to be a noop.

tomMoral (commented):
Thanks @Simon-Free  for the proposal. Here is a small pass with some comment.

Simon-Free:
Ok, a few updates on my PR :

**IMPORTANT**: some acceptances tests are not working any more on the master branch. This is probably due to some package updates, which has changed the way they deal with RNG. I haven't found from which package it comes from, if someone has an idea. Anyway, I find the same final results on my branch than the one currently output on master.

Patch note:

I implemented `merging_two_signals` which much better shows the usefulness of the whole structure: let's examine it together:

during the `__getitem__`, `merging_two_signals` is applied. It merges the current signal with another random signal with the same label, randomly picked. Seems pretty simple, right ? But if you look closely, it implies that the function should have access to :

* the label `y` of the data,
* a dictionnary, the keys being the label, the values being the list of all indexes of data having this label,
* the train dataset, from where the data should be picked,
* a method to recover the raw data, because using `__getitem__` leads to.. the transformed data, creating a never-ending recursive loop.

The `Datum` object is here to store a reference to all theses variables. Otherwise, the transforms only has access to X and y. Most of the new complexity of WindowsDataset and WindowsConcatdataset are coming from the computation of `label_index_dict` and the new `get_raw_data` method to access unaugmented data. However, if you see a better way to implement it, let's discuss 👍 !


I also implemented an `AugmentedSubset` class, because you often need to see how your augmentation behave on smaller datasets, and the usual Subset method does not work any more on augmented datasets.

robintibor:
I will try have a look more deeply when I find time, from your description of merging_two_signals, it seems something that should not happen at a dataset-augmentation level I think as there should be like single example in, single example out to me. Seems something that should happen at sampler level to me. If one really insists on doing it at dataset level, one could give reference to dataset to the transformer in order to sample the other example. This way one can prevent adding more couplings/complexity inside core dataset classes code

gemeinl:
> One small question : I can integrate the `TransformDataset` class into the `WindowsDataset` class, but I had not seen that the `WindowsConcatDataset` was not existing. However, in `TransformConcatDataset`, I used to propagate a modification of the transforms into the `TransformDataset` objects. Since the notion of transform is only defined at the WindowsDataset level, I seemingly cannot simply reuse a `BaseConcatDataset`, and need to create at least a `WindowsConcatDataset` class (I though this one was already existing). Would it be ok ?

So far, we did not have a WindowsConcatDataset since our BaseConcatDataset can handle __both__ BaseDatasets and WindowsDatasets and the windowing functions transform the first to the latter. I think, I agree with @robintibor to not augment at a dataset level, if possible. We should try to implement augmentation at example level using transforms in the __getitem__ of the existing dataset classes. If the augmentation technique requires for more information than is accessible at this point, it could probably be implemented with an augmentation sampler class (https://pytorch.org/docs/stable/_modules/torch/utils/data/sampler.html#Sampler). Like this we can probably avoid actually changing our dataset classes and avoid adding WindowsConcatDataset and AugmentedSubset.

Simon-Free:
Ok, I restart fresh from the master's base.py file today, and found a way more simpler implementation of how to augment a dataset : only one new class, AugmentedDataset (the Datum class is hidden in), and only 25 lines of code 😄 ! I'll come back later with some comments and precisions about the framework.

Simon-Free:
Ok, let's explain and analyze step-by-step the new architecture. Some minimal example of use would be. 

```python
from braindecode.augmentation import Transform
from list_of_operations import op_a, op_b, op_c


unaugmented_dataset = get_dataset() # Usual braindecode WindowsDataset/BaseConcatDataset(WindowsDataset)
transform_list = [Transform(op_a), Transform(op_b), Transform(op_c)]

augmented_dataset = AugmentedDataset(unaugmented_dataset, transform_list)

y_train = np.array([data[1] for data in iter(augmented_dataset)])
clf.fit(augmented_dataset, y=y_train, epochs=model_args["n_epochs"])
```

AugmentedDataset is just a minimal overlay, as I think you wanted since the beginning and I misunderstood/could not find a way to implement, since I was incrementally building on what I did before. I do believe I created myself way too much problems by mixing the core functionnalities of WindowsDataset and the transform implementation. 

Getting deeper in the code, we have as an `__init__`
```python
class AugmentedDataset(Dataset):

    def __init__(self, ds, list_of_transforms=[Transform(identity)]) -> None:
        self.list_of_transforms = list_of_transforms
        self.ds = ds
        self.required_variables = self.initialize_required_variables()

    def initialize_required_variables(self):
        for transform in self.list_of_transforms:
            for key in transform.required_variables.keys():
                self.required_variables[key] = \
                    transform.required_variables[key](self.ds)
```
Here, the only nontrivial point is `required_variables` is especially made for variables like `label_index_dict` needed for certain complex transforms. The framework would be to define, for each transform, the **variable name** and the **variable function** that compute the needed variable given the unaugmented dataset. Example
```python
# The transform:
def merge_two_signals()
   # need datum.required_variables["label_index_dict"]

# It requires a variable : let's define the way it is computed:
def init_label_index_dict()

# Now associate it with the variable name
MERGE_TWO_SIGNALS_REQUIRED_VARIABLES = {
    "label_index_dict": init_label_index_dict} 

# Note: an attribute required_variables was previously added to the Transform class:
class Transform:
    def __init__(self, other_args, required_variables={})
        self.required_variables = required_variables

# Now one can define properly define the transform:
Transform(merge_two_signals, probability=..., required_variables=MERGE_TWO_SIGNALS_REQUIRED_VARIABLES)
```
All needed variables will be computed during the initialization of the AugmentedDataset.

Since you don't have to apply transforms at the WindowsDataset level, the `__getitem__` part is super simple :
```python

    def __getitem__(self, index):
        # get the index of the unaugmented image and the index of the transform
        tf_index = index % len(self.list_of_transforms)
        img_index = index // len(self.list_of_transforms)

       # get the unaugmented image
        X, y, crops_ind = (self.ds[img_index])

        # Define metadata class (probably not computationnaly optimal to define it here,
        # did this mainly for explainatory purposes
        class Datum:
            def __init__(self, X, y, crops_ind, ds, required_variables):
                self.X = X
                self.y = y
                self.crops_ind = crops_ind
                self.ds = ds
                self.required_variables = required_variables
        
        # Compute metadata
        transf_datum = self.list_of_transforms[tf_index](
            Datum(X, y, crops_ind, self.ds, self.required_variables))

        # Apply transform
        X, y, crops_ind = transf_datum.X, transf_datum.y, transf_datum.crops_ind

        return X, y, crops_ind
```

And.. that's it. That's probably much closer of what you originally wanted. 

**Remaining work & questions :**

* Does this framework seems convincing enough? At least the AugmentedDataset class.
* @robintibor I did not forget your remarks about the Transform class, which could be a StochasticTransform class. I did also not forget your concerns about magnitude. I'll rethink the architecture of Transform I believe.
* Should the transforms operations like `masking_along_frequency` or `merge_two_signals` be implemented in braindecode or outside ?

braindecode/augmentation/transform_class.py
agramfort: +1

braindecode/augmentation/transform_class.py
```diff
@@ -0,0 +1,29 @@
+from functools import partial
```
agramfort, line 1: author and license header missing

braindecode/augmentation/transform_class.py
```diff
@@ -0,0 +1,29 @@
...
+import numpy as np
```
agramfort, line 2: I would rename the file to maybe base.py

braindecode/augmentation/transforms/masking_along_axis.py
```diff
@@ -0,0 +1,138 @@
+import torch
```
agramfort, line 1: I would avoid a new nested directory here and put files in the augmentation module.
Doing

from braindecode.augmentation.transforms import ...

seems too much. 

also maybe braindecode.augment could be enough

braindecode/augmentation/transforms/masking_along_axis.py
```diff
@@ -0,0 +1,138 @@
...
+     ``uniform(0, magnitude * v_max)`` and rounded,
+    and ``v_start`` from ``uniform(0, v_max - v)`` and rounded.
+
+    Args:
+        X (Tensor): data
+        params (dict): a dict containing the necessary parameters for applying
+            the transform : the magnitude, the axis and the masking value.
+
+    Returns:
+        Tensor: Masked data
```
agramfort, line 50: your docstring formats do not follow the numpydoc standard.

see an example:

https://github.com/braindecode/braindecode/blob/master/braindecode/classifier.py#L190

with Parameters and Returns section.

you could run pydocstyle of your doc to check these things. https://pypi.org/project/pydocstyle/

braindecode/augmentation/transforms/masking_along_axis.py
```diff
@@ -0,0 +1,138 @@
...
+
+def time_frequency_to_signal(X):
+    """Transforms a time-frequency representation back into a signal
+    Args:
+        X (Tensor): Real spectrogram (channel, freq, time)
+
+    Returns:
+        Tensor: Temporal signal (channel, time)
+    """
+    global FFT_ARGS, DATA_SIZE
```
agramfort, line 132: global should be avoided

braindecode/augmentation/transforms/merge_two_signals.py
```diff
@@ -0,0 +1,38 @@
+import numpy as np
```
agramfort, line 1: missing header (author + license)

braindecode/augmentation/transforms/merge_two_signals.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def merge_two_signals(datum, magnitude):
+
```
agramfort, line 5: no docstring. Maybe should be private function?

braindecode/augmentation/transforms/merge_two_signals.py
```diff
@@ -0,0 +1,38 @@
...
+    datum.X = final_signal
+
+    return datum
+
+
+def init_label_index_dict(dataset):
+    """Create a dictionnary, with as key the labels available in the
+        multi-classification process, and as value for a given label
+        all indexes of data that corresponds to its label.
+        """
```
agramfort, line 23: idem

braindecode/datasets/base.py
```diff
@@ -16,6 +16,8 @@
 import pandas as pd
 
 from torch.utils.data import Dataset, ConcatDataset
+from braindecode.augmentation.transform_class import Transform
+from braindecode.augmentation.transforms import identity
```
agramfort, line 20: I feel that defining a public identity function that does `lambda x: x` is an overkill

braindecode/datasets/base.py
```diff
@@ -160,3 +166,39 @@ def split(self, by=None, property=None, split_ids=None):
         return {str(split_name): BaseConcatDataset(
             [self.datasets[ds_ind] for ds_ind in ds_inds])
             for split_name, ds_inds in split_ids.items()}
+
+
+class AugmentedDataset(Dataset):
+
```
agramfort, line 172: no docstring :(

braindecode/datasets/base.py
```diff
@@ -160,3 +166,39 @@ def split(self, by=None, property=None, split_ids=None):
...
+                self.ds = ds
+                self.required_variables = required_variables
+
+        transf_datum = self.list_of_transforms[tf_index](
+            Datum(X, y, crops_ind, self.ds, self.required_variables))
+        X, y, crops_ind = transf_datum.X, transf_datum.y, transf_datum.crops_ind
+
+        return X, y, crops_ind
+
+    def initialize_required_variables(self):
```
agramfort, line 200: does it have to be public?

braindecode/datasets/sleep_physionet.py
```diff
@@ -1,18 +1,21 @@
-# Authors: Hubert Banville <hubert.jbanville@gmail.com>
-#
-# License: BSD (3-clause)
```
agramfort, line 3: hum ...

braindecode/datasets/sleep_physionet.py
```diff
@@ -99,3 +106,151 @@ def _load_raw(raw_fname, ann_fname, preload, load_eeg_only=True,
...
+        recording=[1],
+        crop_wake_mins=0,
+        preprocessing=preprocessing)
+    # for i in range(len(train_sample)):
+    #     train_sample[i] = (train_sample[i][0][:50], train_sample[i][1],
+    #                        train_sample[i][2])
+    # for i in range(len(test_sample)):
+    #     test_sample[i] = (test_sample[i][0][:50],
+    #                       test_sample[i][1], test_sample[i][2])
+    # print("lol")
```
agramfort, line 126: cleanup

braindecode/datasets/sleep_physionet.py
```diff
@@ -99,3 +106,151 @@ def _load_raw(raw_fname, ann_fname, preload, load_eeg_only=True,
...
+    for i in subset_dict.keys():
+        concat_dataset.datasets[i] = take_dataset_subset(
+            concat_dataset.datasets[i], subset_dict[i])
+    concat_dataset.cumulative_sizes = concat_dataset.cumsum(
+        concat_dataset.datasets)
+    return(concat_dataset)
+
+# Authors: Hubert Banville <hubert.jbanville@gmail.com>
+#
+# License: BSD (3-clause)
```
agramfort, line 256: :)

braindecode/datasets/sleep_physionet.py
```diff
@@ -99,3 +106,151 @@ def _load_raw(raw_fname, ann_fname, preload, load_eeg_only=True,
...
+        windows_dataset.y = windows_dataset.y[indice_list]
+        windows_dataset.crop_inds = windows_dataset.crop_inds[indice_list]
+        return(windows_dataset)
+
+    for i in subset_dict.keys():
+        concat_dataset.datasets[i] = take_dataset_subset(
+            concat_dataset.datasets[i], subset_dict[i])
+    concat_dataset.cumulative_sizes = concat_dataset.cumsum(
+        concat_dataset.datasets)
+    return(concat_dataset)
```
agramfort, line 252: I don't feel this is library code. It's utils for your own experiments on sleep data. I would not be part of the core library

test/acceptance_tests/test_augmented_training.py
```diff
@@ -0,0 +1,77 @@
+from braindecode.augmentation.transforms.masking_along_axis import \
+    mask_along_frequency, mask_along_time
+import torch
+from skorch.callbacks import LRScheduler
+from braindecode import EEGClassifier
+from braindecode.util import set_random_seeds
+from braindecode.models import SleepStagerChambon2018
+from braindecode.datasets.sleep_physionet import get_dummy_sample
+from braindecode.augmentation.transform_class import Transform
+from braindecode.augmentation.transforms.identity import identity
```
agramfort, line 10:
```suggestion
from braindecode.augment import Transform, identity
```
?

test/acceptance_tests/test_augmented_training.py
```diff
@@ -0,0 +1,77 @@
+from braindecode.augmentation.transforms.masking_along_axis import \
```
agramfort, line 1: header missing

test/acceptance_tests/test_augmented_training.py
```diff
@@ -0,0 +1,77 @@
...
+    mask_along_frequency, mask_along_time
```
agramfort, line 2: wrong import order

test/unit_tests/augmentation/test_transforms.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+import torch
+from braindecode.augmentation.transform_class import Transform
+from braindecode.augmentation.transforms.masking_along_axis \
+    import mask_along_axis_random
```
agramfort, line 5:
```suggestion
from braindecode.augment import mask_along_axis_random
```
?

agramfort (commented):
some feedback

sbbrandt:
Hello Simon,

thanks for pushing the data augmentation part to braindecode.

I have currently work in progress on the same topic, as I am doing some experiments with data augmentation. I have not yet looked completely through your PR, but maybe my code helps here.

My attempt was to modify the DataLoader and manipulate the data on the batch level. That way, I don't need a new dataset class, but I can simply pass a custom training iterator to the classifier.

You can find my code here (this is code in progress and not yet completed):

https://github.com/sbbrandt/braindecode-1/blob/data_augmentation/examples/plot_data_augmentation.py

Best,

Simon

robintibor:
Hi, since we have multiple people working on somewhat similar things here, maybe good to get a bit more input.
I see two main functionalities wrt to data augmentation:

1) Single example in/single example out , e.g. adding noise to an example
2) Mixing multiple inputs, possibly depending on label or relative time position etc.

To me, I would expect 1) is easy to implement at dataset level by calling transforms in __getitem__ 
2) seems the more difficult part, and one can have multiple ways of doing it, I guess mostly either pushing it more towards dataloader level or staying inside dataset level

Since I myself don't have full overview over the possibilities and pro/cons here, would be nice if @sbbrandt @hubertjb could summarize a bit in your code, which cases (of 1 an 2) do you cover and how are they in general implemented? (e.g., I saw in PR of @hubertjb #178  you also need to create specialized dataset?).
Maybe then we can see more clearly what would be the best way

agramfort:
if feels we need to organize a short call...

>

robintibor:
yes but would be nice to have something written here first before call to me. understanding every approach in call may be too hectic, at least for me

agramfort:
ok fair enough

>

Simon-Free:
Hey @sbbrandt, thank you a lot for sharing your work ! As an example, what your Mixup transform you tried would give in my framework, if I understood well what you wanted to do (not sure what the Mixup Loss was used for, since you already send a mixup label `y`)

without the `beta_per_sample`, this is very simple:
```python
def mixup_beta(datum, params):

    alpha = params["alpha"]
    if alpha > 0:
        lam = np.random.beta(alpha, alpha)
    else:
        lam = 1
    other_data_idx = np.random.randint(len(datum.ds))
    other_data = datum.ds[other_data_idx]
    datum.X = lam * datum.X + (1 - lam) * other_data[0]
    datum.y = lam * datum.X + (1 - lam) * other_data[1]
    return datum

transform_list = Transform(mixup_beta, params={alpha: 0.4})
aug_dataset = AugmentedDataset(train_set, transform_list)
```

And here we go ! 

The "beta_per_sample" seems a bit trickier, since it looks like you change the beta constant at every epoch, and we don't have a direct access to this information. However, it is still implementable, and looks like that. Probably not the prettiest implementation but:

```python
def mixup_two_signal(datum, params):

    alpha = params["alpha"]
    if "epoch_counter" not in params.keys():
        params["acc_in_epoch"] = 0
    if alpha > 0:
        if params["beta_per_sample"]:
            lam = np.random.beta(alpha, alpha)
        else:
            if not (params["acc_in_epoch"] %
                    datum.required_variable["true_len_of_ds"]):
                params["beta"] = np.random.beta(alpha, alpha)
                lam = params["beta"]
            else:
                lam = params["beta"]
            params["acc_in_epoch"] += 1
    else:
        lam = 1
    other_data_idx = np.random.randint(len(datum.ds))
    other_data = datum.ds[other_data_idx]
    datum.X = lam * datum.X + (1 - lam) * other_data[0]
    datum.y = lam * datum.X + (1 - lam) * other_data[1]
    return datum

MIXUP_BETA_REQUIRED_VARIABLES = {"true_len_of_ds": lambda dataset, transf_list:
                                 len(dataset) * len(transf_list)}
transform_list = Transform(mixup_beta, params={alpha: 0.4}
                           required_variables = MIXUP_BETA_REQUIRED_VARIABLES)
aug_dataset = AugmentedDataset(train_set, transform_list)
```
Does the dataloader approach generalizes well too ? 🤔 It seems tricky if you need to also adapt the loss, but maybe I just don't know enough pytorch 😄 !

I had though to make two generalizations of my framework to implement it : first, replacing the `magnitude` by a more general `params` eventually containing more informations, and to give as argument for required_variables not only the dataset but also the transform list. But i do think now that the most complex cases of augmentation are pretty consistently handlables.

Simon-Free:
Ok, I understood alone the issue.. 'll work on it ! 😄

Simon-Free:
Ok, solved and generalized. First, adding:

```python
class AugmentedDataset(Dataset):
   def __getitem__(self, index):
        X, y, crops_ind = (self.ds[img_index])
        y = {y: 1} # A dict {label: proportion}
```

Second, changing the mixup transform with:
```python
def mixup_two_signal(datum, params):
    
    datum.X = lam * datum.X + (1 - lam) * other_data[0]
    for label in datum.y:
        datum.y[label] *= lam
    if other_data[1] in datum.y
        datum.y[other_data[1]] += (1-lam)
    else:
        datum.y[other_data[1]] = (1-lam)
    return datum
```

Third, changing the loss with the more general:
```python
import torch

class general_mixup_criterion:

    def __init__(self, loss=torch.nn.functional.nll_loss):
        self.loss = loss

    def __call__(self, preds, target):
        return self.loss_function(preds, target)
    
    def loss_function(self, preds, target):
        ret = None
        for label in target.keys():
            prop = target[label]
            loss_val = self.loss(preds, label, reduction='none')
            if ret is None:
                ret = torch.mul(prop, loss_val)
            else:
                ret += torch.mul(prop, loss_val)
        return ret.mean()
```

Now you can not only mixup, but compose it with other transforms, even remixup after, with a third, a fourth data with another label. Not yet commited and tested, but will come soon.

hubertjb:
> Since I myself don't have full overview over the possibilities and pro/cons here, would be nice if @sbbrandt @hubertjb could summarize a bit in your code, which cases (of 1 an 2) do you cover and how are they in general implemented? (e.g., I saw in PR of @hubertjb #178 you also need to create specialized dataset?).
> Maybe then we can see more clearly what would be the best way

@robintibor (I haven't followed the whole discussion here so I hope this is relevant) in the case of #178 I wasn't thinking of the window pairs extraction as a data augmentation transform per se. I implemented a `Sampler` class which randomly chooses two windows, and returns their indices. I then needed to adapt the `BaseConcatDataset` class so that `__getitem__` could receive 2  indices instead of 1 and return the corresponding windows. I guess this makes it more similar to the second case you're talking about, but again I'm not sure I would call this data augmentation.

Simon-Free:
Ok, pytorch is not really dictionary-friendly. Will find a better way to deal with the issue tomorrow (well, today...).

Simon-Free:
Finished the implementation, everything should work as expected. Administrative stuff such as docstrings & authoring missing, but the core methods are there. Gonna make them, comment, code one or two examples, but I finish my internship tomorrow and clearly won't have as much time as now to work on it, so if you have any requests, questions, remarks, objections, etc., feel free to ask them now 😉

sbbrandt:
> Hi, since we have multiple people working on somewhat similar things here, maybe good to get a bit more input.
> I see two main functionalities wrt to data augmentation:
> 
>     1. Single example in/single example out , e.g. adding noise to an example
> 
>     2. Mixing multiple inputs, possibly depending on label or relative time position etc.
> 
> 
> To me, I would expect 1) is easy to implement at dataset level by calling transforms in `__getitem__`
> 2) seems the more difficult part, and one can have multiple ways of doing it, I guess mostly either pushing it more towards dataloader level or staying inside dataset level
> 
> Since I myself don't have full overview over the possibilities and pro/cons here, would be nice if @sbbrandt @hubertjb could summarize a bit in your code, which cases (of 1 an 2) do you cover and how are they in general implemented? (e.g., I saw in PR of @hubertjb #178 you also need to create specialized dataset?).
> Maybe then we can see more clearly what would be the best way

For my experiment with data augmentation I needed to manipulate data on the batch level. I.e. I wanted to draw a random number each batch and apply a transformation depending on that random number on the whole batch.

I had issues, respectively found it too complicated, to implement that on the dataset level. As far as I know, for each batch, the __get_item__ function is called on every sample of the batch, such that any transform relying on repeatedly drawn random numbers, gets one random number per sample. My solution was to pass the transfrom as the collate function ('collate_fn') to the dataloader. That way, the transform is applied on the batch level. In my mixup example, I also provide a custom loss function, as I need to deentangle the mixup for my loss criterion.

Of course there are many other data augmentation methods that can (or even have to) be applied on the dataset level.

braindecode/augment/base.py
```diff
@@ -0,0 +1,238 @@
...
+            return(self.loss(preds, target, reduction='mean'))
+
+
+class Compose:
+    """Composes several transforms together.
+
+        Parameters
+        ----------
+        transforms: list(Transform)
+            list of transforms to compose.
```
tomMoral, line 229: extra indent to remove

braindecode/augment/transforms/merge_two_signals.py
```diff
@@ -0,0 +1,46 @@
...
+
+    signal = datum.X
+    y = datum.y
+    train_sample = datum.ds
+    label_index_dict = datum.required_variables["label_index_dict"]
+    other_signal = np.zeros(datum.X.shape)
+    for label_idx in range(len(y)):
+        # y[label] is the proportion of the label in y
+        other_signal_index = np.random.choice(
+            label_index_dict[datum.list_of_labels[label_idx]])
```
tomMoral, line 18: I would rather have the closing parenthesis `)` on the next line.

braindecode/augment/transforms/merge_two_signals.py
```diff
@@ -0,0 +1,46 @@
...
+        signal + params["magnitude"] * other_signal
+
+    return datum
+
+
+def init_label_index_dict(dataset, transform_list):
+    """Create a dictionnary, with as key the labels available in the
+        multi-classification process, and as value for a given label
+        all indexes of data that corresponds to its label.
+        """
```
tomMoral, line 31: extra indent?

braindecode/augment/transforms/merge_two_signals.py
```diff
@@ -0,0 +1,46 @@
...
+    for label in list_labels:
+        label_index_dict[label] = []
+    for i in range(len(subset_unaug_indices)):
+        label_index_dict[subset_unaug_labels[i]].append(
+            subset_unaug_indices[i])
+    return label_index_dict
+
+
+MERGE_TWO_SIGNALS_REQUIRED_VARIABLES = {
+    "label_index_dict": init_label_index_dict}
```
tomMoral, line 46: `}` on the next line

braindecode/datasets/base.py
```diff
@@ -160,3 +165,21 @@ def split(self, by=None, property=None, split_ids=None):
...
+            list of transforms to compose.
+    """
+
+    def __init__(self, transforms):
+        self.transforms = transforms
+
+    def __call__(self, datum):
+        for t in self.transforms:
+            datum = t(datum)
+        return datum
```
tomMoral, line 185: Already defined in `augment`.

tomMoral (commented):
Some comments

robintibor:
Summary of what I understand so far.

Pytorch data loading pipeline:

`Datasets`:
Supply single dataset elements via `__getitem__`. (in case of "map-style" datasets https://pytorch.org/docs/stable/data.html)

Input of `__getitem__`:
An index - typically an integer, although could be arbitrary in principle

Output of `__getitem__`:
A tuple describing an element of the dataset - typically a datapoint and a label

`DataLoaders`:
Supply batches of elements from the dataset (typically), with using multithreading per default.

It works by getting a list of indices from a sampler, retrieving the corresponding elements of the dataset via `__getitem__` and then making a batch out of it using collate_fn.

Sampler should be an iterable supplying indices.

Therefore possible strategies for data augmentation:

1) Directly in `__getitem__` of `Dataset`.
torchaudio and torchvision both follow this by allowing user to supply callables transform and target_transform which get called directly with waveform/image and label and can modify it arbitrarily. This takes advantage of multithreading. It is logically to me most clear for transforms that only work on the concrete element of the dataset (do not mix multiple elements).
torchaudio example: https://github.com/pytorch/audio/blob/551750031e0797b68af81a89dffd08aad89abcdd/torchaudio/datasets/vctk.py#L155-L159
torchvision example: https://github.com/pytorch/vision/blob/6db1569c89094cf23f3bc41f79275c45e9fcb3f3/torchvision/datasets/cifar.py#L119-L123
This is done in this PR.

2) Inside `collate_fn` of `Dataloader` 
In the `collate_fn` you have access to the elements in the batch as returned by accessing indices supplied by the sampler.
For mixup collate_fn is enough, as done here in sbbrandt fork https://github.com/sbbrandt/braindecode-1/blob/9adf8b7d62eaf8f6954947690a61c3a0f413ad45/braindecode/datautil/iterators.py
If one wants to do something more involved, e.g., pick the signals to merge according to some criterion that relies on the dataset and cannot be ensured just from the random elements in a batch, one also needs to modify the sampler such that the needed samples are there to merge in the batch. I guess this could have been another way to implement your self-supervised examples @hubertjb? So concretely, by keeping dataset as original, writing a batch sampler that ensures whatever examples are wanted are in the same batch and then putting them in the right way inside `collate_fn`. Assume this would be more error-prone way as there would be hard-to-enforce expectations between what the sampler does and what the `collate_fn` expects. 
But overall, just to me seems important that the `collate_fn` way may be flexible enough to also over other use cases than mixup.


Finally, of course one can also perform some data augmentations completely in separate code / "by hand", e.g. as here in original mixup code. For this one would need to override train_step of skorch I assume
https://github.com/facebookresearch/mixup-cifar10/blob/eaff31ab397a90fbc0a4aac71fb5311144b3608b/train.py#L148-L172

Overall, I would like to think a bit more about it, but right now would lean towards:
1) Have both ways of transforms, element-level and batch-level.
2) Element-level, keep simple, so like in torchaudio/torchvision, simply element in/out. main thing we have to see whether we supply some mne object there into the transform (probably)
3) Batch-level, first keep simple to cover mixup-case, see whether further extension is needed to cover other cases in the future

So @Simon-Free  from your comments I gather you may not have much time to work on this anymore? So how should we proceed, we should have rights to also push into this PR, so we could just continue from current state? Or how is it from your side?

Simon-Free:
Yes, don't hesitate to commit & edit my PR :) ! 

I think the understanding may cover how we do it, but also what we want to do: 
refering to the main papers on data augmentation for computer vision (listed below), the ideal requirements would be:

- Applying several transform on a dataset: it means, for each epochs, if you have `n` data, you should be able to implement `s` transform so that at each epochs your dataset sees `n * s` data, each one corresponding to the transform `i` applied on the data `j`.
- Composing transforms
- Applying a transform with a given probability.
- Eventually, applying a "schedule", i.e. a list of transforms that varies depending on the epoch you are in.

(Papers: [Fast AutoAugment](https://arxiv.org/abs/1905.00397), [AutoAugment](https://arxiv.org/abs/1805.09501), [Population Based Augmentation](https://arxiv.org/pdf/1905.05393.pdf))

I would not think that two ways to do transforms are the best way to go, mainly for API complexity purposes, I would either go the `Dataset` or the `Dataloader` approach but not both, and would like to underline that both can do the job, each one having their advantages and drawbacks. I would consider the `Dataset` way easier to implement and understand, but the `Dataloader` seems to have some advantages, like an easy access to some variables and a way to easily see at which epoch one is.

robintibor:
Now already superseded by on-the-fly transforms in https://github.com/braindecode/braindecode/pull/198 , closing

PR #178, opened 2020-11-04
[WIP] Self-supervised learning example on Sleep Physionet

hubertjb:
This PR introduces a new example, along with a new `samplers` module, to demonstrate how self-supervised learning can be used to learn representations in an unsupervised manner on raw EEG. The example implements the relative positioning task described in this [preprint](https://arxiv.org/abs/2007.16104).   

I've kept the number of recordings used in the example as well as the number of self-supervised samples and training epochs pretty low, so the example could run relatively quickly. The results are already interesting in that setting, although nicer results are obtained when more examples are included. For instance, with 40 recordings:

rp_learning_curves_40subjects
rp_pca_40subjects

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -161,13 +161,8 @@ class BaseConcatDataset(ConcatDataset):
             [self.datasets[ds_ind] for ds_ind in ds_inds])
             for split_name, ds_inds in split_ids.items()}
 
-    @property
-    def metadata(self):
+    def get_metadata(self):
         """Concatenate the metadata and description of the wrapped Epochs.
-
-        NOTE: This is implemented as a property to avoid having to keep a very
-              large DataFrame in case the dataset contains very long or many
-              recordings.
 
         Returns
         -------
@@ -177,7 +172,7 @@ class BaseConcatDataset(ConcatDataset):
             for each window.
         """
         if not all([isinstance(ds, WindowsDataset) for ds in self.datasets]):
-            raise TypeError('Property metadata can only be computed when all '
+            raise TypeError('Metadata dataframe can only be computed when all '
                             'datasets are WindowsDataset.')
 
         all_dfs = list()
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -160,3 +160,31 @@ class BaseConcatDataset(ConcatDataset):
         return {str(split_name): BaseConcatDataset(
             [self.datasets[ds_ind] for ds_ind in ds_inds])
             for split_name, ds_inds in split_ids.items()}
+
+    @property
+    def metadata(self):
+        """Concatenate the metadata and description of the wrapped Epochs.
+
+        NOTE: This is implemented as a property to avoid having to keep a very
+              large DataFrame in case the dataset contains very long or many
+              recordings.
+
+        Returns
+        -------
+        pd.DataFrame:
+            DataFrame containing as many rows as there are windows in the
+            BaseConcatDataset, with the metadata and description information
+            for each window.
+        """
+        if not all([isinstance(ds, WindowsDataset) for ds in self.datasets]):
+            raise TypeError('Property metadata can only be computed when all '
+                            'datasets are WindowsDataset.')
+
+        all_dfs = list()
+        for ds in self.datasets:
+            df = ds.windows.metadata
+            for k, v in ds.description.items():
+                df[k] = v
+            all_dfs.append(df)
+
+        return pd.concat(all_dfs)
--- a/braindecode/datasets/sleep_physionet.py
+++ b/braindecode/datasets/sleep_physionet.py
@@ -23,13 +23,14 @@ class SleepPhysionet(BaseConcatDataset):
 
     Parameters
     ----------
-    subject_ids: list(int) | int
-        (list of) int of subject(s) to be loaded.
-    recording_ids: list(int)
+    subject_ids: list(int) | int | None
+        (list of) int of subject(s) to be loaded. If None, load all available
+        subjects.
+    recording_ids: list(int) | None
         Recordings to load per subject (each subject except 13 has two
-        recordings). Can be [1], [2] or [1, 2].
+        recordings). Can be [1], [2] or [1, 2] (same as None).
     preload: bool
-        if True, preload the data of the Raw objects.
+        If True, preload the data of the Raw objects.
     load_eeg_only: bool
         If True, only load the EEG channels and discard the others (EOG, EMG,
         temperature, respiration) to avoid resampling the other signals.
@@ -45,7 +46,8 @@ class SleepPhysionet(BaseConcatDataset):
         if recording_ids is None:
             recording_ids = [1, 2]
 
-        paths = fetch_data(subject_ids, recording=recording_ids)
+        paths = fetch_data(
+            subject_ids, recording=recording_ids, on_missing='warn')
 
         all_base_ds = list()
         for p in paths:
--- a/braindecode/samplers/base.py
+++ b/braindecode/samplers/base.py
@@ -24,7 +24,7 @@ class RecordingSampler(Sampler):
 
     Attributes
     ----------
-    info : pd.Series
+    info : pd.DataFrame
         Series with multiindex, which contains the subject, session, run and
         window indices information in an easily accessible structure for
         quickly sampling.
@@ -39,8 +39,8 @@ class RecordingSampler(Sampler):
                 if k in self.metadata.columns]
         if not keys:
             raise ValueError(
-                'metadata must contains at least one of the following columns:'
-                ' subject, session or run.')
+                'metadata must contain at least one of the following columns: '
+                'subject, session or run.')
 
         self.metadata = self.metadata.reset_index().rename(
             columns={'index': 'window_index'})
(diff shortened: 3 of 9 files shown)


hubertjb:
It looks like the failed CI is caused by differences in the recorded loss of other tests (`test_cropped_decoding.py` etc.), similar to what happened some time ago due to a change in skorch version. I'm not sure there's a link with the changes of this PR.

In any case, the rendered example can be found [here](https://562-232335424-gh.circle-artifacts.com/0/dev/index.html)!

braindecode/datasets/base.py
```diff
@@ -160,3 +160,31 @@ def split(self, by=None, property=None, split_ids=None):
...
+                            'datasets are WindowsDataset.')
+
+        all_dfs = list()
+        for ds in self.datasets:
+            df = ds.windows.metadata
+            for k, v in ds.description.items():
+                df[k] = v
+            all_dfs.append(df)
+
+        return pd.concat(all_dfs)
```
robintibor, line 190: I guess this is the critical part of this PR with regards to discussing. I am not sure I have a good mental overview over the dataset part. so metadata here will be concatenated from windows datasets metadata as far as I see. 
Regarding implementation as property, could be fine, I am a bit worried as I don't think we use properties in other places, and it can always cause some confusion if there is non-neglible execution time on access to a member (meaning looks like access to a member from caller side). But as said, could be fine.

@gemeinl do you think this metadata here fits in nicely with rest of data structures as you see them? or do you envision any potential conflicts/confusions?

braindecode/samplers/base.py
```diff
@@ -0,0 +1,69 @@
...
+
+class RecordingSampler(Sampler):
+    """Base sampler simplifying sampling from recordings.
+
+    Parameters
+    ----------
+    metadata : pd.DataFrame
+        DataFrame describing at least one of {subject, session, run}
+        information for each window in the BaseConcatDataset to sample examples
+        from. Normally obtained as the `metadata` propery of BaseConcatDataset.
```
robintibor, line 21: typo on "propery".
Also this is still not so clear to me, maybe you an also put an minimal example in the docstring? Does not have be something that runs by itself, like just show relevant lines for understanding what is in metadata and what would happen.

robintibor:
> It looks like the failed CI is caused by differences in the recorded loss of other tests (`test_cropped_decoding.py` etc.), similar to what happened some time ago due to a change in skorch version. I'm not sure there's a link with the changes of this PR.
> 
> In any case, the rendered example can be found [here](https://562-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_relative_positioning.html#sphx-glr-auto-examples-plot-relative-positioning-py)!

Hm does it not help to rebase from master? like I thought the problem is solved now.

robintibor:
In any case thanks for sharing self-supervised code, really appreciate it!

braindecode/datasets/base.py
hubertjb: I agree with you, this part requires some discussion. I guess this could instead be a method `get_metadata()`, which would make the fact that this requires computations more explicit.

hubertjb:
> > It looks like the failed CI is caused by differences in the recorded loss of other tests (`test_cropped_decoding.py` etc.), similar to what happened some time ago due to a change in skorch version. I'm not sure there's a link with the changes of this PR.
> > In any case, the rendered example can be found [here](https://562-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_relative_positioning.html#sphx-glr-auto-examples-plot-relative-positioning-py)!
> 
> Hm does it not help to rebase from master? like I thought the problem is solved now.

It looks like it's already based on master actually. I haven't looked yet but maybe there's been another skorch update in the meantime...?

robintibor:
> > > It looks like the failed CI is caused by differences in the recorded loss of other tests (`test_cropped_decoding.py` etc.), similar to what happened some time ago due to a change in skorch version. I'm not sure there's a link with the changes of this PR.
> > > In any case, the rendered example can be found [here](https://562-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_relative_positioning.html#sphx-glr-auto-examples-plot-relative-positioning-py)!
> > 
> > 
> > Hm does it not help to rebase from master? like I thought the problem is solved now.
> 
> It looks like it's already based on master actually. I haven't looked yet but maybe there's been another skorch update in the meantime...?

yeah since in sliwys PR the same problem, feel free to ignore. we will have to look at it separately again :(

braindecode/datasets/base.py
robintibor: @gemeinl can maybe also give is view tomorrow on metadata vs get_metadata() tomorrow. I am not sure. metadata more consistent with windows.metadata, get_metadata() more explicit that there is a computation.... what does yoda @agramfort think?

robintibor:
rebasing on master should now make all tests run again @hubertjb. Seems some strange travis UI delays atm, but I received a mail that they pass.
Then you could also fix typo, and we can merge once decision about metadata vs get_metadata() is done (within next few days?).

agramfort:
>
> > what does yoda @agramfort <https://github.com/agramfort> think?
>

😂

+1 to avoid properties is they involved computations unless you cache the
results

hubertjb:
@robintibor I just pushed the latest changes, including changing the `metadata` property to a `get_metadata` method, and making sure that the model that is used for visualization is the one that achieved the best valid loss.

gemeinl:
I am asking myself if it is necessary or good that the dataset implements ```get_metadata```. From what I see, all the information is accessible from the outside, ```dataset.description``` as well as ```dataset.windows.metadata``` so we could just move the code into the ```RecordingSampler``` and ```RelativePositionSampler```. This would avoid all the problems regarding class properties or naming and could also avoid potential confusion about the term metadata and what it is. Any thoughts on this?

braindecode/datasets/sleep_physionet.py
```diff
@@ -23,13 +23,14 @@ class SleepPhysionet(BaseConcatDataset):
...
+        (list of) int of subject(s) to be loaded. If None, load all available
+        subjects.
+    recording_ids: list(int) | None
         Recordings to load per subject (each subject except 13 has two
-        recordings). Can be [1], [2] or [1, 2].
+        recordings). Can be [1], [2] or [1, 2] (same as None).
     preload: bool
-        if True, preload the data of the Raw objects.
+        If True, preload the data of the Raw objects.
     load_eeg_only: bool
```
gemeinl, line 34: Is seems that this is currently not used. Do you think it is necessary to have this option? It looks like a shortcut to avoid using ```mne.Raw.drop_channels``` or ```mne.Raw.pick``` / ```mne.Raw.pick_channels``` / ```mne.Raw.pick_types``` in a preprocessing step.

robintibor:
> I am asking myself if it is necessary or good that the dataset implements `get_metadata`. From what I see, all the information is accessible from the outside, `dataset.description` as well as `dataset.windows.metadata` so we could just move the code into the `RecordingSampler` and `RelativePositionSampler`. This would avoid all the problems regarding class properties or naming and could also avoid potential confusion about the term metadata and what it is. Any thoughts on this?

Hm to me `get_metadata` seems still ok to me, like it makes sense to have in the class, since it uses a lot of the class members, and to me logically makes sense to be there... also does not introduce any additional changes outside the new function itself...

braindecode/datasets/sleep_physionet.py
gemeinl: Ok, nevermind. It is used and channels are already exluded on read from HDD.

hubertjb:
> > I am asking myself if it is necessary or good that the dataset implements `get_metadata`. From what I see, all the information is accessible from the outside, `dataset.description` as well as `dataset.windows.metadata` so we could just move the code into the `RecordingSampler` and `RelativePositionSampler`. This would avoid all the problems regarding class properties or naming and could also avoid potential confusion about the term metadata and what it is. Any thoughts on this?
> 
> Hm to me `get_metadata` seems still ok to me, like it makes sense to have in the class, since it uses a lot of the class members, and to me logically makes sense to be there... also does not introduce any additional changes outside the new function itself...

@gemeinl I think it makes sense to have BaseConcatDataset return a `metadata` dataframe directly as it's pretty much what we're already doing with the `description` attribute (the `description` attribute of the individual `WindowsDatasets` are also concatenated into a single dataframe, also called `description`). I think there are other scenarios in which it might be convenient to have access to this concatenated `metadata` DataFrame directly too, e.g., for splitting datasets, converting to a (X,y) dataset, sanity checking, etc. For that reason I would at least implement `get_metadata` outside of the Sampler objects so it can be easily reused down the line.

robintibor:
ok, will merge this!

Keep in mind:
1) If we agree on some specific data augmentation implementation 
AND 
2) this way of implementing augmentations can also very nicely cover the case here (I am aware this is not same as augmentation here)
AND
3) it is different from what you have done

We may want to adapt the code at that point.

But until then:
https://twitter.com/malonebarry/status/1325284260777955328 :+1: :100: :partying_face: :smiley:

agramfort:
🍾 🎉 🍺

hubertjb:
Awesome, thanks @robintibor! :)

Yes, I agree, if this can be implemented with data augmentations it will be worth it to look into changing the implementation of this example. To keep in mind though, I think using a `Sampler` object is pretty natural and intuitive here, so this could be a reason to keep this implementation.

PR #184, opened 2020-11-16
Fix trial stop offset samples 

gemeinl:
In ```creating_windows_from_events```, ```trial_stop_offset_samples``` was not considered when ```window_size``` was not specified.
So far, our tests did not cover this case at all and did not test for the size of the resulting windows.


robintibor:
great!

PR #188, opened 2020-12-14
[MRG] Load TUH EEG Corpus

gemeinl:
For very large datasets (as the TUH EEG Corpus), the data structure within braindecode is insufficient, since creating instances of mne.io.RawArray requires too much RAM. With this PR I want to add
- Code that builds an index of the .edf files without creating mne.io.RawArrays. Files will be loaded on `__getitem__` to allow for preprocessing. After preprocessing compute windows should be stored as .fif files.
- Code that builds an index of the .fif files. Individual compute windows will be loaded on `__getitem__` to build batches for training

Have a look at this notebook for an example: 
https://gist.github.com/gemeinl/32796028e26a3a93b03bfbbdf19bf0f0

Missing/unfinished/to improve:
- Tests, especially the backtracking of compute window ids to recording ids
- Interaction with existing base dataset classes
- Check if code duplicates with TUHAbnormal can be avoided


hubertjb:
Very cool @gemeinl! 

So it looks like with datasets the size of TUH MNE uses too much memory... Do you know what objects take up that much space even when `preload=False`? I'm not familiar enough with the internal implementations of MNE base classes to know, but maybe there could be some optimization on MNE's side that partially solves this problem?

braindecode/datasets/base.py
```diff
@@ -183,3 +185,61 @@ def get_metadata(self):
...
+    return file_paths
+
+
+class LazyDataset(Dataset):
+    """A class that loads stored compute windows when getitem is called.
+
+    Params
+    ------
+    path: str
+        Parent directory of the .fif files serialized with braindecode.datautil.serialization.save_concat_dataset
```
agramfort, line 219: please avoid lines with more than 80c

agramfort (commented):
do you have a specific question on MNE?

gemeinl:
Here are some new observations:
I only encounter memory issues with raw EDFs. There, I observe ~3.3 GB of RAM usage when I read about ~1500 recordings with `read_raw_edf` and `preload=False`. However, since preprocessing like filtering will be too slow to do it "on-the-fly" we would need to store a preprocessed version of the data to disk anyways.

I observe < 100 MB of RAM usage when I read ~1250 recordings that survived preprocessing with `read_epochs` and `preload=False`. So we should be able to use quite a big amount of data before this becomes a problem.

@agramfort 
Is this massive difference of RAM usage of `read_raw_edf`and `read_epochs` expected?
In both cases, it seems that RAM usage scales with the duration of the recording although `preload=False` (see plots below). Is this expected?
image
image

As I see it right now, we can stick to the existing way of loading preprocessed datasets with `load_concat_dataset` and `preload=False` and do not need to introduce new data structures due to RAM limitations. Thanks to @hubertjb we already have runtime observations which suggest that this will also be fast enough for training. It would probably still be good to re-evaluate and verify this on a large dataset.

agramfort:
it's not normal what you observe on raw.

can you share a script using an mne dataset so I can easily replicate?

>

gemeinl:
Below is a code snippet I use to track RAM usage from within python:

```
import os
import mne
mne.set_log_level("WARNING")

import psutil


def mem_increase(func):
    def wrapper(*args, **kwargs):
        # https://stackoverflow.com/questions/276052/how-to-get-current-cpu-and-ram-usage-in-python
        process = psutil.Process(os.getpid())
        mem_before = process.memory_info().rss / 1024**2
        return_value = func(*args, **kwargs)
        mem_after = process.memory_info().rss / 1024**2
        mem_usage = mem_after - mem_before
        print(f"RAM usage increased by {mem_usage:.2f} MB")
        return return_value
    return wrapper


paths = mne.datasets.sleep_physionet.age.fetch_data(subjects=[1], recording=[1])
raw = mem_increase(mne.io.read_raw_edf)(paths[0][0], preload=False)
```

If I run this, I get the following output:
image
For me one of the questions is why MNE shows ~19kB when psutil states 65 MB. 


I guess, the effect shows better with many recordings. So if you have access to the TUH Abnormal EEG Corpus (that was used by David and Denis), you could run the following snippet:
```
import glob


def read_raws(paths, preload):
    return [mne.io.read_raw_edf(path, preload=False) for path in paths]


TUH_PATH = "/data/schirrmr/gemeinl/tuh-abnormal-eeg/raw/"
file_paths = glob.glob(
    os.path.join(TUH_PATH, "**/*.edf"), 
    recursive=True
)

raws = mem_increase(read_raws)(file_paths, preload=False)
```
For the ~3000 recordings of the dataset, I get the following output:
image
In addition, htop shows ~8500 MB usage for this process.

Please let me know if you need any more details.

agramfort:
see https://github.com/mne-tools/mne-python/issues/8795

agramfort:
you should use https://pypi.org/project/memory-profiler/ for this @gemeinl

gemeinl:
Ok, got it. Will use memory profiler next time.
Thanks!

larsoner:
Feel free to try on https://github.com/mne-tools/mne-python/pull/8801 to see if it helps

gemeinl:
Looks great! Memory usage decreased from ~3300 MB to ~80 MB! 
mem_profile

agramfort:
nice !

>

agramfort (commented):
diff looks clean !

@hubertjb can you test on the TUH database we have on Inria server?

gemeinl:
It might be preferable to only have one function for loading of preprocessed datasets. Will look into possibilities of combining `load_concat_dataset` and `load_concat_datasets`.

hubertjb:
@gemeinl very cool stuff! :) @agramfort good idea, I'll test the example on my side.

braindecode/datautil/serialization.py
```diff
@@ -62,42 +64,63 @@ def save_concat_dataset(path, concat_dataset, overwrite=False):
...
     """
-    assert ((os.path.isfile(os.path.join(path, '0-raw.fif')) +
-             os.path.isfile(os.path.join(path, '0-epo.fif'))) == 1), (
-        "Expect either raw or epo to exist inside the directory")
+    # assume we have a single concat dataset to load
     concat_of_raws = os.path.isfile(os.path.join(path, '0-raw.fif'))
+    concat_of_epochs = os.path.isfile(os.path.join(path, '0-epo.fif'))
+    paths = [path]
+    # assume we have multiple concat datasets to load
+    if not (concat_of_raws + concat_of_epochs):
```
robintibor, line 91:
```suggestion
    if not (concat_of_raws or concat_of_epochs):
```

seems better as it is a logical comparison

gemeinl:
> @gemeinl very cool stuff! :) @agramfort good idea, I'll test the example on my side.

Thanks! It would be awesome if you could try to run the code on the entire dataset, since I am currently limited to a small subset (~1500 recordings) of it.

examples/tuh_eeg_corpus.py
```diff
@@ -0,0 +1,258 @@
...
+        # save memory by deleting raw recording
+        del tuh_subset
+        # store the number of windows required for loading later on
+        tuh_windows.description["n_windows"] = [len(d) for d in
+                                                tuh_windows.datasets]
+
+        # some recordings fail with
+        # info["meas_date"] seconds must be between "(-2147483648, 0)" and "(2147483647, 0)", got "-2209161600"  # noqa
+        # during save, so catch it
+        try:
```
gemeinl, line 228: Saving of recordings sometimes failed due to an invalid measurement date. I was able to detect these invalid dates using `mne.io.maes_info._check_dates`. However, this does not return `True` or `False` but raises a `RuntimeError` instead. Is there a better way to handle this @agramfort ?

examples/tuh_eeg_corpus.py
```diff
@@ -0,0 +1,258 @@
...
+            mne.io.meas_info._check_dates(tuh_windows.datasets[0].windows.info)
```
agramfort, line 229: using mne private functions outside mne is usually a bad idea. What's the pb with the meas_date ? what's the value that crashes?

gemeinl: I cannot say. I would have to dig back into it, it's been a while since I wrote the code... Would it be possible to repair the date or make mne offer a function that returns a bool instead of rasing an error?

agramfort:
>
> I cannot say. I would have to dig back into it, it's been a while since I
> wrote the code... Would it be possible to repair the date or make mne offer
> a function that returns a bool instead of rasing an error?
>

yes that would be much better.

examples/tuh_eeg_corpus.py
gemeinl: I ran the example on the entire TUH EEG Corpus and checked the dates and I cannot reproduce this behavior. I suspect that something happened when I copied a tiny subset of the data to my local machine to write the code.
I will hence remove the `try` statement as well as the call to `mne.io.meas_info._check_dates`.

robintibor:
Sow now seems ready to merge @gemeinl ?

gemeinl:
Ready from my side

robintibor:
Great great thanks merged :1st_place_medal: :100:

PR #195, opened 2021-03-18
Update documentation of windowers.py

gifale95:
The documentation string of the parameter drop_last_window of such function starts with "If True, an additional overlapping window...", whereas is should be saying "If False, ad additional overlapping window...".

### DIFF
--- a/braindecode/datautil/windowers.py
+++ b/braindecode/datautil/windowers.py
@@ -55,7 +55,7 @@ def create_windows_from_events(
         trial size of the first trial and trial_start_offset_samples and
         trial_stop_offset_samples.
     window_stride_samples: int | None
-        Stride between windows, in samples. If None, the window size is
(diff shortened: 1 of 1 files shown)


robintibor:
Amazing! If you are interested, you could also check if you find the other doc strings for drop_last_window (you can search in this repo for `drop_last_window: bool` to find them) to be clear to you or whether you would also update them to this version. 

If you are not so interested/busy I can just merge this already, definitely good to not have wrong documentation, thank you very much! :)

gifale95:
I found another possible typo in the documentation: please see my updated versions of them.
The other doc string instances of *drop_last_window* seem fine to me, as they are formulated in a different way.

Happy to help, and thank you for this toolbox!:)

robintibor:
Great! Thanks for your help! Merged! :)

PR #199, opened 2021-04-08
[MRG] Parallelizing windowers over BaseDatasets

hubertjb:
In my previous experiments on large datasets I found that an easy way to dramatically speed up the preprocessing is to parallelize the windowing step over `BaseDatasets` using joblib. This PR introduces this feature through a small refactoring of the windowing functions. I ended up adding `joblib` as a dependency - I wasn't sure whether it should actually be a soft dependency though (i.e. only try to import it if using this feature).

Also, I wasn't sure how to test that multiple processes are actually used when `n_jobs>1`. For the moment I only made sure that the results were the same as with `n_jobs=1`.

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -12,7 +12,6 @@ Dataset classes.
 
 import warnings
 
-import numpy as np
 import pandas as pd
 
 from torch.utils.data import Dataset, ConcatDataset
--- a/braindecode/datautil/windowers.py
+++ b/braindecode/datautil/windowers.py
@@ -18,6 +18,174 @@ from joblib import Parallel, delayed
 from ..datasets.base import WindowsDataset, BaseConcatDataset
 
 
+def create_windows_from_events(
+        concat_ds, trial_start_offset_samples, trial_stop_offset_samples,
+        window_size_samples=None, window_stride_samples=None,
+        drop_last_window=False, mapping=None, preload=False,
+        drop_bad_windows=True, picks=None, reject=None, flat=None,
+        on_missing='error', n_jobs=1):
+    """Create windows based on events in mne.Raw.
+
+    This function extracts windows of size window_size_samples in the interval
+    [trial onset + trial_start_offset_samples, trial onset + trial duration +
+    trial_stop_offset_samples] around each trial, with a separation of
+    window_stride_samples between consecutive windows. If the last window
+    around an event does not end at trial_stop_offset_samples and
+    drop_last_window is set to False, an additional overlapping window that
+    ends at trial_stop_offset_samples is created.
+
+    Windows are extracted from the interval defined by the following:
+
+                                                  trial onset +
+                    trial onset                     duration
+    |--------------------|------------------------------|---------------------|
+    trial onset -                                                 trial onset +
+    trial_start_offset_samples                                       duration +
+                                                      trial_stop_offset_samples
+
+    Parameters
+    ----------
+    concat_ds: BaseConcatDataset
+        A concat of base datasets each holding raw and description.
+    trial_start_offset_samples: int
+        Start offset from original trial onsets, in samples.
+    trial_stop_offset_samples: int
+        Stop offset from original trial stop, in samples.
+    window_size_samples: int | None
+        Window size. If None, the window size is inferred from the original
+        trial size of the first trial and trial_start_offset_samples and
+        trial_stop_offset_samples.
+    window_stride_samples: int | None
+        Stride between windows, in samples. If None, the window stride is
+        inferred from the original trial size of the first trial and
+        trial_start_offset_samples and trial_stop_offset_samples.
+    drop_last_window: bool
+        If False, an additional overlapping window that ends at
+        trial_stop_offset_samples will be extracted around each event when the
+        last window does not end exactly at trial_stop_offset_samples.
+    mapping: dict(str: int)
+        Mapping from event description to numerical target value.
+    preload: bool
+        If True, preload the data of the Epochs objects. This is useful to
+        reduce disk reading overhead when returning windows in a training
+        scenario, however very large data might not fit into memory.
+    drop_bad_windows: bool
+        If True, call `.drop_bad()` on the resulting mne.Epochs object. This
+        step allows identifying e.g., windows that fall outside of the
+        continuous recording. It is suggested to run this step here as otherwise
+        the BaseConcatDataset has to be updated as well.
+    picks: str | list | slice | None
+        Channels to include. If None, all available channels are used. See
+        mne.Epochs.
+    reject: dict | None
+        Epoch rejection parameters based on peak-to-peak amplitude. If None, no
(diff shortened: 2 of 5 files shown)


agramfort (commented):
so far so good. Let us know what's it's ready to merge from your end @hubertjb

hubertjb:
> so far so good. Let us know what's it's ready to merge from your end @hubertjb

Ready to merge on my side!

agramfort:
let's fix CIs before we merge PRs

ping me when green. thx

robintibor:
ok then let's use this to switch entire to github actions I would say. CircleCI now just times out even if errors are fixed. And from my side Github Actions PR almost ready (only deploy to github pages missing, should be quick). Then hubert could rebase his PRs after Github Actions PR is merged to master.

agramfort:
you will look an easy access to artifacts... we have this one click button
to see artifact now. It's super convenient.
If it's not broken don't fix it. Travis is not working as before we need a
replacement. circle-ci works just fine
and offers something GH action does not offer.

>

robintibor:
1. CircleCI atm is broken for us as it always times out. Would need to fix this again, possibly caching problem, problems we had before also already...
2. Github Actions resulting setup seems quite a bit simpler to me

agramfort:
what caused the time out on Circle? new pytorch version is slower? new
examples?

yes GH action is easier... but I really like that we can see the rendered
examples
in one click....

>

robintibor:
To be honest, I don't really know what causes the slowdown on CircleCI. Definitely the sleep staging and self supervised (relative positioning) examples are fairly slow (from https://app.circleci.com/pipelines/github/braindecode/braindecode/533/workflows/5426103a-34c8-412d-bc53-027bd37fdb63/jobs/635, last successful build on CircleCI): 
```
computation time summary:
    - ../examples/plot_sleep_staging.py:              985.15 sec    204.4 MB
    - ../examples/plot_relative_positioning.py:       620.72 sec    559.9 MB
    - ../examples/plot_bcic_iv_2a_moabb_cropped.py:   452.23 sec    925.8 MB
    - ../examples/plot_bcic_iv_2a_moabb_trial.py:     438.41 sec   1101.4 MB
    - ../examples/plot_regression.py:                  24.56 sec    201.9 MB
    - ../examples/plot_split_dataset.py:               17.79 sec    406.9 MB
    - ../examples/plot_dataset_example.py:              9.01 sec    385.5 MB
    - ../examples/plot_custom_dataset_example.py:       2.82 sec    116.9 MB
    - ../examples/plot_mne_dataset_example.py:          0.92 sec     69.1 MB
    - ../examples/benchmark_lazy_eager_loading.py:      0.00 sec      0.0 MB
 ```

But in my experience, CircleCI always seemed to have quite variable running times, even for identical commits, indeed also this last successful build was a rerun, same commit failed by timeout before: https://app.circleci.com/pipelines/github/braindecode/braindecode/533/workflows/10edeb1e-cf55-4554-bdec-d911abac0c79/jobs/649

MNE Data caching seems to not work properly according to output in successful build. Unless cache was not there since simply too long time since last successful build...

agramfort:
I'll try to fix circle-ci in
https://github.com/braindecode/braindecode/pull/206

let's focus on killing travis using GH actions

>

robintibor:
here you could also rebase. and maybe if you move the new private functions below the old public ones diff could get much smaller? now it is huge diff and hard to see what changed

agramfort:
GH actions tests don't run in the PRs now. Can you rebase now that I merged https://github.com/braindecode/braindecode/pull/208 ? thx @hubertjb

robintibor:
So seems some problem with tests @hubertjb ? Can you have a look, are those running locally?

Also seems the idea for making diff smaller didn't work but ok - so did you only wrap the existing functions to parallelize them or also change something inside? :)

robintibor:
Btw @agramfort here we also see a recurring problem for us with CircleCI: ```urllib.error.HTTPError: HTTP Error 403: Forbidden ``` that seemingly randomly appears at times (https://app.circleci.com/pipelines/github/braindecode/braindecode/613/workflows/29ecd173-4dc8-4bc0-a2ab-33ec8fe5e9ae/jobs/723 ) and normally disappears on rerunning the workflow (https://app.circleci.com/pipelines/github/braindecode/braindecode/613/workflows/e3930173-7518-453c-9221-22c872352ba5/jobs/725)

agramfort:
I suspect the data http server starts to filter connections from
certain IPs and it suggests our caching
still doesn't work as it should be.

>

robintibor:
Moving discussion of CircleCI to https://github.com/braindecode/braindecode/pull/209

hubertjb:
> So seems some problem with tests @hubertjb ? Can you have a look, are those running locally?
> 
> Also seems the idea for making diff smaller didn't work but ok - so did you only wrap the existing functions to parallelize them or also change something inside? :)

Mmh weird, the tests did work locally. I'll investigate more.

About the functions - it's really only a refactoring, so the content didn't change.

hubertjb:
I'm not sure why these two tests are failing, they seem to be working locally. In any case I've made a little change so that the test is more meaningful (the parallelization was only done over 1 dataset, now it's over 8), I don't think this will fix the problem though.

Could it be that multiprocessing jobs are handled differently on the CI?

robintibor:
Did you see my comment in the code?

hubertjb:
> Did you see my comment in the code?

Mmh I don't see your comments. Did you leave them as PR review comments?

robintibor:
Yes exactly 
image

hubertjb:
> Yes exactly
> image

The comments seem to be "pending", I think you have to click on "finish review" or something like that!

hubertjb:
Good point about the == @robintibor . Here it's done on mne.Epochs objects but I'll test that more extensively.

robintibor (changes requested):
Ah! Right well then here i finalize review, check equality

hubertjb:
@robintibor I just had a look, and the `mne.Epochs` object does have an overridden `__eq__` method (defined in the `SizeMixin` https://github.com/mne-tools/mne-python/blob/main/mne/utils/mixin.py#L39). The method computes the hash of the two objects to be compared, and so both a difference in the underlying numpy arrays or in the other attributes of the objects will return False.

I've also added missing checks for the other attributes of the `WindowsDataset` and `BaseConcatDataset` objects.

I'm not sure I get your point about `np.allclose` though. Can't we expect the numpy arrays to be exactly the same in this case here?

robintibor:
In general, I think it almost always bad practice to compare any floating points numbers or arrays via `==`, e.g. see https://bitbashing.io/comparing-floats.html https://possiblywrong.wordpress.com/2013/11/15/floating-point-equality-its-worse-than-you-think/  https://floating-point-gui.de/errors/comparison/

So better do a suitable check for closeness of floating point numbers.

hubertjb:
@robintibor this last commit fixes the tests if you want to take a look!

robintibor:
Great thanks! merged @hubertjb

PR #203, opened 2021-04-12
[WIP] Play with GitHub actions

robintibor:
@agramfort likes it and says it may be faster let's follow the wizard and get to know it


robintibor:
that was not correct

PR #190, opened 2021-02-24
Fix CI Problems PyTorch CPU

robintibor:
So we seem to have issues related to https://github.com/pytorch/pytorch/issues/49462 that have caused Travis CI to fail.

E.g. 
```python
a = torch.ones(16, 1, 100, 2)
conv = Conv2d(1, 40, kernel_size=(20, 1), stride=(1, 1))
conv(a)
```
even causes crashes already on CPU PyTorch, unrelated to braindecode.

Installing via pip seems to resolve this for now, probably have to wait for next PyTorch release for proper fix. Segmentation Fault still happens locally for unknown reasons on CPU, but at least one can see that tests pass.

Also need to add wfdb and pyunpack to installations for CircleCI and Travis.

PR #205, opened 2021-04-12
[WIP] Github actions in branch in this repo

robintibor:
Hm does it work this way in PR in same repo?


docs/requirements.txt
```diff
@@ -3,3 +3,4 @@ sphinx_rtd_theme
 numpydoc
 memory_profiler
 pillow
+ipython
```
agramfort, line 6: I would have rather killed the `from IPython.display import display` from the examples :)

robintibor:
Were you aware of this https://github.com/actions/upload-artifact/issues/14 @agramfort ? So basically the HTML Documentation Artifact will be a single zip file that you have to download, can't just click through it, e.g., see https://github.com/braindecode/braindecode/actions/runs/742481627

agramfort:
yes I've used it before. It's not as convenient as circle-ci though as you
don't have a rendered version online in one click. You
need to download and you cannot point people to one page with an HTML link

>

robintibor:
But plan would still be to use it to replace both CircleCI and Travis no? otherwise so far seems to me could turn out to be less complex than before, so I like that part.

agramfort:
shall we merge this after removing travis? you'll need to update badges on the readme file.

robintibor:
well right now it also builds docs, if we stick with CircleCI would we still want it to build docs? as like second check to compare to CircleCI behavior maybe not even so stupid?

robintibor:
@agramfort I suggest let's first merge https://github.com/braindecode/braindecode/pull/190 then rebase this one? Then no need to rename the plot_relative_positioning and plot_sleep_staging.py at least if one is willing to wait the 30min

agramfort (approved):
@robintibor feel free to merge

robintibor:
But don't we want to revent/undo the renamings e.g. examples/plot_relative_positioning.py → examples/relative_positioning.py ?

agramfort:
let's get the CIs green and can put it back in a next PR. One example at a
time

>

PR #207, opened 2021-04-13
[WIP] Reenable examples

robintibor:
Reenabling the examples and clarifying that github actions is doing both tests and docs


agramfort (approved):
+1 for MRG when green.

@robintibor can you just rebase following my fix in https://github.com/braindecode/braindecode/pull/208 ?

PR #209, opened 2021-04-14
[WIP] Copy HTML Docs From Github Actions to CircleCI

robintibor:
@agramfort Is there a reason we only cache specific parts of mne_data? This may be reason also for some of the caching issues. Additionally, I don't know why we need the additional Get data step in .circleci/config.yaml that downloads specific parts? I think the step also doesn't cover all the datasets we later use. Can't we just download datasets during running the examples, and then cache in the end? Anyways the  downloads that are not covered in Get data step probably currently happen during running examples anyways?


agramfort (commented):
let's see if this works !

robintibor:
So now I have a, I think workable,if quite hacky solution:

1. Make Tests, Build Docs on Github Actions
2. Create HTML Doc Artifact on Github Actions
3. Trigger Fetch-Doc CircleCI Pipeline from Github Actions, and supply it the link to the artifact using https://nightly.link/
4. CircleCI Fetch-Doc just fetches Github Artifact, uncompresses it and stores HTML Doc on CircleCI
5. Show Link to CircleCI HTML Doc in Github as before (see also in this PR already)

I feel in past I may have lost more time with CircleCI than it has provided benefit. I hope this setup can reverse that @agramfort 
It also makes it easy in the future to completely get rid of CircleCI if Github Actions does eventually provide functionality to directly browse built documentation. 

Missing now just to add deployment of github docs to the github actions, but this should be easy to do. Also then could remove old main.yml which is superseded by the link-to-doc.yaml, as long as some PRs still on old setup, could leave it inside.
Cannot guarantee that there won't be hickups with this somewhat hacky setup, but at least so far it seems to work and one person fully understands it. In the current version we would have to live with this always-passing "dummy-job", but I don't mind about it. If it is really an issue there could probably be ways around it.

agramfort (approved):
works nicely !

thx @robintibor

robintibor:
Great, merging, let's see if deploy step succeeds as well

PR #214, opened 2021-04-29
Load / save example

gemeinl:
- added an example to showcase saving and loading of braindecode datasets
- minor improvements to serialization functions

### DIFF
--- a/examples/plot_load_save_datasets.py
+++ b/examples/plot_load_save_datasets.py
@@ -23,7 +23,8 @@ ds = MOABBDataset(
 )
 
 ###############################################################################
-# We can or cannot apply preprocessing steps to the dataset to our liking.
+# We can apply preprocessing steps to the dataset. It is also possible to skip
+# this step and not apply any preprocessing.
 preprocess(
     concat_ds=ds,
     preprocessors=[MNEPreproc(fn='resample', sfreq=10)]
--- a/examples/plot_load_save_datasets.py
+++ b/examples/plot_load_save_datasets.py
@@ -1,0 +1,82 @@
+"""Load and save dataset example
+================================
+
+In this example, we show how to load and save braindecode datasets.
+"""
+
+# Authors: Lukas Gemein <l.gemein@gmail.com>
+#
+# License: BSD (3-clause)
+
+from braindecode.datasets.moabb import MOABBDataset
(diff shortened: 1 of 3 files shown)


examples/plot_load_save_datasets.py
```diff
@@ -0,0 +1,82 @@
+"""Load and save dataset example
+================================
```
agramfort, line 2:
```suggestion
"""
Load and save dataset example
=======================
```

agramfort (commented):
did you consider a different design where you could do:

ds.save(path='./')

load_dataset(path='./')

it would allow to save the object without a new function to import
and this API would be the same for all datasets (no need to say  "concat").

gemeinl:
No, we have not thought about it yet, but I like it.
Would it be an overkill or confusing to offer both? Due to the loading function, we will not get rid of the serialization module, so we could just call `save_concat_dataset` from within `ds.save()`, right?

The thing with 'concat' is always a bit tricky to me. I named the functions like this to be precise about what's going on. But actually, whenever we talk about datasets in braindecode, we mean concat of datasets, either holding mne raws or mne eopchs. I am always very insecure how to deal with this. For me it would be fine to just always call it 'dataset' If I am not mistaken, the BaseDataset is never really exposed to users, so there should not be too much confusion. What do you think?

agramfort:
>
> No, we have not thought about it yet, but I like it.
> Would it be an overkill or confusing to offer both? Due to the loading
> function, we will not get rid of the serialization module, so we could just
> call save_concat_dataset from within ds.save(), right?
>
$python -c "import this" | grep preferably
There should be one-- and preferably only one --obvious way to do it.

so I would favor only one way with the .save method

> The thing with 'concat' is always a bit tricky to me. I named the
> functions like this to be precise about what's going on. But actually,
> whenever we talk about datasets in braindecode, we mean concat of datasets,
> either holding mne raws or mne eopchs. I am always very insecure how to
> deal with this. For me it would be fine to just always call it 'dataset' If
> I am not mistaken, the BaseDataset is never really exposed to users, so
> there should not be too much confusion. What do you think?
>

agreed ! we always concatenate sessions or subjects etc.

maybe others have some opinion too...

robintibor:
Great, nice to have a documented example. During sprint we should consider how to make overall doc/tutorials nicely structured so peoplesee things easily.

PR #215, opened 2021-05-17
[WIP] Adding explicit handling of `dtype` argument for preprocessor

hubertjb:
This short PR adds the explicit handling of the `dtype` argument in `Preprocessor`. This argument is then passed to the `apply_function` method of `Raw` and `Epochs` objects when `fn` is a callable and `apply_on_array` is True. This allows casting the output of the preprocessor to another type, e.g. float32 since that's what pytorch uses by default.

### DIFF
--- a/braindecode/datautil/preprocess.py
+++ b/braindecode/datautil/preprocess.py
@@ -45,7 +45,9 @@ class Preprocessor(object):
     def __init__(self, fn, apply_on_array=True, **kwargs):
         if callable(fn) and apply_on_array:
             channel_wise = kwargs.pop('channel_wise', False)
-            kwargs = dict(fun=partial(fn, **kwargs), channel_wise=channel_wise)
+            dtype = kwargs.pop('dtype', None)
+            kwargs = dict(fun=partial(fn, **kwargs), channel_wise=channel_wise,
+                          dtype=dtype)
             fn = 'apply_function'
         self.fn = fn
         self.kwargs = kwargs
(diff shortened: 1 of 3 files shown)


braindecode/datautil/preprocess.py
```diff
@@ -45,7 +45,9 @@ class Preprocessor(object):
     def __init__(self, fn, apply_on_array=True, **kwargs):
```
robintibor, line 45:
```suggestion
    def __init__(self, fn, apply_on_array=True, dtype=None, **kwargs):
```

How about adding dtype then as an explicit param (with default None) to the constructor? And adding docstring saying it is only valid for apply_on_array? Or you think it would be worse this way?

robintibor:
Although I wonder if we really need it if use case is casting to right type for pytorch, then could be just like a final preproc step anyways? mhmh

hubertjb:
> Although I wonder if we really need it if use case is casting to right type for pytorch, then could be just like a final preproc step anyways? mhmh

@robintibor you're right, the following works too:

```python
def cast_array(x, dtype):
    return x.astype(dtype)

dtype = np.float32
preprocessors = [Preprocessor(cast_array, dtype=dtype)]
preprocess(windows_concat_ds, preprocessors)
assert windows_concat_ds[0][0].dtype == dtype
```

I liked the idea of using the functionality directly with MNE (instead of explicitly) as you avoid a step that is required or at least desirable most of the time. You can just add `dtype=np.float32` to your last preprocessing step, e.g. in the scaling to uV. I agree it might make the API less consistent though as you could expect this to be a separate step.

I don't have a strong opinion - do you have a preference @robintibor @agramfort ?

agramfort:
if it means we have 2 ways of doing the same thing I would say no
if it avoids a data copy and it's actually a bottleneck I would say yes

>

hubertjb:
> if it means we have 2 ways of doing the same thing I would say no if it avoids a data copy and it's actually a bottleneck I would say yes
> […](#)

With these criteria then it's a no. :P

Issue #189, opened 2021-02-12
Dataset improvements, including Incosistency and inefficiency of description


gemeinl:
Each `BaseDataset` has a pandas Series holding valuable information.
When creating a `BaseConcatDataset` description is currently copied into a DataFrame.
Users might want to change the description from the outside, for example adding the number of compute windows after windowing, which will result in inconsistent descriptions.

Suggested imrpovement:
`BaseConcatDataset` should not copy descriptions of `BaseDataset`.
It should instead provide a getter which concatenates all descriptions when called.
In addition, it should provide a setter which adds an entry to the description of `BaseDataset`.

hubertjb:
I like the idea. Only reason I see this could be problematic is if the concatenation of pd.Series is slow. I guess this is not likely to happen though as datasets usually don't have more than a few thousands of `description` series?

gemeinl:
`description` is used at rare occasions in our code. One place would be the splitting of datasets which is typically not required to be fast. It is more for convenience and called specifically by the user. If he intends to call it many times, he could create a copy.

I just realized that there is another copy of the `description` in the `WindowsDataset`.
Also, to me it is not ideal that the windowing functions do not take care of the `BaseDataset` / raw signals.
I feel like the `WindowsDataset` should replace the `BaseDataset` in the `BaseConcatDataset` or the windowing functions could give an option to delete the raws afterwards. Any thoughts on this?

hubertjb:
So to summarize, we have copies of `description` in `BaseDataset`, `WindowsDataset` and `BaseConcatDataset`. You're suggesting we don't actually copy `description` when creating `WindowsDatasets` or `BaseConcatDatasets`, but instead to provide getters to dynamically create `description` based on the underlying `BaseDataset`. Is that right? If so, I agree that `BaseConcatDataset` should probably collect `descriptions` dynamically. However, just to be clear, I believe`WindowsDataset` would still need its own copy since you can have a `BaseConcatDataset` that only encapsulates `WindowsDatasets`.

As for the windowers taking care of the `BaseDataset`, I agree that we could have something like an `inplace` argument that controls whether we overwrite the input `BaseConcatDataset` by the output of the windowers.

robintibor:
* multi-target
   * description part uses target column, a bit limiting
   * maybe keep target as a separated variable (like numpy array that can be returned)
      * might be possible to keep in dataframe column, but do we want that? 
[...]

PR #218, opened 2021-05-19
In TUHAbnormal added pathological label to base datasets.

gemeinl:
Through adding of the TUH class, pathology labels in TUHAbnormal are added manually to the concat dataset description. However, this results in accidentally dropping the labels when using `split()`, since it uses the descriptions of base datasets.


gemeinl:
Bug is connected to issue #189 regarding the duplication of description.

Issue #220, opened 2021-05-21
EEGResNet with `final_pool_length='auto'`


alamie:
I tried to use the `EEGResNet` as follows:

```python
import torch
from braindecode.models.eegresnet import EEGResNet

n_channels, n_times = 8, 4096

model = EEGResNet(in_chans=n_channels,
                  n_classes=1,
                  input_window_samples=n_times,
                  final_pool_length='auto',
                  n_first_filters=2)

inp = torch.rand((1, n_channels, n_times, 1))
out = model(inp)
```

But this code gives me the following error:
```
RuntimeError: Calculated padded input size per channel: (4096 x 1). Kernel size: (262081 x 1). Kernel size can't be greater than actual input size
```

I tried to increate the value of `n_times` but it does not help. Except for `final_pool_length='auto'`, everything is similar to [`test_eegresnet`](https://github.com/braindecode/braindecode/blob/41d6fc4479b8a3b7db443648d288e47981be29b1/test/unit_tests/models/test_models.py#L62). 

Is this behavior to be expected?

sliwy:
This is not an expected behavior. `EEGResNet` model does not work with `final_pool_length='auto'` at the moment.

I think the reason for that is the dilation parameter (`cur_dilation`) passed to the `AvgPool2dWithConv`
```python
self.add_module('mean_pool', AvgPool2dWithConv(
(self.final_pool_length, 1), (1, 1), dilation=(int(cur_dilation[0]),(cur_dilation[1]))))
```
You can find a possible fix (to be confirmed with @robintibor) and more details in the https://github.com/braindecode/braindecode/pull/221

alamie:
@sliwy I agree with you and think that the `dilation` parameter is the problem here. In my example, the `AvgPool2dWithConv` layer will receive a feature map with shape `(C, 4096, 1)`. Say that we have `dilation=(64, 1)` ; the forward pass of `AvgPool2dWithConv` expects a very "tall" input (the height of the input tensor must be strictly larger than 64 * (4096 - 1) for the dilation to make sense). It works if you set `dilation = (1, 1)`: you will pool a feature map from `(C, 4096, 1)` to `(C, 1, 1)`. 

[...]

PR #221, opened 2021-05-21
Removing dilated Avg Pooling from EEGResNet

sliwy:
It replaces dilated Global AVG Pooling in the `EEGResNet` with a standard implementation without dilation from pytorch and solves  #220 

I may not fully understand the Resnet architecture, so @robintibor please verify if I'm correct. I think that dilation is not needed in the Global Avg Pooling layer. It should create one number per feature map (in the case `final_pool_length=='auto'`), so we should use standard Avg Pooling from pytorch and just average without using dilated pooling kernel.

### DIFF
--- a/braindecode/models/deep4.py
+++ b/braindecode/models/deep4.py
@@ -3,7 +3,7 @@ from torch import nn
 from torch.nn import init
 from torch.nn.functional import elu
 
-from .modules import Expression, AvgPool2dWithConv, Ensure4d
+from .modules import Expression, Ensure4d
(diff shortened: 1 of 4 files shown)


robintibor:
Unfortunately this is not correct, for cropped decoding logic we do need to use dilations and need to get dense predictions (one prediction for each timestep, without gaps) in the end. The logic here would make deep4 break, the only reason it doesn't break already is because there is no test checking deep4 results with avg pooling.

robintibor:
Ok superseded by https://github.com/braindecode/braindecode/pull/223, closing

PR #223, opened 2021-05-25
[MRG] ResNet auto length

robintibor:
This should work I think @sliwy @alamie to solve https://github.com/braindecode/braindecode/issues/220
test copied from https://github.com/braindecode/braindecode/pull/221

### DIFF
--- a/braindecode/models/eegresnet.py
+++ b/braindecode/models/eegresnet.py
@@ -152,9 +152,11 @@ class EEGResNet(nn.Sequential):
                 dtype=np.float32)))
             n_out_time = out.cpu().data.numpy().shape[2]
             self.final_pool_length = n_out_time
+            pool_dilation = 1
+        else:
+            pool_dilation = int(cur_dilation[0]), int(cur_dilation[1])
         self.add_module('mean_pool', AvgPool2dWithConv(
-            (self.final_pool_length, 1), (1, 1), dilation=(int(cur_dilation[0]),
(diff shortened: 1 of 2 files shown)


sliwy:
@robintibor thanks for your feedback and corrections! :)
Regarding `Deep4Net` architecture, I had checked in the code and to me the dilation parameter was not used there. I think that `Deep4Net` is just using standard pooling without dilation, so we can replace `AvgPool2dWithConv` with `AvgPool2d`?

A quick question about cropped decoding, why do we use dilation in time in the last pooling layer in the ResNet architecture? I thought that crops are always continuous, so the only things we need is suitable stride and poling size in the last layer. Is using dilation in pooling a design choice to increase receptive field or something more used to obtain proper prediction per crop? (I don't try to change it, I just want to understand better :smile: )

robintibor:
Sure. So, we use dilation in deep4/shallow to perform cropped decoding in a fast way. Normally you would have to push each "crop" separately through the network to get an output for each timestep. However, for architectures without padding, it is possible to get *exactly* identical results for each timestep for many neighbouring crops by replacing strides by appropiate dilations (done in https://github.com/braindecode/braindecode/blob/7b80b71685951d3da991f244f22ff061abc4466b/braindecode/models/util.py#L5-L48 , explained in  https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.23730 ConvNet Training -> Cropped Training, especially Figure 4).

Issue #222, opened 2021-05-21
Variable Length Trials not working? Or are they?


robintibor:
It seems at the moment cropped scoring does not work for variable length trials? Although I have not really clearly confirmed that yet.
Here is an example that fails (unfortunately not runnable for others as it relies on local data on our cluster):
https://gist.github.com/robintibor/343b7b54fea558a7c73ea93d167490a4

Anyways @sliwy I had some problem to understand the code here:
https://github.com/braindecode/braindecode/blob/41d6fc4479b8a3b7db443648d288e47981be29b1/braindecode/training/scoring.py#L178-L182

Looks to me like this could never work for variable length trials? Still for some older version the above code apparently ran? Next steps here might be to create a simpler example with artificial variable length trial dataset (we would need it anyways).. Or also run TUHAbnormal that also actually shouldn't work I guess? I can also look at it more next week just since I saw you active @sliwy  thought I put it public here :)

sliwy:
I agree that it should not work at the moment. The solution might to pad each element of `trial_preds` with `np.nan` to be the same length. Then, each time we compute average over time we should use `np.nanmean` to drop those values. What do you think? @robintibor  It would be great if we can have a test scenario with variable trial length which should work in the end.

robintibor:
The thing is as far as I understand the code we just forward the preds to skorch and have to rely on their logic to work, no idea what they do with regard to nanmean... have to look inside a bit again

robintibor:
It seems in this commit we removed the averaging of the predictions in the scoring function (see towards bottom): 
https://github.com/braindecode/braindecode/commit/bd85babeeb9d90848d9e938b8f3946be2649cce6
This breaks some use cases, actually then it's also not really a "cropped trial" rather a "cropped crop" scoring function right?

robintibor:
Probably here it's explained what is supposed to happen: https://github.com/braindecode/braindecode/pull/171 Maybe this you meant with the nanmean there?

robintibor:
But regarding nan padding and nanmean... not so sure if it is good idea, seems make everything even harder to understand and also will hide errors when you actually do have nans inside.... wonder if there is a cleaner solution

sliwy:
@robintibor I think we can do this without nans. It's good you reminded me about this PR. The quickest solution would be in my opinion to rollback EpochScoring to this version:
```python 
            y_preds_per_trial = np.array(
                [np.mean(p, axis=1) for p in trial_preds]
            )
```
[...]

PR #224, opened 2021-05-26
[WIP] Variable length trials fix

sliwy:
Just two small changes to fix cropped decoding to work with variable length trials. Tests does not work locally in my case (no clue why), so let's check here. Trying to solve #222

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -221,7 +221,7 @@ class EEGClassifier(NeuralNetClassifier):
 
         """
         y_pred = super().predict_proba(X)
-        if self.cropped:
+        if self.cropped and len(y_pred.shape) == 3:
             return y_pred.mean(axis=-1)
         else:
             return y_pred
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -221,6 +221,11 @@ class EEGClassifier(NeuralNetClassifier):
 
         """
         y_pred = super().predict_proba(X)
+        # Normally, we have to average the predictions across crops/timesteps
+        # to get one prediction per window/trial
+        # Predictions may be already averaged in CroppedTrialEpochScoring (y_pred.shape==2).
+        # However, when predictions are computed outside of CroppedTrialEpochScoring
+        # we have to average predictions, hence the check if len(y_pred.shape) == 3
         if self.cropped and len(y_pred.shape) == 3:
             return y_pred.mean(axis=-1)
         else:
--- a/braindecode/training/scoring.py
+++ b/braindecode/training/scoring.py
@@ -175,8 +175,11 @@ class CroppedTrialEpochScoring(EpochScoring):
                 pred_results['preds'],
                 pred_results['i_window_in_trials'],
                 pred_results['i_window_stops'])
-            # trial preds is a list
-            # each item is an 2d array classes x time
+
+            # Average across the timesteps of each trial so we have per-trial
+            # predictions already, these will be just passed through the forward
+            # method of the classifier/regressor to the skorch scoring function.
+            # trial_preds is a list, each item is a 2d array classes x time
             y_preds_per_trial = np.array(
                 [np.mean(p, axis=1) for p in trial_preds]
             )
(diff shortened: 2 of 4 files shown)


braindecode/classifier.py
```diff
@@ -221,7 +221,7 @@ def predict_proba(self, X):
 
         """
         y_pred = super().predict_proba(X)
-        if self.cropped:
+        if self.cropped and len(y_pred.shape) == 3:
```
robintibor, line 224: Please add some comment here why this happens! And what we expect etc....

braindecode/training/scoring.py
```diff
@@ -177,7 +177,9 @@ def on_epoch_end(self, net, dataset_train, dataset_valid, **kwargs):
                 pred_results['i_window_stops'])
             # trial preds is a list
             # each item is an 2d array classes x time
-            y_preds_per_trial = np.array(trial_preds)
+            y_preds_per_trial = np.array(
```
robintibor, line 180: Also here some comments why we are averaging here would be great I think

robintibor (changes requested):

robintibor:
Great I think this can be merged, but would be great if you can add some comments so we know what we were doing there and why

braindecode/classifier.py
robintibor:
```suggestion
        if self.cropped and len(y_pred.shape) == 3:
```

```suggestion
# Normally, we have to average the predictions across crops/timesteps
# to get one prediction per time window/trial
# In the cropped scoring , we may have averaged these predictions 
# already, hence the check if len(y_pred.shape) == 3:
        if self.cropped and len(y_pred.shape) == 3:
```

braindecode/training/scoring.py
```diff
@@ -177,7 +177,9 @@ def on_epoch_end(self, net, dataset_train, dataset_valid, **kwargs):
...
+                [np.mean(p, axis=1) for p in trial_preds]
+            )
```
robintibor, line 182:
```suggestion
            # Average across the timesteps of each trial so we have per-trial predictions already,
            # these will be just passed through the forward method of the classifier/regressor
            # to the skorch scoring function
            y_preds_per_trial = np.array(
                [np.mean(p, axis=1) for p in trial_preds]
            )
```

robintibor: Do we also need to adapt the forward function of the regressor @sliwy ?

robintibor (changes requested):

braindecode/classifier.py
sliwy: this is what I left for later (together with the comments to do) if the changes from this PR works with your example. @robintibor have you checked if ti works with your example with variable trial length?

robintibor: I checked it in my mind and it seemed to work! Now also checked in reality and also worked! :)

robintibor: so do you have capacity to finish it? @sliwy  Otherwise I can have a try :)

sliwy: @robintibor I can take care of it today or tomorrow evening. If you need it faster feel free to finish that. :)

robintibor: sounds good! should be quick enough! :)

sliwy:
I added comments you asked for, I also modified EEGRegressor to work the same way as EEGClassifier. 

I still have some doubts whether the `predict_proba` should work with variable length trials (for scoring it should be ok). In particular, I don't know what will be the output of skorch `predict_proba` method when provided with observation of different shape. I've never done this before. Maybe we should create a test for that? It would need a custom DataLoader?  @robintibor  As you've already tested it on your side maybe you have a minimal working example for variable length trials decoding that we can use a starting point for the test?

robintibor:
I think my example is too complex, also requires loading TUH etc., but I can also see whether I can distill something out of it, if you have time before to make a test, also great, otherwise I will see

robintibor:
ok since my colleague needs it quite urgently now, merging this and we can do test in separate PR potentially.

robintibor (approved):

PR #226, opened 2021-06-03
Fixing evaluation of fully connected layer size in sleep stager

hubertjb:
Another quick fix to `SleepStagerChambon2018`. I realized I recently introduced a bug in the private method that computes the size of the last fully connected layer (it was accumulating the gradient and updating the batch norm and dropout layers, which shouldn't happen). This PR fixes this issue.


robintibor:
yap changes make sense! good that you caught it!

hubertjb:
Thanks @robintibor! :)

PR #243, opened 2021-06-16
[MRG] Data modules reorganization

### LINKED ISSUE #156: Reorganization of data-related modules
From @gemeinl https://github.com/braindecode/braindecode/pull/152/files/fac6cea5d6ea0af49dcb39c85d3be33dfa791c22#r470535380 
Personally, I don't like the name "datautil" at all and feel like we could use some major refactoring here. My thoughts:
Currently, ``braindecode.datautil`` for me contains a mess of 
- Loading and saving datasets (``serialization.py``) 
- Creating datasets (``xy.py``, ``mne.py``)
- Preprocessing (``preprocess.py``)
- Creating windows ((``windowers.py``) which is also a bad name, since we don't have windower classes anymore). Also, creating windows is a kind of preprocessing

We could probably rename ``braindecode.datautil`` to ``braindecode.preprocessing`` with ``preprocessing.py`` which contains a merge of ``preprocess.py`` and ``windowers.py``.

We could add ``braindecode.serialization`` with ``serialization.py``

I am not so sure what to do about ``xy.py`` and ``mne.py``, though.

robintibor:
As agreed in https://github.com/braindecode/braindecode/issues/156

Closes #156


robintibor:
So this should be the complete reorganization as written in #156 :
- datasets
   - base.py
   - xy.py
   - mne.py
   - bbci.py
   - moabb.py
   - tuh.py
- preprocessing
  - preprocess.py
  - windower.py
- datautil
  - serialization.py
  - util.py
 
It is backwards compatible with deprecation warnings, this also necessitates that the old files still exist, just emptied and with the deprecation warning there, e.g., see https://github.com/braindecode/braindecode/blob/6ded712d6d2703798fd7801267cdfa52f2500594/braindecode/datautil/mne.py#L1-L8

In order to be able to do the Flake8 stuff, I would like to merge this still later today or  latest early tomorrow, so please voice any objections @agramfort @hubertjb @gemeinl @sliwy

robintibor:
since no objections, now merging this, docs should not get updated on braindecode.org now, that is intended see #248

PR #248, opened 2021-06-18
[WIP] Push new docs to master subdirectory

robintibor:
So I want to prevent that we have docs again that are different from current version on PyPI.

I have not yet figured out how to create proper drop down menu like MNE and others have, but thought for now let's at least push new docs to subdirectory already, especially before merging https://github.com/braindecode/braindecode/pull/243

So that we don't have issues again from people confused that things don't work for them as they are shown in docs...

Is that fine as a first step now @agramfort ? I don't want to do the full dropdown stuff yet as it is blocking https://github.com/braindecode/braindecode/pull/243 and then also flake8 preparation before sprint.


agramfort:
have a look at
https://holzhaus.github.io/sphinx-multiversion/master/index.html

it's used here for example:
https://github.com/mne-tools/mne-nirs/blob/master/doc/conf.py#L49

>

robintibor:
great thanks!! seems like a good sprint task! then merging this hack for now, to be later replaced by proper thing with sphinx-multiversion.

PR #251, opened 2021-06-20
more flake8 + mimic MNE config

agramfort:


robintibor:
Ah right there is the setup.cfg . I had put flake8 config into tox.ini, no special reason, should we then merge the tox.ini stuff into setup.cfg and delte the tox.ini ? @agramfort

agramfort:
merge if happy @robintibor

robintibor:
ah great bad people are always happy so bad people can always merge! Thanks @agramfort

PR #254, opened 2021-06-21
[WIP] Starting point for batch-level data augmentation

cedricrommel:


braindecode/augmentation/base.py
```diff
@@ -0,0 +1,208 @@
...
+        Valid range of the argument mapped by `magnitude` (e.g. standard
+        deviation, number of sample, etc.):
+        ```
+        argument = magnitude * mag_range[1] + (1 - magnitude) * mag_range[0].
+        ```
+        If `magnitude` is None it is ignored. Defaults to None.
+    random_state: int, optional
+        Seed to be used to instatiate numpy random number generator instance.
+        Used to decide whether or not to transform given the probability
+        argument. Defaults to None.
```
sbbrandt, line 52: possible interference with `set_random_seeds`?

robintibor:
Hi, very interesting thansk, seems a lot of work, will need to have a closer look! This works fine with what was done in https://github.com/braindecode/braindecode/pull/198/files ? Note the discussion there as well

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Parameters
+    ----------
+    transforms: list
+        Sequence of Transforms to be composed.
+    """
+
+    def __init__(self, transforms):
+        self.transforms = transforms
+        super().__init__(operation=identity)
```
tomMoral, line 152: This feels weird. Maybe pass `operation=None` here?

braindecode/augmentation/functionals.py
```diff
@@ -0,0 +1,581 @@
+from numbers import Real
```
tomMoral, line 1: Add a doc string saying it is private API and move to `_functionals.py`?

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+import numpy as np
+from sklearn.utils import check_random_state
+import torch
+
+from braindecode.augmentation.base import Transform
+from braindecode.augmentation.functionals import time_reverse, sign_flip,\
+    downsample_shift_from_arrays, fft_surrogate, channel_dropout,\
+    channel_shuffle, add_gaussian_noise, permute_channels, random_time_mask,\
+    identity, random_bandstop, freq_shift, random_rotation,\
+    get_standard_10_20_positions
```
tomMoral, line 16: Maybe don't use `\` and use the `(...)` syntax?

tomMoral (commented):
A first batch of comments

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    ----------
+    probability : object, optional
+        Always ignored, exists for compatibility.
+    magnitude : object, optional
+        Always ignored, exists for compatibility.
+    mag_range : object, optional
+        Always ignored, exists for compatibility.
+    factor: int, optional
+        Factor by which X should be downsampled. For example, if factor is 2,
+        only every other column of X will be kept. Defaults to 2.
```
tomMoral, line 102: Not sure what you mean with "only every other column" (but this might be my bad english).
Maybe "For example, if factor is 2, this will keep one sample out of 2 in X".

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    offset: int, optional
+        Offset (in number of columns) to be used to time-shift the data.
+        Offset needs to be less than factor.
+        When downsampling by a factor N, you have N different offsets possible.
+        If not value is passed to ofset, it is randomly relected between 0 and
```
tomMoral, line 107:
```suggestion
        If no value is passed to ``offset``, it is randomly selected between 0 and
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        factor-1.
```
tomMoral, line 108:
```suggestion
        ``factor-1``.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    magnitude : object, optional
+        Always ignored, exists for compatibility.
+    mag_range : object, optional
+        Always ignored, exists for compatibility.
+    factor: int, optional
+        Factor by which X should be downsampled. For example, if factor is 2,
+        only every other column of X will be kept. Defaults to 2.
+    offset: int, optional
+        Offset (in number of columns) to be used to time-shift the data.
+        Offset needs to be less than factor.
```
tomMoral, line 105:
```suggestion
        Offset needs to be less than ``factor``.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        When downsampling by a factor N, you have N different offsets possible.
```
tomMoral, line 106:
```suggestion
        When downsampling with ``factor=N``, you have `N` different offsets possible.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        self,
+        probability=None,
+        magnitude=None,
+        mag_range=None,
+        factor=2,
+        offset=None,
+        random_state=None
+    ):
+        assert isinstance(factor, int), "factor has to be an int"
+        kwargs = {'random_state': random_state}
```
tomMoral, line 132: you could drop this `kwargs` and simply modify `random_state = self.rng` in line 136. This will be more explicit.

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+
+    Parameters
+    ----------
+    probability: float | callable
+        Either a float between 0 and 1 defining the uniform probability of
+        applying the operation, or a `callable` (modelling a non-uniform
+        probability law for example) taking no input and returning a boolean.
+    magnitude : object, optional
+        Float between 0 and 1 setting the range over which the phase
+        pertubation is uniformly sampled: [0, `magnitude` * 2 * `pi`]. Defaults
```
tomMoral, line 164:
```suggestion
        pertubation is uniformly sampled: ``[0, magnitude * 2 * pi]``. Defaults
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    ----------
+    probability : float | callable
+        Either a float between 0 and 1 defining the uniform probability of
+        applying the operation, or a `callable` (modelling a non-uniform
+        probability law for example) taking no input and returning a boolean.
+    magnitude : object, optional
+        Always ignored, exists for compatibility.
+    mag_range : object, optional
+        Always ignored, exists for compatibility.
+    random_state: int | numpy.random.Generator, optional
```
tomMoral, line 32:
```suggestion
    random_state : int | numpy.random.Generator, optional
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
```
tomMoral, line 65:
```suggestion
    random_state : int | numpy.random.Generator, optional
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            min_val, max_val = mag_range
+            self.std = magnitude * max_val + (1 - magnitude) * min_val
+        else:
+            self.std = std
+
+        super().__init__(
+            operation=add_gaussian_noise,
+            probability=probability,
+            magnitude=magnitude,
+            mag_range=mag_range,
```
tomMoral, line 351: Actually, the `mag_range` is not used in `Transform` so I think it would be best to remove it and simply pass the `std`. This is the same for other transform and it will avoid many of the `always ignored` for this arguments.

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    Parameters
+    ----------
+    probability : float | callable
+        Either a float between 0 and 1 defining the
+        uniform probability of applying the operation, or a `callable`
+        (modelling a non-uniform probability law for example) taking no input
+        and returning a boolean.
+    magnitude : float | None, optional
+        Float between 0 and 1 encoding the number of consecutive samples within
+        `mag_range` to set to 0:
```
tomMoral, line 443:
```suggestion
        ``mag_range`` to set to 0:
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            operation=operation,
+            magnitude=magnitude,
+            mag_range=mag_range,
+            mask_len_samples=self.mask_len_samples,
+            random_state=random_state,
+        )
+
+
+class BandstopFilter(Transform):
+    """Applies a stopband filter with desired bandwidth at a randomly selected
```
tomMoral, line 512:
```suggestion
    """Apply a band-stop filter with desired bandwidth at a randomly selected
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    frequency position between 0 and `max_freq`.
```
tomMoral, line 513:
```suggestion
    frequency position between 0 and ``max_freq``.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    ):
+        assert isinstance(bandwidth, Real) and bandwidth >= 0,\
+            "bandwidth should be a non-negative float."
+        assert isinstance(sfreq, Real) and sfreq > 0,\
+            "sfreq should be a positive float."
+        assert isinstance(max_freq, Real) and max_freq > 0,\
+            "max_freq should be a positive float."
+        nyq = sfreq / 2
+        if max_freq is None or max_freq > nyq:
+            max_freq = nyq
```
tomMoral, line 580: If `max_freq > nyq`, you should probably output a warning no?

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        if self.bandwidth == 0:
+            operation = identity
+        else:
+            operation = random_bandstop
+
+        super().__init__(
+            probability=probability,
+            operation=operation,
+            magnitude=magnitude,
+            mag_range=mag_range,
```
tomMoral, line 603:
```suggestion
            # kwargs for the random_bandstop operation
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            bandwidth=self.bandwidth,
+            max_freq=self.max_freq,
+            sfreq=self.sfreq,
+            random_state=random_state,
+        )
+
+
+class FrequencyShift(Transform):
+    """Add a random shift in the frequency domain to all channels.
+
```
tomMoral, line 613:
```suggestion

Note that here, the shift is the same for all channels of a single sample.

```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    Parameters
+    ----------
+    probability : float | callable
+        Either a float between 0 and 1 defining the
+        uniform probability of applying the operation, or a `callable`
+        (modelling a non-uniform probability law for example) taking no input
+        and returning a boolean.
+    sensors_positions_matrix : numpy.ndarray
+        Matrix giving the positions of each sensor in a 3D cartesian coordiante
```
tomMoral, line 693:
```suggestion
        Matrix giving the positions of each sensor in a 3D cartesian coordinate
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        systemsof. Should have shape (3, n_channels), where n_channels is the
```
tomMoral, line 694:
```suggestion
        system. Should have shape (3, n_channels), where n_channels is the
```

cedricrommel:
Hey @robintibor, thanks a lot ! Glad it looks interesting. @sbbrandt will soon add his nice MixUp implementation as well and we are still working on polishing things a little with him and @tomMoral. Concerning @hubertjb's PR, yes we are aware. We've discussed that when the PR was opened and have a few ideas on how to ensure both are compatible (since both logics of batch-level and sample-level data augmentation have their own pros and cons). We will get to that as soon as we are done polishing :)

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+
+    Parameters
+    ----------
+    probability : float | callable
+        Either a float between 0 and 1 defining the
+        uniform probability of applying the operation, or a `callable`
+        (modelling a non-uniform probability law for example) taking no input
+        and returning a boolean.
+    magnitude : float | None, optional
+        Float between 0 and 1 encoding the `max_shift` parameter:
```
tomMoral, line 622:
```suggestion
        Float between 0 and 1 encoding the ``max_shift`` parameter:
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        ```
+        max_shift = magnitude * mag_range[1] + (1 - magnitude) * mag_range[0]
+        ```
+        Defaults to None (ignored).
+    mag_range : tuple of two floats | None, optional
+        Range of possible values for `max_shift` settable using the magnitude
```
tomMoral, line 628:
```suggestion
        Range of possible values for ``max_shift`` settable using the magnitude
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        (see `magnitude`). If omitted the range (0 Hz, 5 Hz) will be used.
```
tomMoral, line 629:
```suggestion
        (see `magnitude`). If omitted the range [0, 5] Hz will be used.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        Float between 0 and 1 encoding the `max_degree` parameter:
+        ```
+        max_degree = magnitude * mag_range[1] + (1 - magnitude) * mag_range[0]
+        ```
+        Defaults to None (ignored).
+    mag_range : tuple of two floats | None, optional
+        Range of possible values for `max_degree` settable using the magnitude
+        (see `magnitude`). If omitted, the range (0, 30 degrees) will be used.
+    axis : 'x' | 'y' | 'z', optional
+        Axis around which to rotate. Defaults to 'z'.
```
tomMoral, line 816: For this one, this is fixed to `z`.

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
```
tomMoral, line 888: Ditto
```suggestion
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            magnitude=magnitude,
+            mag_range=mag_range,
+            axis='z',
+            max_degrees=max_degrees,
+            random_state=random_state
+        )
+
+
+class RandomYRotation(RandomSensorsRotation):
+    """Interpolates EEG signals over sensors rotated around the Z axis
```
tomMoral, line 860: A fast copy paste ;)
```suggestion
    """Interpolates EEG signals over sensors rotated around the Y axis
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            max_degrees=self.max_degrees,
+            sensors_positions_matrix=self.sensors_positions_matrix,
+            spherical_splines=self.spherical_splines,
+            random_state=random_state
+        )
+
+
+class RandomZRotation(RandomSensorsRotation):
+    """Interpolates EEG signals over sensors rotated around the Z axis
+    with an angle sampled uniformly between 0 and `max_degree`.
```
tomMoral, line 789: Inconsistent with `max_degree` doc string.
```suggestion
    with an angle sampled uniformly between ``-max_degree`` and ``max_degree``.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            mag_range=mag_range,
+            axis='z',
+            max_degrees=max_degrees,
+            random_state=random_state
+        )
+
+
+class RandomYRotation(RandomSensorsRotation):
+    """Interpolates EEG signals over sensors rotated around the Z axis
+    with an angle sampled uniformly between 0 and `max_degree`.
```
tomMoral, line 861:
```suggestion
    with an angle sampled uniformly between ``-max_degree`` and ``max_degree``.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        (see `magnitude`). If omitted, the range (0, 30 degrees) will be used.
+    axis : 'x' | 'y' | 'z', optional
+        Axis around which to rotate. Defaults to 'z'.
+    max_degree : float, optional
+        Maximum rotation. Rotation angles will be sampled between `-max_degree`
+        and `max_degree`. Defaults to 15 degrees.
+    spherical_splines : bool, optional
+        Whether to use spherical splines for the interpolation or not. When
+        `False`, standard scipy.interpolate.Rbf (with quadratic kernel) will be
+        used (as in the original paper). Defaults to True.
```
tomMoral, line 823: This is not exposed I guess?
```suggestion
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
```
tomMoral, line 895:
```suggestion
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+            mag_range=mag_range,
+            axis='y',
+            max_degrees=max_degrees,
+            random_state=random_state
+        )
+
+
+class RandomXRotation(RandomSensorsRotation):
+    """Interpolates EEG signals over sensors rotated around the Z axis
+    with an angle sampled uniformly between 0 and `max_degree`.
```
tomMoral, line 933:
```suggestion
    """Interpolates EEG signals over sensors rotated around the X axis
    with an angle sampled uniformly between ``-max_degree`` and ``max_degree``.
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+        Float between 0 and 1 encoding the `max_degree` parameter:
+        ```
+        max_degree = magnitude * mag_range[1] + (1 - magnitude) * mag_range[0]
+        ```
+        Defaults to None (ignored).
+    mag_range : tuple of two floats | None, optional
+        Range of possible values for `max_degree` settable using the magnitude
+        (see `magnitude`). If omitted, the range (0, 30 degrees) will be used.
+    axis : 'x' | 'y' | 'z', optional
+        Axis around which to rotate. Defaults to 'z'.
```
tomMoral, line 960:
```suggestion
```

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,1000 @@
...
+    max_degree : float, optional
+        Maximum rotation. Rotation angles will be sampled between `-max_degree`
+        and `max_degree`. Defaults to 15 degrees.
+    spherical_splines : bool, optional
+        Whether to use spherical splines for the interpolation or not. When
+        `False`, standard scipy.interpolate.Rbf (with quadratic kernel) will be
+        used (as in the original paper). Defaults to True.
```
tomMoral, line 967:
```suggestion
```

tomMoral (commented):
Few more comments on transform.

test/unit_tests/augmentation/test_base.py
```diff
@@ -0,0 +1,156 @@
...
+        dummy_transform(X, y),
+        expected_tensor
+    )
+
+
+def test_transform_proba_exception(random_batch, rng_seed):
+    rng = check_random_state(rng_seed)
+    with pytest.raises(AssertionError):
+        k = rng.randint(10)
+        _ = Transform(
```
tomMoral, line 98: Not really needed here no? Or maybe I am missing something.
```suggestion
        Transform(
```

test/unit_tests/augmentation/test_transforms.py
```diff
@@ -0,0 +1,477 @@
...
+from scipy.fft import fft, fftfreq, fftshift
+from scipy.signal import find_peaks, welch
+from sklearn.utils import check_random_state
+import torch
+
+from braindecode.augmentation.transforms import TimeReverse,\
+    MissingChannels, ShuffleChannels, FTSurrogate,\
+    GaussianNoise, ChannelSymmetry, TimeMask, SignFlip,\
+    BandstopFilter, FrequencyShift, RandomZRotation, RandomYRotation,\
+    RandomXRotation
```
tomMoral, line 16: `\` -> `(`? :slightly_smiling_face:

tomMoral (commented):
Last few comments for the day. I will do a second review tomorrow.
But this looks good overall.

braindecode/augmentation/transforms.py
cedricrommel: Yes, "every other" means '1 out of 2" 😝

braindecode/augmentation/transforms.py
cedricrommel: Nice precision 💪 I think I will just replace `sample` by `example` here, as the former is a bit ambiguous for signals.

braindecode/augmentation/transforms.py
cedricrommel: You're right, but it's a mistake :) I will fix it

robintibor:
Definitely looks like this is growing to something nice! Make sure to also create some examples, an add to API docs etc. .

test/unit_tests/augmentation/test_base.py
cedricrommel: The idea was not to pollute the stdout during tests, but I guess that's not really a problem :)

braindecode/augmentation/base.py
cedricrommel: As discussed, it does not interfere with `set_random_seeds` and just complements it, so I guess I'll leave it here unless it raises some issues :)

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,211 @@
...
+        return out_X, out_y
+
+    def _get_mask(self, batch_size=None) -> torch.Tensor:
+        """Samples whether to apply operation or not over the whole batch
+        """
+        size = (batch_size, 1, 1)
+        return torch.as_tensor(self.probability > self.rng.uniform(size=size))
+
+    # Might seem like an overkill, but making probability and magnitude into
+    # properties is useful in context where Transform is overclassed to make
```
robintibor, line 112: overclassed? you mean subclassed?

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,211 @@
...
+            if type(tr_y) is tuple:
+                out_y = tuple(tmp_y[mask == 1] for tmp_y in tr_y)
+            else:
+                out_y[mask == 1] = tr_y[mask == 1]
+        return out_X, out_y
+
+    def _get_mask(self, batch_size=None) -> torch.Tensor:
+        """Samples whether to apply operation or not over the whole batch
+        """
+        size = (batch_size, 1, 1)
```
robintibor, line 108: why do you need the two empty dimensions at end?

robintibor:
I would still also argue to write transforms that only transform X on a per example basis using the existing  https://github.com/braindecode/braindecode/pull/198/files solution. And keep the dataloader way for those transformations that need access to either X and y or to batch-level examples? Or what do you think? Otherwise it becomes confusing when to use which one? You could try to recreate your Transform Superclass as an "ExampleTransform" superclass it would make this transitition easier.

robintibor:
also @hubertjb let me know what you think

braindecode/augmentation/_functionals.py
```diff
@@ -1,3 +1,11 @@
+""" This module contains the internal functions defining data augmentations.
+They are not supposed to be part of the public API.
```
tomMoral, line 2: They are not part of the public API.

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,208 @@
...
+                isinstance(mag_range, Iterable) and
+                len(mag_range) == 2 and
+                all([isinstance(v, Real) for v in mag_range])
+            ) or mag_range is None
+        ), "mag_range should be None or a tuple of two floats."
+        self.mag_range = mag_range
+        self.args = args
+        self.kwargs = kwargs
+
+    def forward(self, X: Tensor, y: Tensor) -> Output:
```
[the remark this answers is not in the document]
cedricrommel: Yes, so as discussed yesterday, I'll make sure it works in both use-cases :)

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def make_collateable(transform):
+    def _collate_fn(batch):
+        X, y, _ = default_collate(batch)
+        return transform(X, y)
+    return _collate_fn
+
+
+class BaseDataLoader(DataLoader):
```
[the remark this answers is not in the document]
cedricrommel: Indeed, I will rename it

braindecode/augmentation/base.py
cedricrommel: Yes 😛 But this comment was already removed anyway

braindecode/augmentation/base.py
cedricrommel: It comes from a more complex version of Transform but it's indeed useless here. Removing it.

braindecode/augmentation/_functionals.py
```diff
@@ -0,0 +1,744 @@
...
+        rng.uniform(0, 1, size=(batch_size, n_channels)),
+        dtype=torch.float,
+        device=X.device,
+    )
+    # equivalent to a 0s and 1s mask, but allows to backprop through
+    # could also be done using torch.distributions.RelaxedBernoulli
+    return torch.sigmoid(1000*(unif_samples - magnitude)).to(X.device)
+
+
+def channel_dropout(X, y, magnitude, random_state, *args, **kwargs):
```
agramfort, line 161: all channels which are not private MUST have a docstring and be listed in the public API page.
If it's meant to be private call it _channel_dropout

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,223 @@
...
+        assert probability <= 1. and probability >= 0., \
+            "probability should be between 0 and 1."
+        self._probability = probability
+        self.rng = check_random_state(random_state)
+        assert (
+            (
+                isinstance(magnitude, Real) and
+                0 <= magnitude <= 1
+            ) or magnitude is None
+        ), "magnitude can be either a float between 0 and 1 or None"
```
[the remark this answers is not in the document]
cedricrommel: I'm not sure I get it, sorry. You can test AssertionError on pytest too. Or is it because the condition tested is confusing and I should decompose it?

braindecode/augmentation/_functionals.py
```diff
@@ -0,0 +1,746 @@
...
+from scipy.interpolate import Rbf
+from sklearn.utils import check_random_state
+import torch
+from torch.fft import fft, ifft
+from torch.nn.functional import pad, one_hot
+from mne.filter import notch_filter
+from mne.channels import make_standard_montage
+
+
+def identity(x, y, *args, **kwargs):
```
robintibor, line 21: is it necessary to have these *args, **kwargs in all of these functions in the signature?

braindecode/augmentation/_functionals.py
robintibor: and this is even so for this file `_functional.py` which has underscore in the filename?

braindecode/augmentation/base.py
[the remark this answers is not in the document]
robintibor: amazing

braindecode/augmentation/transforms.py
```diff
@@ -0,0 +1,983 @@
...
+        if offset is None:
+            self.rng = check_random_state(random_state)
+            offset = self.rng.choice(np.arange(factor))
+            random_state = self.rng
+        else:
+            assert isinstance(offset, int), "offset has to be an int"
+            assert offset < factor, (
+                "offset needs to be less than factor. When down sampling"
+                " by a factor N, you have N different offsets possible."
+            )
```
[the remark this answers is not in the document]
robintibor: Hm in general I very much like these kinds of checks inside the code, I also use them quite often. and you can check for asserts with this pytest.raises stuff no?

robintibor:
one minor thing I noticed going through the code was the use of the term "features" a lot in the documentation, which so far we did not really use in braindecode. We usually just wrote in terms of input/output, and usually "features" to us implies that there was already some preprocessing to extract something like band-power or whatever, like the difference between "feature-based decoding/learning" (e.g., LDA on bandpower features) and "raw end-to-end learning" (e.g., deep networks on EEG signals). I am not sure it is supercritical but we could consider about this at least, whether we want to rephrase documentation here a bit more towards existing input/output terminology.

robintibor:
Also consider often in deep learning, we speak of the model before the final layer as the "feature_extractor" etc. so I think it may really be a bit confusing

braindecode/augmentation/_functionals.py
cedricrommel: No, it was for compatibility with another repo of ours, but it should not be in braindecode as discussed with @agramfort. I'm going to remove them 👍

braindecode/augmentation/functional.py
```diff
@@ -18,19 +14,98 @@
...
+        EEG labels for the example or batch.
+
+    Returns
+    -------
+    torch.Tensor
+        Transformed inputs.
+    torch.Tensor
+        Transformed labels.
+    """
     return (x, y)
```
agramfort, line 34:
```suggestion
    return x, y
```

braindecode/augmentation/functional.py
```diff
@@ -144,7 +250,7 @@ def fft_surrogate(X, y, magnitude, random_state, *args, **kwargs):
     return transformed_X, y
 
 
-def _pick_channels_randomly(X, magnitude, random_state):
+def _pick_channels_randomly(X, magnitude, random_state=None):
```
agramfort, line 253: I tend to avoid default params in private aux functions to avoid forgetting to pass the param

braindecode/augmentation/functional.py
```diff
@@ -153,17 +259,45 @@ def _pick_channels_randomly(X, magnitude, random_state):
...
+    .. [1] Saeed, A., Grangier, D., Pietquin, O., & Zeghidour, N. (2020).
+       Learning from Heterogeneous EEG Signals with Differentiable Channel
+       Reordering. arXiv preprint arXiv:2010.13694.
+    """
+    mask = _pick_channels_randomly(X, magnitude, random_state=None)
     return X * mask.unsqueeze(-1), y
 
 
-def _make_permutation_matrix(X, mask, random_state):
+def _make_permutation_matrix(X, mask, random_state=None):
```
agramfort, line 300: idem

braindecode/augmentation/functional.py
```diff
@@ -203,11 +403,37 @@ def add_gaussian_noise(X, y, std, random_state, *args, **kwargs):
...
+    ----------
+    .. [1] Deiss, O., Biswal, S., Jin, J., Sun, H., Westover, M. B., & Sun, J.
+       (2018). HAMLET: interpretable human and machine co-learning technique.
+       arXiv preprint arXiv:1803.09702.
+    """
     return X[..., permutation, :], y
 
 
-def _sample_mask_start(X, mask_len_samples, random_state):
+def _sample_mask_start(X, mask_len_samples, random_state=None):
```
agramfort, line 436: idem

braindecode/augmentation/functional.py
```diff
@@ -236,13 +455,84 @@ def _relaxed_mask_time(X, mask_start_per_sample, mask_len_samples):
     return X * mask
 
 
-def random_time_mask(X, y, mask_len_samples, random_state, *args, **kwargs):
+def random_time_mask(X, y, mask_len_samples, random_state=None):
+    """Smoothly replace a contiguous part of all channels by zeros
```
agramfort, line 459:
```suggestion
    """Smoothly replace a contiguous part of all channels by zeros.
```

braindecode/augmentation/transforms.py
```diff
@@ -384,10 +380,10 @@ def __init__(
 
 
 class TimeMask(Transform):
-    """Replace part of all channels by zeros
+    """Smoothly replace a randomly chosen contiguous part of all channels by
+    zeros
```
agramfort, line 384:
```suggestion
    zeros.
```

agramfort (approved):

cedricrommel:
Ok guys, we've brainstormed a lot with @agramfort and it led us to propose some changes to the API which I've started to implement in the last commit. As this commit might be scary and confusing, here are some explanations.

After discussing with @agramfor, we agreed that the Transform API up to
now had some unecessary complexity that was useful in some of the teams
works, but which might be confusing for braindecode users and
contributors. These include mainly the magnitude and mag_range
parameters, as well as the kwargs. Besides the clarity and
over-complexity issue, the kwargs posed an additional problem. While it
is useful to enable the generic Transform class to handle arbitrary
`operation` functions, which may require different types
of arguments, it is difficult to document and potentially
dangerous (as some kwargs may be mispelled and unused within the
functionals).

In order to prevent both these problems, we propose in this commit to
drop unecessary arguments and to get inspiration from the `torchvision`
library to handle operations with different signatures. More precisely:
1) we drop the `magnitude` and `mag_range` parameters;
2) we remove most of the random samplings from operations defined within
`functional.py`, leaving it to the classes defined within `transforms.py`;
3) we use a new get_params method within each class
inheriting from `Transform` to handle operations with different
signatures (as well as the random samplings).

As this change will require some work, I am first commiting this a
version where it was implemented for two augmentations (FrequencyShift
and TimeReverse) so that we can discuss and converge into something
before we implement it for the other augmentations.

Let me know what you think :)

robintibor:
Yeah I was also not too much a fan of all these arguments for a class with such a generic name like `Transformer` with then many of these times the arguments are actually not used. I accepted it as they were just optional arguments, so one could also not supply them and get reasonable behavior. But I also don't mind not having them haha :) Given now the only remaining argument is this probability, one could also more explicitly name the class 'StochasticTransform`? Or even have some hierarchy `Transform` -> `StochasticTransform` -> [...] with some other Transforms inherting directly from transform if they are not meant to be optional? But to me, seems also not crucial in case you feel any Transform should possibly be turned on or off, one could also stick with current verison (and as said possibly rename to `StochasticTransform`or similar if you agree). Is the meaning of `probability` argument consistent now that it is used to make a binary decision whether to apply the transform to the batch? or is it per-example decision? In any case should be consistent across Transforms to ensure no surprises.

tomMoral:
Not a big fan of `StochasticTransform`, I think it is fine to keep the `Transform` if we get by default `probability=1`.
If a user change the `probability`, he probably know what it means, else, this is a deterministic transform.

`Stochastic` makes it unclear to me.

robintibor:
Ok also fine to me then :)

braindecode/augmentation/transforms.py
```diff
@@ -895,4 +890,17 @@ def __init__(
...
+            if self.beta_per_sample:
+                lam = torch.as_tensor(
+                    self.rng.beta(self.alpha, self.alpha, batch_size)
+                ).to(device)
+            else:
+                lam = torch.ones(batch_size).to(device)
+                lam *= self.rng.beta(self.alpha, self.alpha)
+        else:
+            lam = torch.ones(batch_size).to(device)
+        return lam, self.rng
```
sbbrandt, line 906: Great!

If we don't want to pass the 'rng', we could also pass the permutated index to the functional.

cedricrommel: Indeed. It would even make more sense to pass (X1, y1) and (X2, y2) wouldn't it? So that the functional does something really atomic and fully deterministic. The transform could handle the sampling of the permuted index and use it to get (X2, y2) from (X1, y1). What do you think ?

sbbrandt: Doesn't `functional` always gets `X` and `y`? We would need to pass `lam` anyway, so `X, y, lam, idx_perm` would be cleaner, I guess.

cedricrommel: I'm ok with that too :) Go for it

braindecode/augmentation/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+Utilities for data augmentation.
+"""
+from .base import Transform, IdentityTransform, Compose, AugmentedDataLoader
+from .transforms import TimeReverse, FrequencyShift
+# from .transforms import (
+#     TimeReverse, SignFlip, DownsamplingShift, FTSurrogate, ShuffleChannels,
+#     MissingChannels, GaussianNoise, ChannelSymmetry, TimeMask,
+#     BandstopFilter, FrequencyShift, RandomSensorsRotation, RandomZRotation,
+#     RandomYRotation, RandomXRotation,
+# )
```
robintibor, line 11: These ones will be readded?

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,204 @@
...
+        return torch.as_tensor(
+            self.probability > self.rng.uniform(size=batch_size)
+        )
+
+    @property
+    def probability(self):
+        return self._probability
+
+    def to_dict(self):
+        """Returns a dictionary describing the transform """
```
robintibor, line 112: Maybe describe a bit more here, what is inside the dictionary and what is this function good for like why does it exist, what is it used for?

cedricrommel: Good question ! It's legacy code that I will remove :)

braindecode/augmentation/__init__.py
cedricrommel: Already have been yes

braindecode/augmentation/transforms.py
```diff
@@ -840,4 +840,7 @@ def get_params(self, X, y):
                 lam *= self.rng.beta(self.alpha, self.alpha)
         else:
             lam = torch.ones(batch_size).to(device)
-        return lam, self.rng
+        
+        idx_perm = torch.as_tensor(self.rng.permutation(batch_size,))
+
+        return lam, idx_perm
```
cedricrommel, line 838: Great ! :)

braindecode/augmentation/base.py
robintibor: ok great waiting for you to remove before merging then

braindecode/augmentation/base.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def to_dict(self):
+        """Returns a DataFrame describing the transforms making the object"""
+        structure = list()
+        for i, transform in enumerate(self.transforms):
+            transform_struct = transform.to_dict()
+            transform_struct.update({"transform_idx": i})
+            structure.append(transform_struct)
+        res = pd.DataFrame(structure)
+        return res
```
robintibor, line 156: but this one is needed or also legacy to remove?

cedricrommel: Hmm, I had forgotten this. It is hence indeed useful in this context to know what is contained in a `Compose` transform (although not critical).

cedricrommel: I'll just write better docstrings then.

braindecode/training/losses.py
```diff
@@ -19,3 +19,48 @@ def forward(self, preds, targets):
...
+    Implementation based on [2]_.
+
+    Parameters
+    ----------
+    preds : torch.Tensor
+        Predictions from the model.
+    target : torch.Tensor | list of torch.Tensor
+        For predictions without mixup, the targets as a tensor. If mixup has
+        been applied, a list containing the targets of the two mixed
+        samples and the mixing coefficients as tensors.
```
robintibor, line 36: This seems confusing to me that this is under Parameters. Parameters I would expect it is parameters for constructor of the class, not the arguments of the `__call__` function. This is also like that in https://pytorch.org/docs/stable/generated/torch.nn.BCEWithLogitsLoss.html where the parameters are really class constructor arguments. Canw e somehow call this differently? or move parameter part doc to the loss_function?

robintibor (changes requested):

braindecode/augmentation/base.py
cedricrommel: Ok, after discussion with @tomMoral , we've agreed that even though it is a "kinda cool" feature, it has not it's place in the API. So I just removed both `to_dict` methods :)

braindecode/training/losses.py
cedricrommel: Ping @sbbrandt ☝️

robintibor: Can we may be just fix it together tomorrow? @sbbrandt  might be busy with master thesis @cedricrommel

robintibor: and this is only thing keeping me from merge atm

sbbrandt: Yes, I agree. The parameter doc should probably be moved to the static loss function.

agramfort: I made this a plain function but I realize this is not test for this...

@sbbrandt can you push one test for mixup_criterion ?

agramfort:
I made it a plain function

>

braindecode/training/losses.py
sbbrandt: I implemented a basic test for the 'mixup_criterion'.

Should there be added anything to the test @agramfort ?

test/unit_tests/training/test_losses.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def test_mixup_criterion():
+    N_classes = 2
+    N_samples = 5
+    y_a = torch.zeros(N_samples, dtype=torch.int64)
+    y_b = torch.ones(N_samples, dtype=torch.int64)
+    lam = torch.arange(.1, 1, 1/N_samples)
+
+    preds = torch.rand((N_samples, N_classes))
```
agramfort, line 16:
```suggestion
    n_classes = 2
    n_samples = 5
    y_a = torch.zeros(N_samples, dtype=torch.int64)
    y_b = torch.ones(N_samples, dtype=torch.int64)
    lam = torch.arange(.1, 1, 1/N_samples)

    preds = torch.rand((n_samples, n_classes))
```

agramfort: @sbbrandt can you address this?

sbbrandt: Already was implemented by Thomas now.

cedricrommel:
Robin had changed to [WIP] because the example is still missing @agramfort

robintibor:
> Robin had changed to [WIP] because the example is still missing @agramfort

exactly

agramfort:
grrrr ok !

@cedricrommel you are on the example now then?

>

robintibor:
yes then he can change to [MRG] when example is there so I see it as well.

examples/plot_data_augmentation.py
```diff
@@ -0,0 +1,232 @@
...
+]
+
+preprocess(dataset, preprocessors)
+
+######################################################################
+# Cut Compute Windows
+# ~~~~~~~~~~~~~~~~~~~
+#
+
+from braindecode.preprocessing.windowers import create_windows_from_events
```
agramfort, line 56:
```suggestion
from braindecode.preprocessing import create_windows_from_events
```

should be enough

examples/plot_data_augmentation.py
```diff
@@ -0,0 +1,232 @@
...
+
+# Next, I will manipulate one session to show the resulting frequency shift. I
+# will manipulate the data of an mne Epoch to make usage of mne functions.
+
+import torch
+
+epochs = train_set.datasets[0].windows
+epochs_tr = epochs.copy()
+
+X = torch.Tensor(epochs_tr._data)
```
agramfort, line 116:
```suggestion
X = torch.Tensor(epochs_tr.get_data())
```
don't hack private attributes !

examples/plot_data_augmentation.py
```diff
@@ -0,0 +1,232 @@
...
+X_tr = transform.forward(X)
+epochs_tr._data = X_tr
```
agramfort, line 118: 😱 

hacking private attributes is evil. If you to do this use https://mne.tools/stable/generated/mne.Epochs.html#mne.Epochs.apply_function

examples/plot_data_augmentation.py
```diff
@@ -0,0 +1,232 @@
...
+device = 'cuda' if cuda else 'cpu'
+if cuda:
+    torch.backends.cudnn.benchmark = True
+seed = 20200220  # random seed to make results reproducible
+# Set random seed to be able to reproduce results
+set_random_seeds(seed=seed, cuda=cuda)
+
+n_classes = 4
+# Extract number of chans and time steps from dataset
+n_chans = train_set[0][0].shape[0]
```
agramfort, line 166:
```suggestion
n_channels = train_set[0][0].shape[0]
```

examples/plot_data_augmentation.py
robintibor: anyways are you there setting `_data` to torch tensor? would be better to to be numpy array maybe as well?

sbbrandt: Yes, you are right. This should be done via apply_function. I get an error doing this and need to go now. I will fix this tomorrow!

robintibor, line 118:
```suggestion
epochs_tr.apply_function(lambda x: transform.forward(torch.Tensor(x)).numpy())
```
Would this work @agramfort ?

cedricrommel: `forward` should not be called. Just transform(x) should work

robintibor: ah right yes

agramfort: Yes it should

cedricrommel:
I've fixed things. Will push soon. Just making a final check at the html :)

cedricrommel:
Ok Alles gut for me ! 😃 I just need confirmation from @sbbrandt that he is ok with my commit where I put his email address in the file to add him as contributor and then I'm ok for merging 💪

agramfort:
@robintibor green button is yours !

robintibor (approved):

robintibor:
Great merged! @sbbrandt  mail is already in other parts of the repo, so should be fine. 
Here a celebratory gif:
album_1f9292s0f

agramfort:
🎉🎉🎉🎉🎉

cedricrommel:
Niiiice ! 👏 👏 👏 👏 
Thank you both for the really nice event, as well as @hubertjb 💪

sbbrandt:
🎉🎉🎉

Thank you guys! Also for critics and correction of errors.

From me also thanks for the event and thank you @cedricrommel for the nice augmentation implementation :)

PR #258, opened 2021-06-21
pimp doc + fix some sphinx warnigns

agramfort:


braindecode/datasets/base.py
```diff
@@ -214,7 +215,7 @@ def get_metadata(self):
 
         Returns
         -------
-        pd.DataFrame:
+        metadata: pd.DataFrame
```
tomMoral, line 218:
```suggestion
        metadata : pd.DataFrame
```

braindecode/datasets/base.py
```diff
@@ -166,20 +166,21 @@ def split(self, by=None, property=None, split_ids=None):
...
             Some property which is listed in info DataFrame.
-        split_ids: list(int) | list(list(int))
+        split_ids: list
             List of indices to be combined in a subset.
+            It can be a list of int or a list of list of int.
 
         Returns
         -------
-        splits: dict{str: BaseConcatDataset}
+        splits: dict
```
tomMoral, line 183:
```suggestion
        splits : dict
```

braindecode/datasets/base.py
```diff
@@ -166,20 +166,21 @@ def split(self, by=None, property=None, split_ids=None):
...
+            If ``by`` is a string, splitting is performed based on the description
             DataFrame column with this name.
-            If by is a (list of) list of integers, the position in the first
+            If ``by`` is a (list of) list of integers, the position in the first
             list corresponds to the split id and the integers to the
             datapoints of that split.
         property: str
             Some property which is listed in info DataFrame.
-        split_ids: list(int) | list(list(int))
+        split_ids: list
```
tomMoral, line 177:
```suggestion
        split_ids : list
```

braindecode/datasets/base.py
```diff
@@ -166,20 +166,21 @@ def split(self, by=None, property=None, split_ids=None):
 
         Parameters
         ----------
-        by: str | list(int) | list(list(int))
-            If by is a string, splitting is performed based on the description
+        by: str | list
```
tomMoral, line 169:
```suggestion
        by : str | list
```

examples/plot_dataset_example.py
```diff
@@ -64,7 +63,7 @@
 # a target y, and window_ind (which itself contains `i_window_in_trial`,
 # `i_start_in_trial`, and `i_stop_in_trial`, which are required for combining
 # window predictions in the scorer).
-for x, y, window_ind in windows_ds:
+for x, y, window_ind in windows_dataset:
```
tomMoral, line 66: Feels weird that two `datasets` do not have the same API by default.
I think it would be nice to have an option that makes `windows_dataset` output `window_ind` but that is `False` by default no?

tomMoral (approved):
LGTM! just a few comments.

robintibor:
LGTM, waiting for comments to be addressed

examples/plot_dataset_example.py
robintibor: what other dataset do you mean?

agramfort:
good to go from my end @robintibor

examples/plot_dataset_example.py
tomMoral: The non-window one, for instance `MOABBDataset` or the one from `pytorch`.
I feel like using the `window_ind` is for specific use cases and thus should only be return when the user request it.

But maybe I do not understand completely this part, happy to discuss if you want.

robintibor (approved):
Amazing great

Issue #246, opened 2021-06-17
Align windowing functions argument names


gemeinl:
Although technically correct, `create_windows_from_events` accepts `trial_start_offset_samples` and `trial_stop_offset_samples` whereas `create_fixed_length_windows` accepts `start_offset_samples` and `stop_offset_samples`, since no trials exist.
This might be confusing to users or it could complicate switching from one task to another. Maybe both functions could just accept `start_offset_samples` and `stop_offset_samples`.

dcwil:
@vytjan, @Ann-KathrinKiessner, @HenrikBons & I are working on this

PR #260, opened 2021-06-21
[WIP] Remove "trial_" prefix, issue 246

dcwil:
Fix for [246](https://github.com/braindecode/braindecode/issues/246):

"trial_" prefix removed from "trial_start_offset_samples" & "trial_stop_offset_samples" . Effects create_windows_from_events and create_from_mne_raw

changes by @Ann-KathrinKiessner, @vytjan, @HenrikBons & I

### DIFF
--- a/braindecode/datasets/mne.py
+++ b/braindecode/datasets/mne.py
@@ -11,7 +11,7 @@ from .base import BaseDataset, BaseConcatDataset, WindowsDataset
 
 
 def create_from_mne_raw(
-        raws, trial_start_offset_samples, trial_stop_offset_samples,
+        raws, start_offset_samples, stop_offset_samples,
         window_size_samples, window_stride_samples, drop_last_window,
         descriptions=None, mapping=None, preload=False, drop_bad_windows=True):
     """Create WindowsDatasets from mne.RawArrays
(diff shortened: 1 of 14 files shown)


robintibor:
Due to other PRs being merged, you probably have to rebase this from master, see end of the Contributing guide https://github.com/braindecode/braindecode/blob/765cd046a63f051a96a9019a5322473b4a501129/CONTRIBUTING.md

agramfort:
you need to fix conflicts @dcwil

robintibor:
@hubertjb @sliwy how do you feel about this change do we want it?
Change is renaming
`trial_start_offset_samples` and `trial_stop_offset_samples`
to 
`start_offset_samples` and `stop_offset_samples`
in `create_windows_from_events` to make it same name as in `create_from_mne_raw`

I mean variables do not mean same things but are similar, do we want them to have same name there?

sliwy:
It's good to have similar API in both functions. However we should clarify all the doubts in the docstring. 
If we can crop signal with Preprocessors, let's do that in all tutorials and then remove those parameters from `create_fixed_length_windows`, so we have no confusion there.

robintibor:
so @dcwil can you remove start_offset_samples and stop_offset_samples from create_fixed_length_windows ? also from _create_fixed_length_windows of course, just remove entire usage, so as if they were set to 0. in one test it is set to 100, but I think test should still just run fine when the argument is removed.

gemeinl:
Closed in favor of https://github.com/braindecode/braindecode/issues/269. Will be done in separate PR.

PR #262, opened 2021-06-21
[MRG] Description get set

gemeinl:


braindecode/datasets/base.py
```diff
@@ -84,6 +84,28 @@ def transform(self, value):
...
         self._transform = value
 
+    @property
+    def description(self):
+        return self._description
+
+    def set_description(self, description, overwrite=False):
+        """Update (add or overwrite) the dataset description.
+
+        Parameters
```
robintibor, line 94: overwrite is not present in parameter docstring

robintibor: Another option is not to have overwrite and use property setters and always overwrite... like https://github.com/braindecode/braindecode/blob/765cd046a63f051a96a9019a5322473b4a501129/braindecode/datasets/base.py#L77-L81 has property setters... but maybe this version safer?

braindecode/datasets/base.py
```diff
@@ -141,6 +163,28 @@ def transform(self, value):
...
```
robintibor, line 173: is this code same as in BaseDataset? then atm it is inherited anyways right?

braindecode/datasets/base.py
robintibor: in this version if you do 
```python
dataset.description = whatever
```
it will just throw an error? or what happens?

gemeinl: Well, do we want to always overwrite?

braindecode/datasets/base.py
gemeinl: Yes, code is the same and currently inherited. Same goes for the `transform` stuff. So either we inherit and remove all code duplicates or don't inherit and have the code in both classes.

braindecode/datasets/base.py
gemeinl: Gives `AttributeError: can't set attribute`.

braindecode/datasets/base.py
gemeinl: Earlier the tendency was to not inherit. So I will keep the code?

braindecode/datasets/base.py
gemeinl: I don't think we want to allow dataset.description = whatever, since it will destroy everything. The description should always be a pandas.Series / pandas.DataFrame. Also, my overwrite refers to individual columns in the Series/DataFrame that should be erplaced, not the entire object.

braindecode/datasets/base.py
robintibor: Atm I would tend towards making 3 classes:
1. BaseDataset (common functionality, doesn't get any MNE object passed, __getitem__ raises NotImplementedError)
2. WindowsDataset 
3. ContinousDataset (or RawDataset, so gets raws)
Then we could prevent duplicating code and structure would be clean to understand I think.
What do you think?

robintibor: Since it may create some conflicts with other PR https://github.com/braindecode/braindecode/pulls , we could also do this later in a separate PR

braindecode/datasets/base.py
```diff
@@ -290,3 +338,26 @@ def save(self, path, overwrite=False):
         if concat_of_raws:
             json.dump({'target_name': target_name}, open(target_file_name, 'w'))
         self.description.to_json(description_file_name)
+
+    @property
+    def description(self):
+        df = pd.DataFrame([ds.description for ds in self.datasets])
+        df.reset_index(inplace=True, drop=True)
+        return df
```
robintibor, line 346: Is this normally called during a training loop? If so, we should probably check if we get slowdowns from it. If not called during training loop, should not be a problem

robintibor: I assume it is not called?

agramfort (approved):
LGTM

braindecode/datasets/base.py
gemeinl: No, not called during training loop.

robintibor:
So now this would be ready to merge? @gemeinl

robintibor:
Great thanks merged!

Issue #170, opened 2020-09-23
Add TIDNet Model


robintibor:
From https://github.com/SPOClab-ca/ThinkerInvariance/blob/master/models.py and https://github.com/SPOClab-ca/ThinkerInvariance/blob/master/layers.py
Paper: https://iopscience.iop.org/article/10.1088/1741-2552/abb7a7/meta
Then we can compare against it also.

robintibor:
Implement this into models, also make a test for it

PR #265, opened 2021-06-22
[MRG] Added TIDNet

HenrikBons:
As per issue #170
changes by @Ann-KathrinKiessner, @vytjan, @HenrikBons & @dcwil

### DIFF
--- a/braindecode/models/tidnet.py
+++ b/braindecode/models/tidnet.py
@@ -226,7 +226,7 @@ class TIDNet(nn.Module):
         self.input_window_samples = input_window_samples
         self.temp_len = ceil(temp_span * input_window_samples)
 
-        self.dscnn = _tidnet_features(s_growth=s_growth, t_filters=t_filters, in_chans=in_chans,
+        self.dscnn = _TIDNetFeatures(s_growth=s_growth, t_filters=t_filters, in_chans=in_chans,
                                       input_window_samples=input_window_samples,
                                       drop_prob=drop_prob, pooling=pooling, temp_layers=temp_layers,
                                       spat_layers=spat_layers, temp_span=temp_span,
--- a/braindecode/models/tidnet.py
+++ b/braindecode/models/tidnet.py
@@ -1,0 +1,266 @@
+from torch import nn
+from math import ceil
+from torch.nn import init
+from torch.nn.utils import weight_norm
+from .modules import Ensure4d, Expression
+
+import torch
+
+
+class Expand(nn.Module):
+    def __init__(self, axis=-1):
+        super().__init__()
+        self.axis = axis
+
+    def forward(self, x):
+        return x.unsqueeze(self.axis)
+
+
+class Flatten(nn.Module):
+    def forward(self, x):
+        return x.contiguous().view(x.size(0), -1)
+
+
+class BatchNormZG(nn.BatchNorm2d):
+    def reset_parameters(self):
+        if self.track_running_stats:
+            self.running_mean.zero_()
+            self.running_var.fill_(1)
+        if self.affine:
+            self.weight.data.zero_()
+            self.bias.data.zero_()
+
+
+class ConvBlock2D(nn.Module):
+    """Implements Convolution block with order:
+    Convolution, dropout, activation, batch-norm
+    """
+    def __init__(self, in_filters, out_filters, kernel, stride=(1, 1), padding=0, dilation=1,
+                 groups=1, drop_prob=0.5, batch_norm=True, activation=nn.LeakyReLU, residual=False):
+        super().__init__()
+        self.kernel = kernel
+        self.activation = activation()
+        self.residual = residual
+
+        self.conv = nn.Conv2d(in_filters, out_filters, kernel, stride=stride, padding=padding,
+                              dilation=dilation, groups=groups, bias=not batch_norm)
+        self.dropout = nn.Dropout2d(p=drop_prob)
+        self.batch_norm = BatchNormZG(out_filters) if residual else nn.BatchNorm2d(out_filters) if\
+            batch_norm else lambda x: x
+
+    def forward(self, input, **kwargs):
+        res = input
+        input = self.conv(input, **kwargs)
+        input = self.dropout(input)
+        input = self.activation(input)
+        input = self.batch_norm(input)
+        return input + res if self.residual else input
+
+
+class DenseFilter(nn.Module):
+    def __init__(self, in_features, growth_rate, filter_len=5, drop_prob=0.5, bottleneck=2,
+                 activation=nn.LeakyReLU, dim=-2):
+        super().__init__()
+        dim = dim if dim > 0 else dim + 4
+        if dim < 2 or dim > 3:
+            raise ValueError('Only last two dimensions supported')
+        kernel = (filter_len, 1) if dim == 2 else (1, filter_len)
+
+        self.net = nn.Sequential(
+            nn.BatchNorm2d(in_features),
+            activation(),
+            nn.Conv2d(in_features, bottleneck * growth_rate, 1),
+            nn.BatchNorm2d(bottleneck * growth_rate),
+            activation(),
+            nn.Conv2d(bottleneck * growth_rate, growth_rate, kernel,
+                      padding=tuple((k // 2 for k in kernel))),
+            nn.Dropout2d(drop_prob)
+        )
+
+    def forward(self, x):
+        return torch.cat((x, self.net(x)), dim=1)
+
+
+class DenseSpatialFilter(nn.Module):
+    def __init__(self, in_chans, growth, depth, in_ch=1, bottleneck=4, drop_prob=0.0,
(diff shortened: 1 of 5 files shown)


braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
+from torch import nn
+from math import ceil
+from torch.nn import init
+from torch.nn.utils import weight_norm
+from .modules import Ensure4d, Expression
+
+import torch
+
+
+class Expand(nn.Module):
```
robintibor, line 10: Is this unused? Than maybe remove it?

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+    def __init__(self, axis=-1):
+        super().__init__()
+        self.axis = axis
+
+    def forward(self, x):
+        return x.unsqueeze(self.axis)
+
+
+class Flatten(nn.Module):
```
robintibor, line 19: there is nn.Flatten from PyTorch, do you know why this is needed? try using nn.Flatten instead (with start_dim=1)

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+    def forward(self, x):
+        return x.contiguous().view(x.size(0), -1)
+
+
+class BatchNormZG(nn.BatchNorm2d):
```
robintibor, line 24: Maybe make this private with underscore, otherwise docstring would be necessary

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+    def reset_parameters(self):
+        if self.track_running_stats:
+            self.running_mean.zero_()
+            self.running_var.fill_(1)
+        if self.affine:
+            self.weight.data.zero_()
+            self.bias.data.zero_()
+
+
+class ConvBlock2D(nn.Module):
```
robintibor, line 34: Same here private with underscore

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+    def forward(self, input, **kwargs):
+        res = input
+        input = self.conv(input, **kwargs)
+        input = self.dropout(input)
+        input = self.activation(input)
+        input = self.batch_norm(input)
+        return input + res if self.residual else input
+
+
+class DenseFilter(nn.Module):
```
robintibor, line 60: Also private

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+            nn.Conv2d(bottleneck * growth_rate, growth_rate, kernel,
+                      padding=tuple((k // 2 for k in kernel))),
+            nn.Dropout2d(drop_prob)
+        )
+
+    def forward(self, x):
+        return torch.cat((x, self.net(x)), dim=1)
+
+
+class DenseSpatialFilter(nn.Module):
```
robintibor, line 84: Also private

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+    def forward(self, x):
+        if len(x.shape) < 4:
+            x = x.unsqueeze(1).permute([0, 1, 3, 2])
+        x = self.net(x)
+        if self.collapse:
+            return self.channel_collapse(x).squeeze(-2)
+        return x
+
+
+class TemporalFilter(nn.Module):
```
robintibor, line 106: Also private

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,266 @@
...
+    Code adapted from: https://github.com/SPOClab-ca/ThinkerInvariance/blob/master/
+
+    References
+    ----------
+    .. [TIDNet] Kostas, D. & Rudzicz, F.
+        Thinker invariance: enabling deep neural networks for BCI across more people.
+        J. Neural Eng. 17, 056008 (2020).
+        doi: 10.1088/1741-2552/abb7a7.
+    """
+    def __init__(self, in_chans, n_classes, input_window_samples, s_growth=24, t_filters=32,
```
robintibor, line 234: you can probably remove the **kwargs?

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,251 @@
+from torch import nn
+from math import ceil
+from torch.nn import init
+from torch.nn.utils import weight_norm
+from .modules import Ensure4d, Expression
+
+import torch
```
agramfort, line 7:
```suggestion
from math import ceil

import torch
from torch import nn
from torch.nn import init
from torch.nn.utils import weight_norm

from .modules import Ensure4d, Expression
```

import order should be from standard to specific ie std lib -> local relative imports

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,251 @@
...
+    def forward(self, x):
+        if self.residual_style.lower() == 'netwise':
+            return self.net(x) + self.residual(x)
+        elif self.residual_style.lower() == 'dense':
+            for layer in self.net:
+                x = torch.cat((x, layer(x)), dim=1)
+            return x
+
+
+class _tidnet_features(nn.Module):
```
agramfort, line 125: class names should be CamelCase
```suggestion
class _TIDNetFeatures(nn.Module):
```

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,251 @@
...
+        filter of length 75.
+    bottleneck : int
+        Bottleneck factor within Densefilter
+    summary : int
+        Output size of AdaptiveAvgPool1D layer. If set to -1, value will be calculated
+        automatically (input_window_samples // pooling).
+
+    Notes
+    -----
+    Code adapted from: https://github.com/SPOClab-ca/ThinkerInvariance/blob/master/
```
agramfort, line 210:
```suggestion
    Code adapted from: https://github.com/SPOClab-ca/ThinkerInvariance/
```

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,251 @@
...
+
+        self.classify = self._create_classifier(self.num_features, n_classes)
+
+    def _create_classifier(self, incoming, n_classes):
+        classifier = nn.Linear(incoming, n_classes)
+        init.xavier_normal_(classifier.weight)
+        classifier.bias.data.zero_()
+        return nn.Sequential(nn.Flatten(start_dim=1), classifier, nn.LogSoftmax(dim=-1))
+
+    def forward(self, x, **kwargs):
```
agramfort, line 244: docstring of public method missing

robintibor: do we really want docstring for forward methods? E.g., PyTorch doesn't do this: 
https://github.com/pytorch/pytorch/blob/bac4cfd54d44aa0fbc574e6561b878cb406762cc/torch/nn/modules/conv.py#L442

Neither does torchvision:
https://github.com/pytorch/vision/blob/d391a0e992a35d7fb01e11110e2ccf8e445ad8a0/torchvision/models/densenet.py#L215

robintibor: but I don't think we need **kwargs here? this could be removed?

agramfort: I would say here we could document the shapes of x for this network module? but I will not fight hard for it

robintibor: ok since you're not gonna fight for it and I think we should always have same accepted input shapes: #examples x #EEGchannels x #time (and optionally except empty last dim), we can just keep without docstring but remove **kwargs

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,252 @@
...
+        self.activation = activation()
+        self.residual = residual
+
+        self.conv = nn.Conv2d(in_filters, out_filters, kernel, stride=stride, padding=padding,
+                              dilation=dilation, groups=groups, bias=not batch_norm)
+        self.dropout = nn.Dropout2d(p=drop_prob)
+        self.batch_norm = _BatchNormZG(out_filters) if residual else nn.BatchNorm2d(out_filters) if\
+            batch_norm else lambda x: x
+
+    def forward(self, input, **kwargs):
```
robintibor, line 38:
```suggestion
    def forward(self, input, **kwargs):
```
```suggestion
    def forward(self, input):
```

braindecode/models/tidnet.py
```diff
@@ -0,0 +1,252 @@
...
+        res = input
+        input = self.conv(input, **kwargs)
```
robintibor, line 40:
```suggestion
        input = self.conv(input,)
```

robintibor:
two more **kwargs to go and we should be ready to merge!

robintibor:
@kostasde Just for your information: We have also integrated TIDNet into braindecode, we made some adaptations to the code to fit our assumptions about input shapes etc., but hopefully it is all still functionally equivalent :) Thanks again for having your code online :)

kostasde:
This is great, thanks for your efforts and for letting me know folks!

Issue #242, opened 2021-06-16
How to draw result graphs？


MUMULEO:
How to draw the confusion matrix  and some other result graphs in the paper？

robintibor:
Great question! We may try to include some confusion matrices into the tutorials next week! You refer to https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.23730 ?

robintibor:
So task here could be to add a confusion matrix to one of the tutorials at the end.

MUMULEO:
> Great question! We may try to include some confusion matrices into the tutorials next week! You refer to https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.23730 ?

yes, I refer to this paper, i don't know how to plot some result graphs in this paper.   Thanks for your reply !

MUMULEO:
> So task here could be to add a confusion matrix to one of the tutorials at the end.

please, i want to know when will the confusion matrix be added to one of the tutorials?

robintibor:
not sure yet, we may do in coming days.

Some code for those that attempt to try it, that may help a bit:


```python
def plot_confusion_matrix_paper(confusion_mat, p_val_vs_csp,
                                p_val_vs_other_net,
                                class_names=None, figsize=None, colormap=cm.bwr,
                                textcolor='black', vmin=None, vmax=None,
                                fontweight='normal',
                                rotate_row_labels=90,
                                rotate_col_labels=0,
                                with_f1_score=False,
                                norm_axes=(0, 1),
                                rotate_precision=False):
    # TODELAY: split into several functions
    # transpose to get confusion matrix same way as matlab
    confusion_mat = confusion_mat.T
    # then have to transpose pvals also
    p_val_vs_csp = p_val_vs_csp.T
    p_val_vs_other_net = p_val_vs_other_net.T
    n_classes = confusion_mat.shape[0]
    if class_names is None:
        class_names = [str(i_class + 1) for i_class in xrange(n_classes)]

    # norm by number of targets (targets are columns after transpose!)
    # normed_conf_mat = confusion_mat / np.sum(confusion_mat,
    #    axis=0).astype(float)
    # norm by all targets
    normed_conf_mat = confusion_mat / np.float32(np.sum(confusion_mat, axis=norm_axes, keepdims=True))

    fig = plt.figure(figsize=figsize)
    plt.clf()
    ax = fig.add_subplot(111)
    ax.set_aspect(1)
    if vmin is None:
[...]

PR #274, opened 2021-06-24
[WIP] Added a confusion matrix plot generator.

vytjan:
refering to #242 
@Ann-KathrinKiessner @dcwil @HenrikBons

### DIFF
--- a/braindecode/visualization/confusion_matrices.py
+++ b/braindecode/visualization/confusion_matrices.py
@@ -1,0 +1,245 @@
+# Authors: Robin Schirrmeister <robintibor@gmail.com>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib import cm
+
+
+def plot_confusion_matrix(confusion_mat,
+                          class_names=None, figsize=None,
+                          colormap=cm.bwr,
+                          textcolor='black', vmin=None, vmax=None,
+                          fontweight='normal',
+                          rotate_row_labels=90,
+                          rotate_col_labels=0,
+                          with_f1_score=False,
+                          norm_axes=(0, 1),
+                          rotate_precision=False,
+                          class_names_fontsize=12):
+    """
+
+    Generates a confusion matrix with additional precision and sensitivity metrics as in [1]_.
+
+    Parameters
+    ----------
+    confusion_mat: 2d numpy array
+        A confusion matrix, e.g. sklearn confusion matrix:
+        https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html
+    class_names: array, optional
+        List of classes/targets.
+    figsize: tuple, optional
+        Size of the generated confusion matrix figure.
+    colormap: matplotlib cm colormap, optional
+    textcolor: str, optional
+        Color  of the text in the figure.
+    vmin, vmax: float, optional
+        The data range that the colormap covers.
+    fontweight: str, optional
+        Weight of the font in the figure:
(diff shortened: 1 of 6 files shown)


agramfort:
why not just using https://scikit-learn.org/stable/modules/generated/sklearn.metrics.ConfusionMatrixDisplay.html#sklearn.metrics.ConfusionMatrixDisplay ?

robintibor:
because ours much more beautiful! ;) See discussion end of https://github.com/braindecode/braindecode/issues/242 so it was considered ! But I would try to add the class names in the example

robintibor:
We agreed that we can keep like this (with own function that shows a bit more information than scikitlearn) depending on how strong @agramfort objects to it :D

agramfort:
I won't strongly object I am just more lazy than you ;)

>

examples/plot_bcic_iv_2a_moabb_cropped.py
```diff
@@ -343,3 +343,33 @@
...
+# get the targets
+y_true = valid_set.get_metadata().target
+y_pred = clf.predict(valid_set)
+
+# generating confusion matrix
+confusion_mat = confusion_matrix(y_true, y_pred)
+
+# add class labels
+label_dict = valid_set.datasets[0].windows.event_id.items()
+labels = list(dict(sorted(list(label_dict), key=lambda kv: kv[1])).keys())
```
robintibor, line 372:
```suggestion
# sort the labels by values (values are integer class labels)
labels = list(dict(sorted(list(label_dict), key=lambda kv: kv[1])).keys())
```

examples/plot_bcic_iv_2a_moabb_cropped.py
```diff
@@ -343,3 +343,33 @@
...
+# generate confusion matrices
+# get the targets
+y_true = valid_set.get_metadata().target
+y_pred = clf.predict(valid_set)
+
+# generating confusion matrix
+confusion_mat = confusion_matrix(y_true, y_pred)
+
+# add class labels
+label_dict = valid_set.datasets[0].windows.event_id.items()
```
robintibor, line 371:
```suggestion
# label_dict is class_name : str -> i_class : int
label_dict = valid_set.datasets[0].windows.event_id.items()
```

braindecode/visualization/confusion_matrices.py
```diff
@@ -0,0 +1,239 @@
...
+    plt.yticks(np.arange(0, height), class_names,
+               va='center',
+               fontsize=class_names_fontsize, rotation=rotate_row_labels)
+    plt.grid(False)
+    plt.ylabel('Predictions', fontsize=15)
+    plt.xlabel('Targets', fontsize=15)
+
+    # n classes is also shape of matrix/size
+    ax.text(-1.2, n_classes + 0.2, "Specificity /\nSensitivity", ha='center', va='center',
+            fontsize=13)
```
robintibor, line 178:
```suggestion
    ax.text(-1.2, n_classes + 0.2, "Recall", ha='center', va='center',
            fontsize=13)
```

Specificity/Sensitivity made sense for pathologies but for any classes then this value is recall (could also call sensitivity, but recall maybe better as no association with medicine / sensitivity and specificity)

robintibor: See also https://machinelearningmastery.com/precision-recall-and-f-measure-for-imbalanced-classification/ https://en.wikipedia.org/wiki/Precision_and_recall

robintibor:
And maybe just add same code also for trialwise example, why not?

robintibor:
Great work!

Issue #269, opened 2021-06-23
Remove `start_offset_samples` and `stop_offset_samples` argument in `create_fixed_length_windows`


gemeinl:
Instead or renaming the arguments as discussed in https://github.com/braindecode/braindecode/pull/260, we realized that these arguments do not make sense at all in the setting of `create_fixed_length_windows`. There, these arguments are equivalent to applying a preprocessing step that crops the signals in advance. We hence aim to remove the arguments at some point and deprecate them for now.

dcwil:
#273

PR #275, opened 2021-06-24
[MRG] Added non default arg warning for `start_offset_samples` & …

dcwil:
…`stop_offset_samples` in `create_fixed_length_windows`. Set default arguments to None. #269

### DIFF
--- a/braindecode/preprocessing/windowers.py
+++ b/braindecode/preprocessing/windowers.py
@@ -1,10 +1,13 @@
 """Get epochs from mne.Raw
 """
-
 # Authors: Hubert Banville <hubert.jbanville@gmail.com>
 #          Lukas Gemein <l.gemein@gmail.com>
 #          Simon Brandt <simonbrandt@protonmail.com>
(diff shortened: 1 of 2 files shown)


robintibor:
So now we want:

* `window_stride_samples` and `window_size_samples` and `drop_last_window` should be None by default
* in case `window_stride_samples` and `window_size_samples`are not `None` , but  `drop_last_window` is none raise an error and inform the user he should set drop_last_window.
* if all three are None, assume each window should span the entire recording
* if all three are None, but recordings are of different length, give a warning to the user that these different-length windows will not be aggregatable into a batch
* Finally restore argument order as before and keep warning for `start_offset_samples` / `stop_offset_samples` which are not None

gemeinl:
> if all three are None, but recordings are of different length, give a warning to the user that these different-length windows will not be aggregatable into a batch

This check is independent of `drop_last_window=None`. Raise the warning whenever we observe different length windows.

Issue #268, opened 2021-06-22
Standardizing the preprocessing pipeline for large datasets


hubertjb:
The pipeline we use in most of the examples is not well adapted to large datasets, as the operations are applied sequentially to all the recordings at once (sleep staging: https://braindecode.org/auto_examples/plot_sleep_staging.html, decoding on BCIC dataset: https://braindecode.org/auto_examples/plot_bcic_iv_2a_moabb_trial.html, MOABB: https://braindecode.org/auto_examples/plot_dataset_example.html, etc.). 

This means for instance that all the raw recordings must be loaded at once into memory. This is not practical on larger datasets such as the TUH EEG dataset, which is why @gemeinl made an example for large datasets specifically: https://braindecode.org/auto_examples/tuh_eeg_corpus.html

I'd like to standardize @gemeinl's approach so that we can recommend a "braindecode way" of handling large datasets and, if it's simple enough, maybe even apply it to all the examples so there's a unique way of doing things. Also, this is a good opportunity to think about how to best parallelize the preprocessing/windowing and serialize intermediate results (which I know has been a pain point for people in our lab).

I would suggest providing a helper function (called something like `prepare_dataset`) that automatically handles the parallelization and serialization. It would require a user-defined callable that loads and preprocesses/windows a **single** recording. For instance:

```python
def prepare_dataset(fn, ids, save_dir=None, n_jobs=1):
    """
    Parameters
    ----------
    fn : callable
        User-provided function that loads, preprocess, windows and preprocess a
	single recording.
    ids : dict
	Dictionary specifying which recordings to load from the dataset. Keys 
	specify the argument name to use in the dataset class and values 
	specify the actual values to load. E.g., to load subjects 0 and 1, 
	sessions 1 and 2 from the SleepPhysionet dataset:
	
		ids = {'subject_ids': [0, 1], 'recording_ids': [1, 2]}
		 
    save_dir : str | None
	If provided as string, processed recordings will be saved under that
	directory.
    return_dataset : bool
	If True, return the concatenated dataset.
    n_jobs : int
	Number of jobs.
	
    Returns
    -------
    BaseConcatDataset OR fnames
    """
    for each combination of ids:  # use joblib if n_jobs > 1
        ds = fn(id)
	serialize ds  # could be optional
	
    if return_dataset:
	reload files and concatenate in BaseConcatDataset
	return BaseConcatDataset
    else:
	return file names of serialized recordings
	
def load_and_preprocess_one_record(subject_id)
[...]

PR #277, opened 2021-06-25
[MRG] Parallelization and serialization functionality in `preprocess`

hubertjb:
This is work in progress toward the parallelization of the standard braindecode processing pipeline (see #268).

The current idea is to parallelize the preprocessing over the BaseDatasets with joblib (inside `preprocess`), and to optionally serialize the BaseDatasets so that they can be reloaded with `preload=False`. This should hopefully limit memory usage, which is critical when working with large datasets.

Next steps:
- Add windowing and window preprocessing steps in example
- Test memory usage of example
- Write tests
- Automatic serialization directory naming for smarter caching (like with joblib.memory)?

@gemeinl


braindecode/datautil/serialization.py
```diff
@@ -6,17 +6,129 @@
...
+            if os.path.isfile(f):
+                os.remove(f)
+
+    getattr(dataset, dataset_type).save(fname, overwrite=overwrite)
+    if dataset_type == 'raw':
+        json.dump({'target_name': dataset.target_name},
+                  open(target_fname, 'w'))
+    dataset.description.to_json(description_fname)
+
+    return fname, description_fname, target_fname
```
gemeinl, line 92: In an earlier discussion about `save_concat_dataset` @agramfort argued to offer serialization directly through the object and to only offer one way of serialization. So I would suggest to do the same here: remove `save_base_dataset` and to add its code to `BaseDataset.save`

braindecode/datautil/serialization.py
```diff
@@ -6,17 +6,129 @@
...
 
 import mne
+from mne.io.fiff import raw
 import pandas as pd
 
 from ..datasets.base import BaseDataset, BaseConcatDataset, WindowsDataset
 
 
+def create_fname_id_from_description(description):
+    # XXX: Enforce 'subject', 'recording' and/or 'run' in description?
```
gemeinl, line 22: I think currently we can't assume existence os 'recording', 'run' or 'session', since MOABB is inconsistent about this naming. I created a related issue here: https://github.com/NeuroTechX/moabb/issues/208

gemeinl:
I think we discussed it before, I do not like packing file names with information as in https://github.com/braindecode/braindecode/pull/277/files#diff-ffab4ab4433ba5e07e98af8ac01e96b531cca0eb9e7d1ffe2b24663551be030eR23-R29, since it quickly gets messy and to me it is unclear at which point to stop inclusion of information. Additionally, the information is already contained in `description.json` so putting it into the file name just creates another copy.

gemeinl:
If the aim of the file names was to have a nice way to only load a specific subject / recording / session / run later on, we could think about adding some smart functions to `load...` instead. It could probably offer an argument to do some selection and the loading function would first go through all the `description.json` files prior to actual loading.

gemeinl:
A `BaseDataset` is basically a special case of `BaseConcatDataset` with just one dataset. With `BaseConcatDataset([BaseDataset])` we have an easy conversion. I hence wonder if it was possible to entirely reuse the existing way of saving and loading of `BaseConcatDataset` also for individual `BaseDatasets`. If we decide to offer a `batch_size` argument to `preprocess` at some point, this would then also be consistent, since then we will have several `BaseConcatDataset` with `batch_size` many datasets..

braindecode/datautil/serialization.py
gemeinl: Also, how would this work for custom datasets?

braindecode/datautil/serialization.py
```diff
@@ -6,17 +6,129 @@
...
+        fname.append(f'run{description["run"]}')
+
+    if not fname:
+        raise ValueError(
+            'description must contain any of "subject", "recording" or "run".')
+
+    return '_'.join(fname)
+
+
+def save_base_dataset(dataset, path, overwrite=False):
```
agramfort, line 38: I would not expose anything else than the save method in BaseConcatDataset 

basically this needs to be private at best

braindecode/datautil/serialization.py
agramfort: yes !!!

hubertjb:
I modified `preprocess` to now use the `BaseConcatDataset` serializer. There is still a lot of cleanup to do and some additional logic to handle (e.g. what happens if you serialize the preprocessed data and the windowed data under the same directory?). 

With the current code I was able to confirm that a combination of parallelization and serialization is a good tradeoff in terms of memory usage and speed:
preproc_pipeline

For instance, using `n_jobs=8` and `save=True`, the pipeline took ~17 s to run and used 650 MiB, whereas a naive approach (`n_jobs=1`, `save=False`) was ~2-5 s faster, but took twice as much memory.

gemeinl:
Great to see this working! Awesome!

> I modified `preprocess` to now use the `BaseConcatDataset` serializer. There is still a lot of cleanup to do and some additional logic to handle (e.g. what happens if you serialize the preprocessed data and the windowed data under the same directory?).
Currently `.save()` will crash unless `overwrite = True` was specified. However, overwrite will delete all the `fif` files in the directory on first attempt of saving which might be a problem, since raw files get deleted before windows were cut?

I am currently rewriting the serialization logic a bit. I had trouble to parallelize `load_concat_dataset`, since the saving is inconsistent for `BaseDatasets` and `WindowsDatasets`... With the new logic, handling might become easier. Will open a separate PR on this soon.

hubertjb:
I made a nicer benchmarking example (although it doesn't start with `plot_` so it won't be run by the CI) to explain the parallelization/serialization tradeoff under `examples/benchmark_preprocessing.py`. I still get the same results as above (I just swapped the style/hue arguments in the figure):

preproc_pipeline

@gemeinl Should we wait for #288 to be merged before merging this one? Also, I was wondering whether we should make a version of your TUH example using this new preprocessing/serialization pipeline?

agramfort:
@gemeinl @robintibor are you happy here? I think this is really crucial features for braindecode attractiveness. Please don't neglect this. 🙏

gemeinl:
@agramfort Agreed, I think this functionality is very important. I am not yet happy, there are conflicts as well as open discussions.

@hubertjb There is no need to wait for https://github.com/braindecode/braindecode/pull/288 since it is independent of the changes here.

I don't think we should make a new TUH example but we should instead update it to use the fastest and most memory efficient code we have. It will be the go-to example using a very big dataset (because it is the only one).

braindecode/datasets/base.py
```diff
@@ -108,6 +107,12 @@ def set_description(self, description, overwrite=False):
                 self._description.pop(key)
         self._description = pd.concat([self.description, description])
 
+    def save(self, save_dir, overwrite=False):
+        """See braindecode.datautil.serialization.save_base_dataset.
+        """
+        from braindecode.datautil.serialization import save_base_dataset
```
gemeinl, line 113: Please move saving code here.

braindecode/preprocessing/preprocess.py
```diff
@@ -108,7 +113,8 @@ def __init__(self, fn, channel_wise=False, **kwargs):
                          **kwargs)
 
 
-def preprocess(concat_ds, preprocessors):
+def preprocess(concat_ds, preprocessors, save_dir=None, overwrite=False,
+               n_jobs=None):
```
gemeinl, line 117: I don't see a benefit in accepting `n_jobs=None` here. We could set default to 1, right?

braindecode/preprocessing/preprocess.py
```diff
@@ -130,34 +144,73 @@ def preprocess(concat_ds, preprocessors):
...
+
+
+def _preprocess(ds, ds_index, preprocessors, save_dir=None, overwrite=False):
     """Apply preprocessor(s) to Raw or Epochs object.
 
     Parameters
     ----------
-    raw_or_epochs: mne.io.Raw or mne.Epochs
-        Object to preprocess.
+    ds: BaseDataset
```
gemeinl, line 176: Can also be a WindowsDataset.

braindecode/preprocessing/preprocess.py
```diff
@@ -117,11 +123,19 @@ def preprocess(concat_ds, preprocessors):
...
+        files with the same name.
+    n_jobs : int | None
+        Number of jobs for parallel execution.
 
     Returns
     -------
-    BaseConcatDataset:
-        Preprocessed dataset.
+    None | BaseConcatDataset:
+        Preprocessed dataset, if inplace is False.
```
gemeinl, line 138: There is no inplace argument. I think we coupled this with `save_dir` when we created this table of different options?

braindecode/datasets/base.py
gemeinl: Ah, no. Please remove `save_base_dataset`. We don't need it anymore. We save on `BaseConcatDataset` level.

braindecode/preprocessing/preprocess.py
```diff
@@ -117,11 +123,19 @@ def preprocess(concat_ds, preprocessors):
         A concat of BaseDataset or WindowsDataset datasets to be preprocessed.
     preprocessors: list(Preprocessor)
         List of Preprocessor objects to apply to the dataset.
+    save_dir : str | None
+        If a string, the preprocessed data will be saved under the specified
+        directory.
+    overwrite : bool
+        When `save_dir` is provided, controls whether to overwrite existing
+        files with the same name.
```
gemeinl, line 131: This is not true. If `overwrite=True`, then all `fif` files (no matter whether epo or raw) in the directory are deleted as well as `target_name.json` and `description.json`. See https://github.com/braindecode/braindecode/blob/aecae2228af4f5d778c129154dfd7caeef3b11ab/braindecode/datasets/base.py#L320-L322.

gemeinl: I have the impression we can really get into trouble with this saving and overwriting. Maybe we should discuss a bit more how to do this in a save way. One thing I can think of would be to enforce writing to an empty directory. This could be inconvenient for the user though...

braindecode/preprocessing/preprocess.py
```diff
@@ -117,11 +123,19 @@ def preprocess(concat_ds, preprocessors):
...
+    n_jobs : int | None
```
gemeinl, line 132: See other comment about `n_jobs=None`.

braindecode/preprocessing/preprocess.py
```diff
@@ -117,11 +123,19 @@ def preprocess(concat_ds, preprocessors):
...
+        Number of jobs for parallel execution.
 
     Returns
     -------
-    BaseConcatDataset:
-        Preprocessed dataset.
+    None | BaseConcatDataset:
```
gemeinl, line 137: It still feels strange to me to have two different return types. What happens if we return the in-place modified object? Will it be copied?
I would like to always call 
`concat_ds = preprocess(concat_ds, ...)` no matter which arguments are provided. Is this possible? If not, I see the issue that a user might accidentally overwrite his dataset with None and has to repeat all the preprocessing...

braindecode/preprocessing/preprocess.py
```diff
@@ -130,34 +144,73 @@ def preprocess(concat_ds, preprocessors):
...
     """Apply preprocessor(s) to Raw or Epochs object.
 
     Parameters
     ----------
-    raw_or_epochs: mne.io.Raw or mne.Epochs
-        Object to preprocess.
+    ds: BaseDataset
+        Dataset object to preprocess.
+    ds_index : int
+        Index of the BaseDataset in its BaseConcatDataset. Ignored if save_dir
```
gemeinl, line 179: Or WindowsDataset

braindecode/preprocessing/preprocess.py
```diff
@@ -130,34 +144,73 @@ def preprocess(concat_ds, preprocessors):
...
+        is None.
     preprocessors: list(Preprocessor)
         List of preprocessors to apply to the dataset.
+    save_dir : str | None
+        If provided, save the preprocessed BaseDataset in the
```
gemeinl, line 184: See above

braindecode/preprocessing/preprocess.py
```diff
@@ -130,34 +144,73 @@ def preprocess(concat_ds, preprocessors):
...
+        specified directory.
+    overwrite : bool
+        If True, overwrite existing file with the same name.
```
gemeinl, line 187: Will delete evverything, not just the file with the same name.

braindecode/preprocessing/preprocess.py
```diff
@@ -130,34 +144,73 @@ def preprocess(concat_ds, preprocessors):
         assert hasattr(elem, 'apply'), (
             'Preprocessor object needs an `apply` method.')
```
gemeinl, line 146: Please add a check at the beginning of `preprocess` that crashes when `overwrite=False` and directory contains files. We don`t want to wait for a long time for the preprocessing to finish until it crashes when attempting to save.

braindecode/preprocessing/preprocess.py
hubertjb: My understanding is that if you provide `n_jobs=None` to joblib it can use the value set with a parallel_backend context manager: https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html By default this will be interpreted as 1.

braindecode/preprocessing/preprocess.py
hubertjb: Technically `WindowsDataset` is a `BaseDataset` :P But good point, I'll add it to be clearer.

braindecode/preprocessing/preprocess.py
hubertjb: Yes, thanks for catching this.

braindecode/preprocessing/preprocess.py
hubertjb: Right, we have to trust the user they will be careful... At least this should not overwrite raw data.

We could provide an additional "safer" option, so we would have the following possibilities:
1. Raise an error if there's already data under `save_dir`
2. Automatically create a new directory `save_dir + '_1'` (2, 3...) if the folder already exists
3. Overwrite

braindecode/preprocessing/preprocess.py
hubertjb: Actually right now the function doesn't return anything, the docstring was from a previous implementation. 

Good point about overwriting the dataset... I vote for always returning as well.

braindecode/preprocessing/preprocess.py
hubertjb: When you say everything, you mean `description.json` and `target_name.json`, right?

braindecode/preprocessing/preprocess.py
hubertjb: Great idea!

hubertjb:
@gemeinl I made a few changes to answer your comments. Let me know what you think of the other points (e.g. overwriting safety procedure).

Then we can also modify the TUH example to use this new pipeline.

gemeinl:
How do we handle entries in description of recordings that was changed through preprocessors?
So far, I manually updates the description after the preprocessing step (see https://github.com/braindecode/braindecode/blob/master/examples/plot_tuh_eeg_corpus.py#L214-L218). If `preprocess` now also handles the serialization, how would I do this prior to saving?

hubertjb:
> How do we handle entries in description of recordings that was changed through preprocessors?
> So far, I manually updates the description after the preprocessing step (see https://github.com/braindecode/braindecode/blob/master/examples/plot_tuh_eeg_corpus.py#L214-L218). If `preprocess` now also handles the serialization, how would I do this prior to saving?

That's a good point @gemeinl ... First thing, I wonder whether this kind of information (sampling frequency, reference, etc.) should actually be in `description`? We could add properties to `Base/WindowsDataset` to retrieve this information from `self.raw/windows.info` instead, which might be more flexible and less error-prone than relying on "fixed" info collected when loading the data. (We might need some logic on the side of `BaseConcatDataset` to return these properties when they are not the same across all datasets, e.g. different files have different references.)

If there's no way around it,  we could maybe add a `updated_description` argument (e.g. a dict) to `preprocess` to update each description before serialization?

gemeinl:
> That's a good point @gemeinl ... First thing, I wonder whether this kind of information (sampling frequency, reference, etc.) should actually be in `description`? We could add properties to `Base/WindowsDataset` to retrieve this information from `self.raw/windows.info` instead, which might be more flexible and less error-prone than relying on "fixed" info collected when loading the data. (We might need some logic on the side of `BaseConcatDataset` to return these properties when they are not the same across all datasets, e.g. different files have different references.)

I guess you are right. It is mostly duplicated information from Raw/Epochs.info. I should probably not have added it to the description in the first place. I think I made some pre-selection based on these entries but I could always get it by iterating through the dataset and accessing info instead. I had a comment somewhere that the number of windows is required for loading later on. I am not sure whether this still holds, but if so, this could be information that will be added by the windowers.

braindecode/preprocessing/preprocess.py
```diff
@@ -120,54 +127,109 @@ def preprocess(concat_ds, preprocessors):
...
+            # preprocessing happened in-place
+            # Recompute cumulative sizes as transforms might have changed them
+            concat_ds.cumulative_sizes = concat_ds.cumsum(concat_ds.datasets)
+        else:  # joblib made copies
+            _replace_inplace(concat_ds, BaseConcatDataset(list_of_ds))
+
+    # Store preprocessing keyword arguments in the dataset
     preproc_kwargs = _get_preproc_kwargs(preprocessors)
+    concat_kind = 'raw' if hasattr(concat_ds.datasets[0], 'raw') else 'window'
+    # XXX The kwargs are not currently saved along the datasets when `save_dir`
```
gemeinl, line 176: Why not? `BaseConcatDataset.save()` should take care of it, see https://github.com/braindecode/braindecode/blob/master/braindecode/datasets/base.py#L422-L427.

braindecode/preprocessing/preprocess.py
```diff
@@ -120,54 +127,109 @@ def preprocess(concat_ds, preprocessors):
...
+        _preprocess_raw_or_epochs(ds.windows, preprocessors)
+    else:
+        raise ValueError(
+            'Can only preprocess concatenation of BaseDataset or '
+            'WindowsDataset, with either a `raw` or `windows` attribute.')
+
+    if save_dir is not None:
+        concat_ds = BaseConcatDataset([ds])
+        save_subdir = os.path.join(save_dir, str(ds_index))
+        # XXX Save target_name.json if ds is a BaseDataset
```
gemeinl, line 225: Should be done automatically when calling `BaseConcatDataset.save()`, see https://github.com/braindecode/braindecode/blob/master/braindecode/datasets/base.py#L419-L420.

gemeinl:
@hubertjb What is your preferred order of merging this PR 277 and PR 288?

braindecode/preprocessing/preprocess.py
hubertjb: It's because I hadn't implemented it yet and I didn't want to forget :P The latest commit should cover that.

hubertjb:
> @hubertjb What is your preferred order of merging this PR 277 and PR 288?

If tests pass here we could merge this first as it's pretty much ready, but other than that I don't have a preference.

braindecode/datasets/base.py
```diff
@@ -385,10 +384,8 @@ def save(self, path, overwrite=False):
         description_file_name = os.path.join(path, 'description.json')
         target_file_name = os.path.join(path, 'target_name.json')
         if not overwrite:
-            if (os.path.exists(description_file_name) or
-                    os.path.exists(target_file_name)):
-                raise FileExistsError(
-                    f'{description_file_name} or {target_file_name} exist in {path}.')
+            from braindecode.datautil.serialization import check_save_dir_empty
```
agramfort, line 387: you need this to avoid a circular import?

braindecode/datautil/serialization.py
```diff
@@ -105,14 +105,40 @@ def _load_signals_and_description(path, preload, raws, ids_to_load=None):
...
         signals = mne.io.read_raw_fif(fif_file, preload=preload)
-    else:
+    elif fif_file.endswith('-epo.fif'):
         signals = mne.read_epochs(fif_file, preload=preload)
+    else:
+        raise ValueError('fif_file must end with raw.fif or epo.fif.')
     return signals
+
+
+def check_save_dir_empty(save_dir):
```
agramfort, line 123: do we need to have this public?

examples/benchmark_preprocessing.py
```diff
@@ -0,0 +1,150 @@
...
+ax.set_xlabel('Time (s)')
+ax.set_ylabel('Memory usage (MiB)')
+ax.set_title('Loading and preprocessing 8 recordings from Sleep Physionet')
+plt.show()
+
+
+###############################################################################
+# We see that parallel preprocessing without serialization (blue crosses) is
+# faster than simple sequential processing (blue circles), however it uses
+# significantly more memory.
```
agramfort, line 145: will this be clear on the doc built on CI?

braindecode/datasets/base.py
hubertjb: Yes, that's right.

braindecode/datautil/serialization.py
hubertjb: I guess not, I'll make it private.

examples/benchmark_preprocessing.py
hubertjb: This raises a good point - this example is not currently run by the CI. I could adapt it so the computation time is shorter (it took around 20 mins on my machine if I remember well), but also I'm not sure whether the CI machines are configured to run with multiple cores?

braindecode/datautil/serialization.py
robintibor: So I wait for you to make it private and then merge, unless there are objections @gemeinl ?

hubertjb: @robintibor the latest commit fixes this!

examples/benchmark_preprocessing.py
hubertjb: @robintibor I'm curious to know what you think - should we run this kind of example in the CI?

agramfort: I have mixed feeling here then. I think it's very valuable information but I am wondering how it will be perceived and discovered by users. We could keep it but if not run on CIs is likely to break silently. I would suggest to amend / update existing examples with note and warning sections to explain how the present example can make use of parallel and serialization should the data not fit in RAM.

braindecode/datautil/serialization.py
```diff
@@ -105,14 +105,40 @@ def _load_signals_and_description(path, preload, raws, ids_to_load=None):
...
+    Parameters
+    ----------
+    save_dir : str
+        Directory under which ``concat_ds`` will be saved.
+
+    Raises
+    -------
+    FileExistsError
+        If ``save_dir`` is not a valid directory for saving, i.e., the
+        description.json and/or target_name.json files already exist.
```
gemeinl, line 135: Our logic changed a bit here. No file is allowed to exist in `save_dir`. Subdirectories are fine.

braindecode/datautil/serialization.py
```diff
@@ -105,14 +105,40 @@ def _load_signals_and_description(path, preload, raws, ids_to_load=None):
...
-    else:
+    elif fif_file.endswith('-epo.fif'):
         signals = mne.read_epochs(fif_file, preload=preload)
+    else:
+        raise ValueError('fif_file must end with raw.fif or epo.fif.')
     return signals
+
+
+def _check_save_dir_empty(save_dir):
+    """Make sure a BaseConcatDataset can be saved under a given directory.
```
gemeinl, line 124: There is no dataset in the scope of this function.

braindecode/datautil/serialization.py
```diff
@@ -105,14 +105,40 @@ def _load_signals_and_description(path, preload, raws, ids_to_load=None):
...
+
+    Parameters
+    ----------
+    save_dir : str
+        Directory under which ``concat_ds`` will be saved.
```
gemeinl, line 129: See above

braindecode/preprocessing/preprocess.py
```diff
@@ -120,54 +127,123 @@ def preprocess(concat_ds, preprocessors):
         A concat of BaseDataset or WindowsDataset datasets to be preprocessed.
     preprocessors: list(Preprocessor)
         List of Preprocessor objects to apply to the dataset.
+    save_dir : str | None
+        If a string, the preprocessed data will be saved under the specified
+        directory and the datasets in ``concat_ds`` will be reloaded with
+        `preload=False`.
+    overwrite : bool
+        When `save_dir` is provided, controls whether to overwrite existing
+        files in `save_dir`.
```
gemeinl, line 136: Overwriting logic also changed. We will delete and create new subdirectories if overwrite = True

braindecode/preprocessing/preprocess.py
gemeinl: Overwriting will be on subdirectory level

braindecode/preprocessing/preprocess.py
```diff
@@ -120,54 +127,123 @@ def preprocess(concat_ds, preprocessors):
...
+
+    if save_dir is not None:
+        concat_ds = BaseConcatDataset([ds])
+
+        # Store preprocessing keyword arguments in the dataset
+        _set_preproc_kwargs(concat_ds, preprocessors)
+
+        save_subdir = os.path.join(save_dir, str(ds_index))
+        if not os.path.exists(save_subdir):
+            os.makedirs(save_subdir)
```
gemeinl, line 241: Creation of subdirectories will be handled by BaseConcatDataset.save()

braindecode/preprocessing/preprocess.py
```diff
@@ -120,54 +127,123 @@ def preprocess(concat_ds, preprocessors):
...
+    else:
+        raise ValueError(
+            'Can only preprocess concatenation of BaseDataset or '
+            'WindowsDataset, with either a `raw` or `windows` attribute.')
+
+    if save_dir is not None:
+        concat_ds = BaseConcatDataset([ds])
+
+        # Store preprocessing keyword arguments in the dataset
+        _set_preproc_kwargs(concat_ds, preprocessors)
```
gemeinl, line 237: If this is a BaseConcatDataset of WindowsDatasets, we would also have to set window_kwargs

gemeinl:
I am not so sure how to handle some things here now. They have to be changed, however, if changed they will only work as soon as PR288 is also merged.

examples/benchmark_preprocessing.py
robintibor: I am also not sure, in general alex suggestion makes sense to me but I am not sure if it makes sense to to this in all the existing examples, maybe just in the first one. For me, you could tryy to see if it can be made very fast and if it runs and produces examples on CI as expected? And if not, would be ok for me that it does not run on CI as well.

hubertjb: Good news, I modified the example so that it runs on 2 files in parallel, and it yields the same conclusions as the previous version of the example on 8 files. Given that Github Actions runs on [2-core Azure machines](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), we should be able to get similar results and include the example in the docs!

hubertjb:
> I am not so sure how to handle some things here now. They have to be changed, however, if changed they will only work as soon as PR288 is also merged.

I think we could merge this PR first since it should be ready (with the updated example). I'm happy to then help fix the parts that won't work anymore once we rebase #288 on this.

hubertjb:
Not sure what's going on, it looks like the acceptance tests are now failing. Any idea what might cause this @robintibor? 

From what I saw in the previous CI runs I believe there might still be one problem with the new example as well. I'm not sure why this happens as there is a unit test that currently passes and that does pretty much the same thing... I will have to investigate further.

@gemeinl with this in mind if #288 is ready to be merged I don't mind merging this one after!

robintibor:
So yeah ignore unit test for now maybe, seems just numbers can vary quite a bit randomly unfortunately. However, now seems doc building failed @hubertjb can you have a look there?

gemeinl:
I also restarted this job, now `examples/plot_benchmark_preprocessing.py` fails ...

robintibor:
Rebasing now should also prevent the test from failing  in the future as I have increased tolerance further

robintibor:
#288 is now merged so a lot of conflicts to solve :)

hubertjb:
I had some fun fixing the conflicts :P 

Everything works locally, however I keep getting an error on the CI for the new example when preprocessing the windowed data with `save_dir` not None: https://github.com/braindecode/braindecode/blob/54a417b856963c6a89236d61969f07ee15344283/examples/plot_benchmark_preprocessing.py#L102

The failure comes from mne's `_check_consistency` function: https://github.com/mne-tools/mne-python/blob/401ac32e60d56501cb2c8372d115c09ae97d4e64/mne/epochs.py#L564
Apparently the `times` attribute becomes writeable when `save_dir` is not None, and so this function throws an error. The weird thing is that this doesn't happen locally, with the same mne and numpy versions.

The following fixed the issue, but I don't think it's a great solution:
https://github.com/braindecode/braindecode/blob/54a417b856963c6a89236d61969f07ee15344283/braindecode/datasets/base.py#L563

@agramfort do you know what might be happening here?

gemeinl:
Could you please also update the TUH example according to the new preprocessing and serialization logic?
https://github.com/braindecode/braindecode/blob/master/examples/plot_tuh_eeg_corpus.py#L212-L223
I think it fits the scope of this PR.

robintibor:
> I had some fun fixing the conflicts :P
> 
> Everything works locally, however I keep getting an error on the CI for the new example when preprocessing the windowed data with `save_dir` not None:
> 
> https://github.com/braindecode/braindecode/blob/54a417b856963c6a89236d61969f07ee15344283/examples/plot_benchmark_preprocessing.py#L102
> 
> The failure comes from mne's `_check_consistency` function: https://github.com/mne-tools/mne-python/blob/401ac32e60d56501cb2c8372d115c09ae97d4e64/mne/epochs.py#L564
> Apparently the `times` attribute becomes writeable when `save_dir` is not None, and so this function throws an error. The weird thing is that this doesn't happen locally, with the same mne and numpy versions.
> 
> The following fixed the issue, but I don't think it's a great solution:
> 
> https://github.com/braindecode/braindecode/blob/54a417b856963c6a89236d61969f07ee15344283/braindecode/datasets/base.py#L563
> 
> @agramfort do you know what might be happening here?

Of course would be good to have a try to find out what is going on, if it is not resolvable, I could live with that line that fixes it though (with a bit more explanation in the comment before).

hubertjb:
I've updated the TUH example and made the CI failure comment clearer. I could open an issue to remind us to investigate this in more detail at some point, so we can merge this PR?

braindecode/preprocessing/preprocess.py
robintibor: Is the doc string here now consistent with what happens after merging #288 @hubertjb ?

braindecode/preprocessing/preprocess.py
```diff
@@ -120,54 +126,114 @@ def preprocess(concat_ds, preprocessors):
...
+            # Recompute cumulative sizes as transforms might have changed them
+            concat_ds.cumulative_sizes = concat_ds.cumsum(concat_ds.datasets)
+        else:  # joblib made copies
+            _replace_inplace(concat_ds, BaseConcatDataset(list_of_ds))
+
+    return concat_ds
+
+
+def _replace_inplace(concat_ds, new_concat_ds):
+    """Inplace modification of a BaseConcatDataset with another's attributes.
```
robintibor, line 176: Could you write more here, like all "subdatasets" will be replaced and also the preproc_kwargs

hubertjb:
@robintibor should we do a "squash and merge"? There are a lot of small commits with uninformative messages...

robintibor:
ok will  push button once CIs finished! :D

robintibor:
Oh! Almost forgot to click the button! Another buttonclick another big PR merged, amazing.

agramfort:
🍻

PR #278, opened 2021-06-25
some fixes

agramfort:


braindecode/models/deep4.py
```diff
@@ -13,12 +13,17 @@
...
-    Deep ConvNet model from [1]_.
+    """Deep ConvNet model from Schirrmeister et al 2017.
 
-    References
+    Model described in [2]_.
+
+    Parameters
     ----------
+    in_chans : int
+        XXX
```
agramfort, line 23: @robintibor you see that many models have no docstrings for parameters. Can you assign someone to fix this?

PR #279, opened 2021-06-25
FIX class documentation for inherited methods

tomMoral:
This PR intends to avoid documenting inherited methods and changes the way the classes are described (see the template).

Questions are:

- do we want to modify this template, in particular the way attributes are documented.
- Are we sure we do not want to document any inherited methods? (it seems doable to only filter out the methods from `torch.nn.Module` for instance if necessary).

### DIFF
--- a/braindecode/models/deep4.py
+++ b/braindecode/models/deep4.py
@@ -14,16 +14,17 @@ from ..util import np_to_th
 
 class Deep4Net(nn.Sequential):
     """
-    Deep ConvNet model from [1]_.
+    Deep ConvNet model from [Schirrmeister2017]_.
 
     References
     ----------
-
-    .. [1] Schirrmeister, R. T., Springenberg, J. T., Fiederer, L. D. J.,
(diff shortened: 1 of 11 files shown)


agramfort:
looks great https://1184-232335424-gh.circle-artifacts.com/0/dev/index.html !

agramfort:
@robintibor can you merge this?

tomMoral:
Note that this removes the list of methods at the beginning:

image
image


Also, it  seems that the attributes are documented in the docstring so should I remove them?

tomMoral:
For the before/after on he models:

- before: https://braindecode.org/generated/braindecode.models.EEGNetv1.html
- after: https://1184-232335424-gh.circle-artifacts.com/0/dev/generated/braindecode.models.EEGNetv1.html#braindecode.models.EEGNetv1

agramfort:
I love it as it is !

>

robintibor:
Please indicate if ready to merge @tomMoral

agramfort:
beautiful 

https://1212-232335424-gh.circle-artifacts.com/0/dev/generated/braindecode.models.EEGNetv4.html

thx @tomMoral !

PR #280, opened 2021-06-25
some nitpicks on SequenceSampler PR

agramfort:
cc @robintibor @hubertjb  what do you think?

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -217,18 +217,16 @@ class BaseConcatDataset(ConcatDataset):
     ----------
     list_of_ds : list
         list of BaseDataset, BaseConcatDataset or WindowsDataset
-    seq_target_transform : callable | None
-        Function to call on sequences of targets before returning them. Only
-        called if the dataset is indexed with lists of indices to return
-        sequences.
+    target_transform : callable | None
+        Optional function to call on targets before returning them.
     """
-    def __init__(self, list_of_ds, seq_target_transform=None):
+    def __init__(self, list_of_ds, target_transform=None):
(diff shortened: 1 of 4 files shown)


braindecode/datasets/base.py
```diff
@@ -217,18 +217,16 @@ class BaseConcatDataset(ConcatDataset):
...
         list of BaseDataset, BaseConcatDataset or WindowsDataset
-    seq_target_transform : callable | None
-        Function to call on sequences of targets before returning them. Only
-        called if the dataset is indexed with lists of indices to return
-        sequences.
+    target_transform : callable | None
+        Optional function to call on targets before returning them.
     """
-    def __init__(self, list_of_ds, seq_target_transform=None):
+    def __init__(self, list_of_ds, target_transform=None):
```
robintibor, line 223: We had some longer discussion about this already, that we may want to allow different transforms in the future for single items or sequences. Or at least that changing to one transform that is always applied may break unexpectedly, like if you have a transform that deals with sequences of labels and then maybe this works:
```
dataset[0:3]
```
whereas this will throw some error because the supplied transform won't handle single items well:
```
dataset[0]
```

Therefore we first keep like this make explicit this is transformation applied to sequences and this allows user still to retrieve original single-item labels untransformed

robintibor:
So how about this solution:
* target_transform as part of BaseConcatDataset but always transforms targets per item, independent of if we get sequences or individual items
* What was seq_target_transform we could implement the way data augmentations are implemented, once https://github.com/braindecode/braindecode/pull/254 is merged? I think then everything could resolve very nicely?

robintibor:
Also @hubertjb @cedricrommel

PR #282, opened 2021-06-25
[MRG] USLEEP model

Tgnassou:
Add USleep model for sleep staging
Joint work with @l-omar-chehab


braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+#          Omar Chehab <l-emir-omar.chehab@inria.fr>
+#
+# License: BSD (3-clause)
+
+import numpy as np
+import torch
+from torch import nn
+
+
+def crop_tensors_to_match(x1, x2, axis=-1):
```
agramfort, line 11:
```suggestion
def _crop_tensors_to_match(x1, x2, axis=-1):
```
make this private

braindecode/datasets/base.py
```diff
@@ -227,7 +227,6 @@ def __init__(self, list_of_ds, seq_target_transform=None):
         if list_of_ds and isinstance(list_of_ds[0], BaseConcatDataset):
             list_of_ds = [d for ds in list_of_ds for d in ds.datasets]
         super().__init__(list_of_ds)
```
agramfort, line 229:
```suggestion
        super().__init__(list_of_ds)

```

braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+        index=torch.arange(dim_cropped).to(device=x1.device)
+    )
+    x2_cropped = torch.index_select(
+        x2, dim=axis,
+        index=torch.arange(dim_cropped).to(device=x1.device)
+    )
+    return x1_cropped, x2_cropped
+
+
+class EncoderBlock(nn.Module):
```
agramfort, line 26:
```suggestion
class _EncoderBlock(nn.Module):
```

make it private. Just keep USleep public

braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+    '''Encoding block for a timeseries x of shape (B, C, T).'''
+    def __init__(self,
+                 in_channels=2,
```
agramfort, line 29:
```suggestion
                 in_chans=2,
```

@robintibor @hubertjb we need some consistency here

braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+    def forward(self, x):
+        x = self.block_prepool(x)
+        residual = x
+        if x.shape[-1] % 2:
+            x = nn.ConstantPad1d(padding=1, value=0)(x)
+        x = nn.MaxPool1d(kernel_size=self.downsample, stride=self.downsample)(x)
+        return x, residual
+
+
+class DecoderBlock(nn.Module):
```
agramfort, line 58:
```suggestion
class _DecoderBlock(nn.Module):
```

braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+            x = torch.cat([x, residual], axis=1)  # (B, 2 * C, T)
+        x = self.block_postskip(x)
+        return x
+
+
+class USleep(nn.Module):
+    """Sleep staging architecture from [1]_.
+
+    U-Net (autoencoder with skip connections) feature-extractor for sleep staging described in [1]_.
+
```
agramfort, line 106:
```suggestion
    """Sleep staging architecture from Perslev et al 2021.

    U-Net (autoencoder with skip connections) feature-extractor for sleep staging described in [1]_.

```

braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+        -- the temporal dimension shrinks (via maxpooling in the time-domain)
+        -- the spatial dimension expands (via more conv1d filters in the time-domain)
+    For the decoder ('up'):
+        -- the temporal dimension expands (via upsampling in the time-domain)
+        -- the spatial dimension shrinks (via fewer conv1d filters in the time-domain)
+    Both do so at exponential rates.
+
+    Parameters
+    ----------
+    n_channels : int
```
agramfort, line 117: we need to fix this naming

braindecode/models/usleep.py
```diff
@@ -0,0 +1,281 @@
...
+                             upsample=max_pool_size,
+                             with_skip_connection=with_skip_connection)
+            ]
+        self.decoder = nn.Sequential(*decoder)
+
+        # Instantiate classifier
+        # self.clf = nn.Sequential(
+        #     nn.Dropout(0.5),
+        #     nn.Linear(channels[1] * input_size, n_classes)
+        # )
```
agramfort, line 218: cleanup

docs/whats_new.rst
```diff
@@ -55,7 +55,8 @@ Enhancements
 - Removing code duplicate in BaseDataset and WindowsDataset (:gh:`159` by `Lukas Gemein`_)
 - Only load data if needed during preprocessing (e.g., allow timecrop without loading) (:gh:`164` by `Robin Tibor Schirrmeister`_)
 - Adding option to sort filtered channels by frequency band for the filterbank in :func:`braindecode.datautil.filterbank` (:gh:`185` by `Lukas Gemein`_)
-
+- Adding the USleep model :class:`braindecode.models.USleep` (:gh:`` by `Theo Gnassounou`_ and `Omar Chehab`_)
```
agramfort, line 58:
```suggestion
- Adding the USleep model :class:`braindecode.models.USleep` (:gh:`282` by `Theo Gnassounou`_ and `Omar Chehab`_)
```

examples/plot_sleep_staging.py
```diff
@@ -96,22 +96,27 @@
...
-    'Sleep stage 2': 2,
-    'Sleep stage 3': 3,
-    'Sleep stage 4': 3,
-    'Sleep stage R': 4
+    "Sleep stage W": 0,
+    "Sleep stage 1": 1,
+    "Sleep stage 2": 2,
+    "Sleep stage 3": 3,
+    "Sleep stage 4": 3,
+    "Sleep stage R": 4,
```
agramfort, line 104: you should revert changes in this file.

examples/plot_sleep_staging_sequences.py
```diff
@@ -0,0 +1,429 @@
+"""
+Sleep staging on the Sleep Physionet dataset
+============================================
```
agramfort, line 3: this file should go

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,410 @@
+"""
+Sleep staging on the Sleep Physionet dataset
+============================================
+
+This tutorial shows how to train and test a sleep staging neural network with
+Braindecode. We adapt the time distributed approach of [1]_ to learn on
+sequences of EEG windows using the openly accessible Sleep Physionet dataset
+[1]_ [2]_.
```
agramfort, line 8: you should add ref to usleep paper not Chambon

examples/plot_sleep_staging.py
l-omar-chehab: I'm not sure I understand what this means

agramfort: you should not have touched this file in this PR.

l-omar-chehab: absolutely - I just made sure this script is up-to-date with braindecode master so we don't inadvertently change it.

braindecode/models/usleep.py
l-omar-chehab: Yes. We just changed it to in_chans for consistency with (some of the other) braindecode models. We will open a more general PR for naming.

agramfort: @l-omar-chehab in_chans now

braindecode/models/usleep.py
```diff
@@ -0,0 +1,274 @@
...
+        Number of EEG or EOG channels. Set to 2 in [1]_ (1 EEG, 1 EOG).
+    sfreq : float
+        EEG sampling frequency. Set to 128 in [1]_.
+    depth : int
+        Number of encoding (resp. decoding) blocks in the U-Net.
+        Set to 12 in [1]_ with sfreq=128.
+        Here we set it to 10 with sfreq=100.
+    time_conv_size_s : float
+        Size of filters in temporal convolution layers, in seconds.
+        Set to 0.070 in [1]_ (9 samples at sfreq=128).
```
agramfort, line 127:
```suggestion
        Set to 0.070 s in [1]_ (9 samples at sfreq=128).
```

braindecode/models/usleep.py
```diff
@@ -0,0 +1,274 @@
...
+    complexity_factor : float
+        Multiplicative factor for number of channels at each layer of the U-Net.
+        Set to sqrt(2) in [1]_.
+    input_size_s : float
+        Size of the input, in seconds. Set to 30.
+    n_classes : int
+        Number of classes. Set to 5.
+    apply_batch_norm : bool
+        If True, apply batch normalization after temporal convolutional
+        layers.
```
agramfort, line 143:
```suggestion
        If True, apply batch normalization after temporal convolutional layers.
```

braindecode/models/usleep.py
```diff
@@ -0,0 +1,274 @@
...
+
+    References
+    ----------
+    .. [1] Perslev, M., Darkner, S., Kempfner, L. et al.
```
agramfort, line 147: put full author list

examples/plot_sleep_staging.py
```diff
@@ -1,24 +1,20 @@
 """
 Sleep staging on the Sleep Physionet dataset
 ============================================
-
```
agramfort, line 4: just copy the file 

https://raw.githubusercontent.com/braindecode/braindecode/master/examples/plot_sleep_staging.py

to avoid the diff here

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,369 @@
...
+=================================================================
+
+This tutorial shows how to train and test a sleep staging neural network with
+Braindecode. We adapt the time distributed approach of [1]_ to learn on
+sequences of EEG windows using the openly accessible Sleep Physionet dataset
+[1]_ [2]_.
+
+References
+----------
+.. [1] Perslev, M., Darkner, S., Kempfner, L. et al.
```
agramfort, line 12: put full author list

braindecode/models/usleep.py
l-omar-chehab: Ok. I changed n_channels to in_chans in the final model (USleep) to homogenize with other braindecode models. 

For the intermediate classes (_EncoderBlock, _DecoderBlock), maybe for now we can leave in_channels and out_channels as is, given they are the names of the attributes in pytorch? And we could address the more general naming issue in a separate PR?

agramfort:
>
> Ok. I changed n_channels to in_chans in the final model (USleep) to
> homogenize with other braindecode models.
>
> For the intermediate classes (_EncoderBlock, _DecoderBlock), maybe for now
> we can leave in_channels and out_channels as is, given they are the names
> of the attributes in pytorch? And we could address the more general naming
> issue in a separate PR?
>

OK

examples/plot_sleep_staging.py
l-omar-chehab: ok done

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+######################################################################
+# Window preprocessing
+# ~~~~~~~~~~~~~~~~~~~
+#
+# We also preprocess the windows by applying channel-wise z-score normalization
+# in each window.
+
+from braindecode.preprocessing.preprocess import zscore
```
agramfort, line 118:
```suggestion
from braindecode.preprocessing import zscore
```

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,369 @@
...
+       a sleep-dependent neuronal feedback loop: the slow-wave
+       microcontinuity of the EEG. IEEE-BME 47(9):1185-1194 (2000).
+
+.. [3] Goldberger AL, Amaral LAN, Glass L, Hausdorff JM, Ivanov PCh,
+       Mark RG, Mietus JE, Moody GB, Peng C-K, Stanley HE. (2000)
+       PhysioBank, PhysioToolkit, and PhysioNet: Components of a New
+       Research Resource for Complex Physiologic Signals.
+       Circulation 101(23):e215-e220
+"""
+# Authors: Hubert Banville <hubert.jbanville@gmail.com>
```
agramfort, line 26: fix authors

agramfort:
also you need to merge master to fix conflicts in what's new page

agramfort:
looks good https://1291-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_sleep_staging_usleep.html !

+1 for MRG on my side

agramfort (approved):
@hubertjb @gemeinl @robintibor merge if happy

hubertjbanville:
> @hubertjb @gemeinl @robintibor merge if happy

Taking a look now!

braindecode/models/usleep.py
```diff
@@ -0,0 +1,272 @@
...
+                nn.ELU(),
+                nn.BatchNorm1d(num_features=out_channels),
+            )
+
+    def forward(self, x):
+        x = self.block_prepool(x)
+        residual = x
+        if x.shape[-1] % 2:
+            x = nn.ConstantPad1d(padding=1, value=0)(x)
+        x = nn.MaxPool1d(kernel_size=self.downsample, stride=self.downsample)(x)
```
robintibor, line 54: Why not just instantiate these during object construction like
```python
self.pad = nn.ConstantPad1d(padding=1, value=0)
self.pool = nn.MaxPool1d(kernel_size=self.downsample, stride=self.downsample)
```

Seems strange to me to create object to then just use it, then one could also call torch functional API?

robintibor:
I left one comment regarding classes.
Another thing, can we further reduce example doc building time by using less data or fewer epochs? Right now it definitely increases doc build time quite a bit more than all other examples:
image

hubertjb:
I made a few changes, mainly uniformizing the two sleep staging examples. I've changed the preprocessing to fit with the U-Sleep paper, and also tried to limit running time - we'll see if that has the expected effect on the CI.

braindecode/preprocessing/preprocess.py
```diff
@@ -274,23 +275,53 @@ def zscore(data):
     return zscored
 
 
+def robust_scale(data):
```
agramfort, line 278: @hubertjbanville I don't get why we have this robust_scale function and changes in plot_sleep_staging.py in this PR. It's not already in master? it's required here for usleep training?

hubertjb: There wasn't a robust scaling implementation yet in braindecode. It's required here as it's the only preprocessing step used in the U-Sleep paper. We could use a simple zscore but it was simple enough to implement so I added it in.

The changes to `plot_sleep_staging.py` are just because so the two sleep staging examples are more uniform.

agramfort:
@hubertjbanville we will squash and merge this one. We'll have robust scaler and Usleep in one commit. For me it's a bad example for the younger folks here :) can you make a dedicated PR to show good practices? 🙏

hubertjb:
> @hubertjbanville we will squash and merge this one. We'll have robust scaler and Usleep in one commit. For me it's a bad example for the younger folks here :) can you make a dedicated PR to show good practices? pray

Aah right if we squash that would be a mess. Please see #291 ! :)

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,360 @@
...
+#
+# Next, we preprocess the raw data. We scale each channel recording-wise to
+# have zero median and unit interquartile range. We don't upsample to 128 Hz as
+# done in [1]_ so that we keep the example as light as possible. No filtering
+# is described in [1]_.
+
+from braindecode.preprocessing import preprocess, Preprocessor, zscore
+
+# XXX To be changed with robust_scale
+preprocessors = [Preprocessor(zscore, channel_wise=True)]
```
robintibor, line 71: so here could be possible to just use scale from sklearn as per https://github.com/braindecode/braindecode/pull/291#issuecomment-872072222 ? or would that be different?

robintibor: but could also handle this in separate PR I guess? Then I would merge? Unless you want to change this here already

gemeinl: As the comment above suggests, it should be `sklearn.preprocessing.robust_scale`. But yes, it should be possible to use this here instead.

robintibor: `sklearn.preprocessing.robust_scale` or `sklearn.preprocessing.scale`?

agramfort:
don't merge yet

>

robintibor:
ok then please reindicate in title when ready to merge

hubertjb:
I also removed the window-wise zscoring step, which isn't in the USleep paper.

Re-reading the paper there's actually one more preprocessing step that is not currently in the example:
> We then clip any value which has an absolute deviation from the median of more than 20 times the IQR of that channel.
Not sure this is necessary for this example but we could add it too.

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,348 @@
+"""
+Sleep staging on the Sleep Physionet dataset using U-Sleep network
+==================================================================
+
+.. note::
+    Please take a look at the simpler sleep staging example
+    `here <https://braindecode.org/auto_examples/plot_sleep_staging.html>`_
```
agramfort, line 7: do not do it like this. All sphinx gallery examples have an anchor you can use for a link.
Here this would point to stable release although it should point to dev version.

Anchor is called sphx_glr_auto_examples_plot_sleep_staging.py so I think it could be:
```suggestion
    :ref:`here <sphx_glr_auto_examples_plot_sleep_staging.py>`
```

robintibor:
> I left one comment regarding object creation.
> Another thing, can we further reduce example doc building time by using less data or fewer epochs? Right now it definitely increases doc build time quite a bit more than all other examples:
> image

ah and please also consider about this

braindecode/models/usleep.py
```diff
@@ -0,0 +1,272 @@
...
+                 n_classes=5,
+                 input_size_s=30,
+                 apply_softmax=False
+                 ):
+        super().__init__()
+
+        self.in_chans = in_chans
+
+        # Harcoded (otherwise dims can break)
+        depth = 10
```
hubertjb, line 161: @l-omar-chehab Is this the number of blocks? Isn't it 12 in the paper?

braindecode/models/usleep.py
```diff
@@ -0,0 +1,272 @@
...
+        time_conv_size = 9  # 0.09s at sfreq = 100 Hz
+        max_pool_size = 2   # 0.02s at sfreq = 100 Hz
+        n_time_filters = 5
```
hubertjb, line 164: Could this be a parameter? I'm looking for ways to decrease the complexity of the model so the example runs faster...

braindecode/models/usleep.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # Convert between units: seconds to time-points (at sfreq)
+        input_size = np.ceil(input_size_s * sfreq).astype(int)
+
+        # Define geometric sequence of channels
+        channels = (
+            n_time_filters * complexity_factor * np.sqrt(2) ** np.arange(0, depth + 1)
```
hubertjb, line 171: Should there be both `complexity_factor` and `np.sqrt(2)` here?

braindecode/models/usleep.py
Tgnassou: It's 12 for Usleep because the frequency is 128Hz. With 100Hz it is only possible to take 10 depth because we can divide tensor of size 1

braindecode/models/usleep.py
Tgnassou: We can reduce the depth yes, if it is your question

braindecode/models/usleep.py
hubertjb: Ah got it! Maybe this should be explained in the docstring in case someone wants to use the architecture with sfreq=128?

braindecode/models/usleep.py
Tgnassou: Yes, we just make the same things as USleep

braindecode/models/usleep.py
hubertjb: OK good, in this case I think it should be added to the arguments!

braindecode/models/usleep.py
hubertjb: I'm confused by the description in the paper:
image

Each layer has srqt(2) times more convolutional channels than the previous one. For the encoder at least, there's only one conv layer per block. How can the number of degrees of freedom **double** from one block to the next?

Tgnassou: Why do you want the degree of freedom double? it is just multiply by sqrt(2)
Maybe i don't understand your point

hubertjb: The doubling is just what they say in the paper. It doesn't seem to fit with their equation so that's what I was wondering about. You've looked at their keras implementation - maybe it's clearer in there?

With the current code though, you end up with:
```
np.ceil(5 * np.sqrt(2) * np.sqrt(2) ** np.arange(0, 10 + 1))
Out[5]: array([  8.,  11.,  15.,  21.,  29.,  41.,  57.,  81., 114., 161., 227.])
```
That doesn't look right to me. I think it should be `np.floor` instead of `np.ceil`, and the hardcoded `np.sqrt(2)` should be removed since that's what `complexity_factor` does.

That would give: `np.floor(n_time_filters * complexity_factor ** np.arange(0, depth + 1))`
```
np.floor(5 * np.sqrt(2) ** np.arange(0, 10 + 1))
Out[6]: array([  5.,   7.,  10.,  14.,  20.,  28.,  40.,  56.,  80., 113., 160.])
```

Tgnassou: Ok, it's because Omar doesn't like the way USleep is code and he tries somethings different. So it is not exactly the code in keras, i don't know what is the best

agramfort: ????? !!!! you mean this architecture is not the published architecture ! and this is done silently as we say in the doc it's USleep???

robintibor:
btw, I would in general advocate to mostly try to reduce running time in examples by reducing number of datapoints and reducing number of epochs (drastically if needed, with clear note this is to reduce running time). This can be done for any example and makes things more uniform across examples and also seems easy to understand to me. Feel this is better than making model smaller which will mean something different for any architecture etc.

agramfort:
+1

>

Tgnassou:
@robintibor 2min46 for the USleep examples, is it ok for you ?

robintibor:
2 min 46 s is fine to me.

However, would prefer if:
* example clearly states whenever a choice is done to make stuff run faster **at that point**, not just at end of example
* as much as possible, as written before, just n_epochs is reduced and data size is reduced, e.g., could you keep model same and just reduce `crop_wake_mins=30` quite a lot?

hubertjb:
I just had a chat with @Tgnassou and @l-omar-chehab. There are three things that still need to be done in this PR:

1. Clarify what hyparameters were actually used in the USleep paper. The paper, the Supplementary Materials and the keras implementation all have different descriptions of how the numbers of convolutional channel are computed. A quick summary:

    * [Paper (p.8)](https://www.nature.com/articles/s41746-021-00440-5): c_t = floor(c_{t-1} * np.sqrt(2)); c_1 = 5
    * [Suppl. (p.32)](https://static-content.springer.com/esm/art%3A10.1038%2Fs41746-021-00440-5/MediaObjects/41746_2021_440_MOESM2_ESM.pdf): c_t = floor(c_{t-1} * np.sqrt(1.67)); c_1 = 6 (doesn't seem to fit with the values in the table)
    * [Keras](https://github.com/perslev/U-Time/blob/usleep-paper-version/utime/models/utime.py#L154) : c_t = floor(h_{t-1} * np.sqrt(2)) where h_t = floor(h_{t-1} * 2); h_1 = ?

@Tgnassou and @l-omar-chehab will reach out to the authors to ask them which version they ended up using.

2. Make sure the model can be used with more than just one fixed input sampling rate. Right now the kernel size and maxpool size are hardcoded to avoid previously seen problems with the dimensions of layers. It looks like the actual problem is that the maxpool size and the depth size (i.e. number of blocks) need to be compatible. @Tgnassou is looking at that.

3. We need another test that compares the dimensions of the model to those reported in the paper.

hubertjb:
Also, to answer your previous comment @robintibor : `crop_wake_mins` actually shouldn't be decreased below 30 (it's the typical value used in the literature). I think at this point the data is really as small as it can get, unless we want to only use a few examples from the two sleep recordings that are used in the example.

robintibor:
Maybe a clarification regarding examples as I see it @hubertjb 
Obviously the nicest thing would be to have the best-performing pipeline as the examples, so people can directly use the code and expect great results and at the same time it is validated to run by the CI. 
However, this would take too long on the CI, and in the end we decided that:
* all decoding examples will run in CI
* code should be as close as possible to good pipeline so users may not be confused and run suboptimal hyperparameters expecting good results
* to ensure fast runs on CI, we drastically reduce data and epochs, even if it makes the example nonsensical as it is and clearly note that these are not useful values, indicating what are the useful values

This way, we can make sure the code always runs, is fast enough on CI, and should prevent users from mistakenly adopting suboptimal pipelines. And it is acceptable if the example due to reduced data and reduced number of examples does not learn anything as is (e.g., our trialwise decoding example (https://braindecode.org/master/auto_examples/plot_bcic_iv_2a_moabb_trial.html)  is barely above chance on train and exactly at chance on valid at the end of training).

Otherwise, with examples that are "more realistic" but with suboptimal hyperparameters, we have seen users adopt suboptimal hyperparameters in their training and expect good results.

Tgnassou:
This PR is ready to merge

examples/plot_sleep_staging_usleep.py
```diff
@@ -209,6 +209,7 @@
     with_skip_connection=True,
     n_classes=n_classes,
     input_size_s=input_size_samples / sfreq,
+    apply_softmax=True
```
agramfort, line 212: you need to use nn.CrossEntropyLoss here as you use class-weights? if so can you add a comment
so it's clear why you need to change a default parameter in a first example?

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,355 @@
...
+    before going through this example. The current example uses a more complex
+    architecture and a sequence-to-sequence (seq2seq) approach.
+
+This tutorial shows how to train and test a sleep staging neural network with
+Braindecode. We adapt the U-Sleep approach of [1]_ to learn on sequences of EEG
+windows using the openly accessible Sleep Physionet dataset [2]_ [3]_.
+
+.. warning::
+    The example is written to have a very short excecution time.
+    This number of epochs is here too small and very few recordings are used.
```
agramfort, line 17:
```suggestion
    The number of epochs is here too small and very few recordings are used.
```

examples/plot_sleep_staging_usleep.py
```diff
@@ -0,0 +1,355 @@
...
+    To obtain competitive results you need to use more data and more epochs.
```
agramfort, line 18: Given what you observe can you be more specific? How many epochs will lead to
what accuracy (roughly). Something like : Taking for example at least 30 recordings
to training and doing 40 epochs should lead to a balanced accuracy close to 0.8.

agramfort: can you also put the same warning in plot_sleep_staging.py  ?

examples/plot_sleep_staging_usleep.py
robintibor: what about these questions @Tgnassou ? please do add some comment about the `depth=12` line before the model creation, so it shows up clearly in the example doc why we reduce here. I am also confused about `apply_softmax=True` and using `nn.CrossEntropyLoss` as doc of cross entropy loss states:
https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html

```
This criterion combines LogSoftmax and NLLLoss in one single class.

[...]

The input is expected to contain raw, unnormalized scores for each class.```

examples/plot_sleep_staging_usleep.py
robintibor: Yes and also add the changing of the depth parameter, in case that was for execution time @Tgnassou

robintibor:
See comments by me and alex, and also @Tgnassou I invited you to be maintainer, then you don't have to always wait for me to approve workflows. could you accept this invitation?

robintibor:
So I think the loss comment from me above https://github.com/braindecode/braindecode/pull/282#discussion_r672987582 definitely needs addressing. Is somebody still active on it or do we need to search somebody @hubertjb  @Tgnassou @agramfort

Tgnassou:
I am in holidays, that's why i'm not very active these days. But I will continue the work in september! Is it ok for you @robintibor ?

robintibor:
ok my plan was to have a release ~end of august, but maybe we can shift to midseptember anyways, then this one could be inside as well

hubertjb:
I'm still working on it and should have results soon @robintibor

hubertjb:
Yes, I'll share that asap! 

So far I can say that the balanced sequence sampler does seem to speed up learning a bit. The USleep model, especially in its depth=12 version, is massively overparametrized though, especially given a smallish dataset like Sleep Physionet. Reducing the depth and adding some weight decay helps, but the SleepStagerChambon2018 learns much more quickly and so far reaches higher test performance.

Also I did see a minor improvement in validation performance when combining softmax + CrossEntropyLoss, but so far not as much as what @Tgnassou reported.

More to come soon!

examples/plot_sleep_staging_usleep.py
robintibor: Did you see this comment here @hubertjb ?

hubertjb:
Trying to conclude on this @robintibor @agramfort @Tgnassou 

1. U-Sleep is much harder to train on the "small-ish" SleepPhysionet. I found some configurations that could somewhat learn more quickly (e.g. model depth=8, weight decay=5e-5, learning rate=1e-5), but they never go much above 50% test balanced accuracy. SleepStagerChambon2018 in contrast can easily reach >70% in under 100 epochs. Using really low learning rates (e.g. 1e-7 like in the paper) does yield very smooth learning curves, but we would have to train models for a lot more than the 100 epochs I tested on to get interesting performance.

2. The balanced sampler seems to help quite a lot. The models learn faster and reach slightly better test performance. Here are some learning curves obtained with SleepStagerChambon2018:

image

3. The self-ensembling also does help, with gains of 5-10% balanced accuracy in my limited experiments on (overfitting) U-Sleep.

Overall, I'm confident the model implementation is accurate. I went through the original keras implementation and made sure the dimensions and the number of parameters were exactly the same (there are now unit tests for these). The model does not learn as well on SleepPhysionet, but my guess is this is an issue with the scale of the data and the training budget rather than the model itself.

agramfort:
Great summary. I suggest to merge. Thanks @hubertjb for finishing this and @Tgnassou and @l-omar-chehab for the ground work

Tgnassou:
Very nice ! Thanks you @hubertjb !

agramfort:
@robintibor the green button is yours !

robintibor:
There were some conflicts now due to other PR merge, which I have resolved. 
More importantly @hubertjb , it seems the comment regarding applying softmax and using crossentropyloss has not been addressed I think (https://github.com/braindecode/braindecode/pull/282/#discussion_r672987582) ? Crossentropyloss will apply softmax again, so it is not correct what is happening there I think... Maybe even `apply_softmax=False` could already solve it? Or do I have some misunderstanding? :)

hubertjb:
Thanks @robintibor for fixing the conflicts and for pointing that out! I forgot to change that line - it should indeed be `apply_softmax=False`.

robintibor:
Great stuff, thanks @Tgnassou @l-omar-chehab @hubertjb , merged now

agramfort:
awesome team effort @Tgnassou @l-omar-chehab @hubertjb 🚀 🚀 🚀 🚀 👏 👏 👏 👏

Issue #173, opened 2020-09-30
Download entire MOABB dataset


gemeinl:
MOABB Dataset could accept ```subject_ids=None``` which will download data of all subjects.

agramfort:
+1

>

robintibor:
did you need this functionality @gemeinl ? or why the thought to add it? don't have a problem with it just wondered what is motivation for it

gemeinl:
Well, I just found it inconvenient to get started:
I am not too familiar with the BCIC IV 2a dataset but I wanted to do some decoding on all the subjects.
To download the data, I need to know how many subjects there are, so I either have to search the MOABB source code for the docstring with the dataset description or search for the corresponding publication. Of course, sooner or later one should read about the dataset one is using, but I think downloading could be done without requiring prior knowledge about the dataset.

Alternatively, instead of accepting ```subject_ids=None``` we could make the MOABB docstring available in Braindecode.

robintibor:
ok understood. yeah why not both add subject_ids=None and make the docstring available

robintibor:
This logic here may help to understand how to process docstrings of other functions:

https://github.com/braindecode/braindecode/blob/8f6f1406d85a0412f480662f07595ddeaca4d869/braindecode/util.py#L343-L355
https://github.com/braindecode/braindecode/blob/8f6f1406d85a0412f480662f07595ddeaca4d869/braindecode/regressor.py#L18-L47

gemeinl:
@robintibor I do not see away of how to do this for general class `MOABBDataset`. We can only provide a docstring to our `BNCI2014001` and `HGD`, right?

gemeinl:
Because the `MOABBDataset` docstring will be checked by the user prior to instantiation. But we will receive the name of the dataset the user actually wants to use just then.

gemeinl:
Maybe one could create some kind of an overview table for the docstring of `MOABBDataset` which lists all the datasets fetchable through MOABB, the number of subjects and the pradigm. Could be a lot of work, though..

PR #284, opened 2021-06-25
[MRG] Extending docstrings for HGD & BNCI2014001

dcwil:
See #173

### DIFF
--- a/braindecode/util.py
+++ b/braindecode/util.py
@@ -355,6 +355,12 @@ def update_estimator_docstring(base_class, docstring):
     return out_docstring
 
 
+def update_moabb_docstring(base_class, docstring):
+    base_doc = base_class.__doc__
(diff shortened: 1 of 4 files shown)


braindecode/util.py
```diff
@@ -355,6 +355,12 @@ def update_estimator_docstring(base_class, docstring):
     return out_docstring
 
 
+def update_moabb_docstring(base_class, docstring):
```
agramfort, line 358:
```suggestion
def _update_moabb_docstring(base_class, docstring):
```
please make this function private

robintibor:
And maybe could be great to add the `BNCI2014001` and the `HGD` Dataset to `datasets/__init__.py` as well as to  `docs/api.rst` . Then we can also see in built documentation what it looks like.

robintibor:
Amazing, looks great @dcwil !

PR #287, opened 2021-06-27
[MRG] Save preprocessing and windowing arguments

gemeinl:
-initiated work on saving preprocessing and windowing arguments to BasseConcatDataset

### DIFF
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -185,7 +185,7 @@ def _get_preproc_kwargs(preprocessors):
             if 'fun' in p.fn:
                 func_name = p.kwargs['fun'].func.__name__
                 func_kwargs = p.kwargs['fun'].keywords
-        preproc_kwargs.append({func_name: func_kwargs})
+        preproc_kwargs.append((func_name, func_kwargs))
     return preproc_kwargs
 
 
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -178,9 +178,10 @@ def _get_preproc_kwargs(preprocessors):
         func_kwargs = p.kwargs
         # in case of another function, fn is a function 'fun' partially
         # initialized with keywords
-        if 'fun' in p.fn:
-            func_name = p.kwargs['fun'].func.__name__
-            func_kwargs = p.kwargs['fun'].keywords
+        if isinstance(p.fn, dict):
+            if 'fun' in p.fn:
+                func_name = p.kwargs['fun'].func.__name__
+                func_kwargs = p.kwargs['fun'].keywords
         preproc_kwargs.append((func_name, func_kwargs))
     return preproc_kwargs
 
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -131,11 +131,21 @@ def preprocess(concat_ds, preprocessors):
         assert hasattr(elem, 'apply'), (
             'Preprocessor object needs an `apply` method.')
 
+    # save the preprocessing arguments to the dataset
+    preproc_args = {}
+    for p in preprocessors:
+        if 'fun' in p.__dict__:
+            preproc_args.update({p.__dict__['fun']: p.__dict__['kwargs']})
+        else:
+            preproc_args.update({p.__dict__['fn']: p.__dict__['kwargs']})
+
     for ds in concat_ds.datasets:
         if hasattr(ds, 'raw'):
             _preprocess(ds.raw, preprocessors)
+            setattr(concat_ds, 'raw_preproc_args', preproc_args)
         elif hasattr(ds, 'windows'):
             _preprocess(ds.windows, preprocessors)
+            setattr(concat_ds, 'window_preproc_args', preproc_args)
         else:
             raise ValueError(
                 'Can only preprocess concatenation of BaseDataset or '
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -42,7 +42,9 @@ class Preprocessor(object):
     kwargs:
         Keyword arguments to be forwarded to the MNE function.
     """
-    def __init__(self, fn, apply_on_array=True, **kwargs):
+    def __init__(self, fn, *, apply_on_array=True, **kwargs):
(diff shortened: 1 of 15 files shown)


gemeinl:
As of [PEP 3102](https://www.python.org/dev/peps/pep-3102/) I would like to force all keyword arguments for `Preprocessor`. This avoids accidental misusage of `apply_on_array`, as well as it enables grabbing preprocessing parameters in a unified way to store them (we can simply grab `kwargs`).
Additionally, I would like to raise an error when a lambda function is passed to the preprocessor, since their arguments cannot be accessed in that way.

gemeinl:
@hubertjb I just had to learn that we cannot use `numpy.multiply` with our current `Preprocessor`. The reason is that `numpy.multiply` requires two __positional__ arguments (it actually forces two positional arguments using `/`, see https://numpy.org/doc/stable/reference/generated/numpy.multiply.html) where our `Preprocessor` only supports keyword arguments. Maybe it is worth keeping `braindecode.preprocessing.scale`.

hubertjb:
> @hubertjb I just had to learn that we cannot use `numpy.multiply` with our current `Preprocessor`. The reason is that `numpy.multiply` requires two **positional** arguments (it actually forces two positional arguments using `/`, see https://numpy.org/doc/stable/reference/generated/numpy.multiply.html, https://www.python.org/dev/peps/pep-0570/) where our `Preprocessor` only supports keyword arguments (https://github.com/braindecode/braindecode/blob/master/braindecode/preprocessing/preprocess.py#L45). Maybe it is worth keeping `braindecode.preprocessing.scale`.

Interesting, I was not aware of pep570! And sorry about that, I should have been more careful. Let's maybe reinstate it but rename it `braindecode.preprocessing.multiply`, to avoid confusion with sklearn's `scale` and `robust_scale`?

braindecode/preprocessing/preprocess.py
```diff
@@ -42,7 +42,9 @@ class Preprocessor(object):
     kwargs:
         Keyword arguments to be forwarded to the MNE function.
     """
-    def __init__(self, fn, apply_on_array=True, **kwargs):
+    def __init__(self, fn, *, apply_on_array=True, **kwargs):
+        if hasattr(fn, '__name__') and fn.__name__ == '<lambda>':
+            raise ValueError('lambda function not supported.')
```
robintibor, line 47: Could we give warning instead that the preprocessing then will not be serializable? I am afraid we will make API hard to learn by more and more (hidden) restrictions.

gemeinl:
Are there any other ideas for the structure of the saved preprocessing choices?
Right now they are consistently (for raw preprocessing, windowing and window preprocessing) stored as:
`list(dict(str, dict(kwarg: value)))`.
So it would look like this:
```
[{'scale': {'factor': 1000000.0}}]
```
```
[{'create_windows_from_events': {'on_missing': 'error',
   'n_jobs': 1,
   'drop_bad_windows': True,
   'drop_last_window': False,
   'flat': None,
   'mapping': None,
   'picks': None,
   'preload': False,
   'reject': None,
   'trial_start_offset_samples': 0,
   'trial_stop_offset_samples': 0,
   'window_size_samples': 100,
   'window_stride_samples': 100}}]
```
```
[{'clip': {'mi': -800, 'ma': 800}},
 {'clip': {'a_min': -800, 'a_max': 800}},
 {'crop': {'tmin': 60, 'tmax': 180}},
 {'scale': {'factor': 1000000.0}},
 {'custom_crop': {'tmin': 30, 'tmax': 90, 'include_tmax': False}}]
```
Windowing kwargs are a special case with this structure. The list is not acually needed, since there will always be just one windowing function. I kept the list for consistency. Any suggestions for improvement? @robintibor @hubertjb @sliwy @agramfort ?

braindecode/preprocessing/preprocess.py
gemeinl: > > @hubertjb I just had to learn that we cannot use `numpy.multiply` with our current `Preprocessor`. The reason is that `numpy.multiply` requires two **positional** arguments (it actually forces two positional arguments using `/`, see https://numpy.org/doc/stable/reference/generated/numpy.multiply.html, https://www.python.org/dev/peps/pep-0570/) where our `Preprocessor` only supports keyword arguments (https://github.com/braindecode/braindecode/blob/master/braindecode/preprocessing/preprocess.py#L45). Maybe it is worth keeping `braindecode.preprocessing.scale`.
> 
> Interesting, I was not aware of pep570! And sorry about that, I should have been more careful. Let's maybe reinstate it but rename it `braindecode.preprocessing.multiply`, to avoid confusion with sklearn's `scale` and `robust_scale`?

Ah, no worries. I also was not aware. Yes we should avoid confusion with sklearn, although I am very unhappy about their choices here. Both with forcing positional arguments as well as with naming 'scale'. I think `standard_scale` would be much more appropriate.

gemeinl:
> > @hubertjb I just had to learn that we cannot use `numpy.multiply` with our current `Preprocessor`. The reason is that `numpy.multiply` requires two **positional** arguments (it actually forces two positional arguments using `/`, see https://numpy.org/doc/stable/reference/generated/numpy.multiply.html, https://www.python.org/dev/peps/pep-0570/) where our `Preprocessor` only supports keyword arguments (https://github.com/braindecode/braindecode/blob/master/braindecode/preprocessing/preprocess.py#L45). Maybe it is worth keeping `braindecode.preprocessing.scale`.
> 
> Interesting, I was not aware of pep570! And sorry about that, I should have been more careful. Let's maybe reinstate it but rename it `braindecode.preprocessing.multiply`, to avoid confusion with sklearn's `scale` and `robust_scale`?

Ah, no worries. I also was not aware. Yes we should avoid confusion with sklearn, although I am very unhappy about their choices here. Both with forcing positional arguments as well as with naming 'scale'. I think standard_scale would be much more appropriate.

braindecode/preprocessing/preprocess.py
gemeinl: We could. If the user insists on using a lambda function it will show up as
`{'<lambda>': {}}` in the lists below.

gemeinl:
Maybe `list(tuple(str, dict(kwarg: value)))` would be more intuitive?

gemeinl:
Ready to merge from my side unless there are objections.

robintibor:
Please add description to whats_new

robintibor:
thanks

PR #288, opened 2021-06-27
[MRG] Unified saving logic + parallel loading

gemeinl:
We would like to create large datasets fast by using multiprocessing. Therefore, add joblib parallelization to `load_concat_dataset` as well as to `TUH` and `TUHAbnormal`.

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -481,7 +481,7 @@ class BaseConcatDataset(ConcatDataset):
                 raise FileExistsError(
                     f'{description_file_name} or {target_file_name} exist in {path}.')
         else:
-            _clear_up_directory(
+            self._clear_up_directory(
                 file_name_pattern, path, target_file_name, description_file_name)
         self._save_target_name(target_file_name)
         self._save_datasets_and_description(path, overwrite, file_name_pattern)
@@ -510,7 +510,7 @@ class BaseConcatDataset(ConcatDataset):
                         for ds in self.datasets]), (
                 "All datasets should have same target name")
             with open(target_file_name, 'w') as f:
-               json.dump({'target_name': expected_target_name}, f)
+                json.dump({'target_name': expected_target_name}, f)
 
     @staticmethod
     def _clear_up_directory(
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -15,7 +15,6 @@ import json
 from typing import Iterable
 import warnings
 from glob import glob
-from warnings import warn
 
 import numpy as np
 import pandas as pd
@@ -284,8 +283,8 @@ class BaseConcatDataset(ConcatDataset):
         Parameters
         ----------
         by : str | list
-            If ``by`` is a string, splitting is performed based on the description
-            DataFrame column with this name.
+            If ``by`` is a string, splitting is performed based on the
+            description DataFrame column with this name.
             If ``by`` is a (list of) list of integers, the position in the first
             list corresponds to the split id and the integers to the
             datapoints of that split.
@@ -375,20 +374,23 @@ class BaseConcatDataset(ConcatDataset):
         assure backwards compatibility by still being able to run the old tests.
 
         Save dataset to files.
+
         Parameters
         ----------
         path : str
             Directory to which .fif / -epo.fif and .json files are stored.
         overwrite : bool
-            Whether to delete old files (.json, .fif, -epo.fif) in specified directory
-            prior to saving.
+            Whether to delete old files (.json, .fif, -epo.fif) in specified
+            directory prior to saving.
         """
-        warn('This function only exists for backwards compatibilty purposes. '
-             'DO NOT USE!')
-        assert len(self.datasets) > 0, "Expect at least one dataset"
-        assert (hasattr(self.datasets[0], 'raw') + hasattr(
-            self.datasets[0], 'windows') == 1), (
-            "dataset should have either raw or windows attribute")
+        warnings.warn('This function only exists for backwards compatibility '
+                      'purposes. DO NOT USE!', UserWarning)
+        if len(self.datasets) == 0:
+            raise ValueError("Expect at least one dataset")
+        if not (hasattr(self.datasets[0], 'raw') or hasattr(
+                self.datasets[0], 'windows')):
+            raise ValueError("dataset should have either raw or windows "
+                             "attribute")
         file_names_ = ["{}-raw.fif", "{}-epo.fif"]
         description_file_name = os.path.join(path, 'description.json')
         target_file_name = os.path.join(path, 'target_name.json')
@@ -396,10 +398,12 @@ class BaseConcatDataset(ConcatDataset):
             if (os.path.exists(description_file_name) or
                     os.path.exists(target_file_name)):
                 raise FileExistsError(
-                    f'{description_file_name} or {target_file_name} exist in {path}.')
+                    f'{description_file_name} or {target_file_name} exist in '
+                    f'{path}.')
         else:
             for file_name in file_names_:
-                file_names = glob(os.path.join(path, f"*{file_name.lstrip('{}')}"))
+                file_names = glob(os.path.join(
+                    path, f"*{file_name.lstrip('{}')}"))
                 _ = [os.remove(f) for f in file_names]
             if os.path.isfile(target_file_name):
                 os.remove(target_file_name)
@@ -410,21 +414,22 @@ class BaseConcatDataset(ConcatDataset):
                 if os.path.exists(kwarg_path):
                     os.remove(kwarg_path)
 
-        concat_of_raws = hasattr(self.datasets[0], 'raw')
-        file_name = file_names_[0] if concat_of_raws else file_names_[1]
-        if concat_of_raws:
+        is_raw = hasattr(self.datasets[0], 'raw')
+        file_name = file_names_[0] if is_raw else file_names_[1]
+        if is_raw:
             # for checks that all have same target name and for
             # saving later
             target_name = self.datasets[0].target_name
         for i_ds, ds in enumerate(self.datasets):
             full_file_path = os.path.join(path, file_name.format(i_ds))
-            if concat_of_raws:
+            if is_raw:
                 ds.raw.save(full_file_path, overwrite=overwrite)
-                assert ds.target_name == target_name, "All datasets should have same target name"
+                assert ds.target_name == target_name, (
+                    "All datasets should have same target name")
             else:
                 ds.windows.save(full_file_path, overwrite=overwrite)
 
-        if concat_of_raws:
+        if is_raw:
             json.dump({'target_name': target_name}, open(target_file_name, 'w'))
         self.description.to_json(description_file_name)
         for kwarg_name in ['raw_preproc_kwargs', 'window_kwargs', 'window_preproc_kwargs']:
@@ -458,69 +463,86 @@ class BaseConcatDataset(ConcatDataset):
                 ds.set_description({key: value_}, overwrite=overwrite)
 
     def save(self, path, overwrite=False):
-        """Save datasets to files by creating one subdirectory for each dataset.
+        """Save datasets to files by creating one subdirectory for each dataset:
+        path/
+            target_name.json (if target_name is not None)
+            0/
+                0-raw.fif | 0-epo.fif
+                description.json
+            1/
+                1-raw.fif | 1-epo.fif
+                description.json
+            ...
 
         Parameters
         ----------
         path : str
-            Directory to which .fif / -epo.fif and .json files are stored.
+            Directory to which -raw.fif | -epo.fif and .json files are stored.
         overwrite : bool
-            Whether to delete old files (.json, .fif, -epo.fif) in specified directory
-            prior to saving.
+            Whether to delete old files (target_name.json, description.json,
+             -raw.fif, -epo.fif) in specified directory prior to saving.
         """
-        assert len(self.datasets) > 0, "Expect at least one dataset"
-        assert (hasattr(self.datasets[0], 'raw') + hasattr(
-            self.datasets[0], 'windows') == 1), (
-            "dataset should have either raw or windows attribute")
+        if len(self.datasets) == 0:
+            raise ValueError("Expect at least one dataset")
+        if not (hasattr(self.datasets[0], 'raw') or hasattr(
+                self.datasets[0], 'windows')):
+            raise ValueError("dataset should have either raw or windows "
+                             "attribute")
+        description_files = glob(os.path.join(path, '**/description.json'))
         file_name_pattern = ["{}-raw.fif", "{}-epo.fif"]
-        description_file_name = os.path.join(path, 'description.json')
-        target_file_name = os.path.join(path, 'target_name.json')
+        # read all -epo.fif and -raw.fif files
+        fif_files = []
+        for file_name in file_name_pattern:
+            fif_files.extend(
+                glob(os.path.join(path, f"**/*{file_name.lstrip('{}')}")))
         if not overwrite:
-            if (os.path.exists(description_file_name) or
-                    os.path.exists(target_file_name)):
+            if len(description_files) + len(fif_files) != 0:
                 raise FileExistsError(
-                    f'{description_file_name} or {target_file_name} exist in {path}.')
+                    'Another dataset was already saved to this directory. '
+                    'Please clean up manually, set overwrite to True, or save '
+                    'to another directory.')
         else:
-            self._clear_up_directory(
-                file_name_pattern, path, target_file_name, description_file_name)
-        self._save_target_name(target_file_name)
-        self._save_datasets_and_description(path, overwrite, file_name_pattern)
-
-    def _save_datasets_and_description(self, path, overwrite, file_name_pattern):
+            self._clear_up_directory(path, description_files, fif_files)
         is_raw = hasattr(self.datasets[0], 'raw')
         file_name = file_name_pattern[0] if is_raw else file_name_pattern[1]
+        self._save_datasets_and_description(path, overwrite, file_name, is_raw)
+        if is_raw:
+            self._save_target_name(path)
+
+    def _save_datasets_and_description(self, path, overwrite, file_name, is_raw):
         # save all the datasets and description
         for i_ds, ds in enumerate(self.datasets):
             # create one subdirectory per dataset
             this_path = os.path.join(path, str(i_ds))
-            os.makedirs(this_path)
+            if not os.path.exists(this_path):
+                os.makedirs(this_path)
             full_file_path = os.path.join(this_path, file_name.format(i_ds))
             if is_raw:
-                ds.raw.save(full_file_path, overwrite=overwrite)
+                ds.raw.save(full_file_path)
             else:
-                ds.windows.save(full_file_path, overwrite=overwrite)
+                ds.windows.save(full_file_path)
             ds.description.to_json(os.path.join(this_path, 'description.json'))
 
-    def _save_target_name(self, target_file_name):
+    def _save_target_name(self, path):
         # make sure we do not have an inconsistency in the target name
-        concat_of_raws = hasattr(self.datasets[0], 'raw')
-        if concat_of_raws:
-            expected_target_name = self.datasets[0].target_name
-            assert all([expected_target_name == ds.target_name
-                        for ds in self.datasets]), (
-                "All datasets should have same target name")
-            with open(target_file_name, 'w') as f:
-                json.dump({'target_name': expected_target_name}, f)
+        expected_target_name = self.datasets[0].target_name
+        assert all([expected_target_name == ds.target_name
+                    for ds in self.datasets]), (
+            "All datasets should have same target name")
+        with open(os.path.join(path, 'target_name.json'), 'w') as f:
+            json.dump({'target_name': expected_target_name}, f)
 
     @staticmethod
-    def _clear_up_directory(
-            file_name_pattern, path, target_file_name, description_file_name):
-        # remove all existing "{}-raw.fif" / "{}-epo.fif" files
-        for file_name in file_name_pattern:
-            file_names = glob(os.path.join(path, f"*{file_name.lstrip('{}')}"))
-            _ = [os.remove(f) for f in file_names]
-        # remove existing json files
+    def _clear_up_directory(path, description_files, fif_files):
+        # remove all existing "{}-raw.fif" | "{}-epo.fif" and description.json
+        # files. afterwards delete subdirectories if empty
+        for fif_file, description_file in zip(fif_files, description_files):
+            os.remove(fif_file)
+            os.remove(description_file)
+            subdir_name = os.path.dirname(fif_file)
+            if len(os.listdir(subdir_name)) == 0:
+                os.rmdir(subdir_name)
+        # remove target name file if it exists
+        target_file_name = os.path.join(path, 'target_name.json')
         if os.path.isfile(target_file_name):
             os.remove(target_file_name)
-        if os.path.isfile(description_file_name):
-            os.remove(description_file_name)
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -407,9 +407,9 @@ class BaseConcatDataset(ConcatDataset):
                     f'{description_file_name} or {target_file_name} exist in '
                     f'{path}.')
         else:
-            for file_name_temple in file_name_templates:
+            for file_name_template in file_name_templates:
                 file_names = glob(os.path.join(
-                    path, f"*{file_name_temple.lstrip('{}')}"))
+                    path, f"*{file_name_template.lstrip('{}')}"))
                 _ = [os.remove(f) for f in file_names]
             if os.path.isfile(target_file_name):
                 os.remove(target_file_name)
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -462,10 +462,10 @@ class BaseConcatDataset(ConcatDataset):
             for ds, value_ in zip(self.datasets, value):
                 ds.set_description({key: value_}, overwrite=overwrite)
 
-    def save(self, path, overwrite=False):
+    def save(self, path, overwrite=False, offset=None):
         """Save datasets to files by creating one subdirectory for each dataset:
         path/
-            target_name.json (if target_name is not None)
+            target_name.json (if target_name is not None and dataset is raw)
             0/
                 0-raw.fif | 0-epo.fif
                 description.json
@@ -481,6 +481,11 @@ class BaseConcatDataset(ConcatDataset):
         overwrite : bool
             Whether to delete old files (target_name.json, description.json,
              -raw.fif, -epo.fif) in specified directory prior to saving.
+        offset : int | None
+            If provided the integer is added to the id of the datasets in the
+            concat. This is useful in the setting of very large datasets, where
+            one dataset has to be processed and saved at a time to account for
+            its original position.
         """
         if len(self.datasets) == 0:
             raise ValueError("Expect at least one dataset")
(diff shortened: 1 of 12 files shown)


gemeinl:
Unfortunately, saving functionality was inconsistent for `BaseDatasets` and `WindowsDatasets`. Added a new saving logic which will:
- add `target_name.json` to the root directory if we have a concat of `BaseDatasets`
- create a subdirectory for every dataset in the concat
- save `description.json` as well as a `-fif` file to this directory

gemeinl:
I was unable to parallelize `load_concat_dataset` for windowed data using `preload=False`. The problem appears to be `mne.read_epochs` when `preload=False`, since this will crash with `TypeError: cannot pickle '_io.BufferedReader' object`. I therefore disabled parallel loading for this case in https://github.com/braindecode/braindecode/pull/288/files#diff-ffab4ab4433ba5e07e98af8ac01e96b531cca0eb9e7d1ffe2b24663551be030eR167-R171,

agramfort:
let's fix this upstream 

https://github.com/mne-tools/mne-python/issues/9508

braindecode/datasets/tuh.py
```diff
@@ -192,10 +200,11 @@ class TUHAbnormal(TUH):
         description.
     """
     def __init__(self, path, recording_ids=None, target_name='pathological',
-                 preload=False, add_physician_reports=False):
+                 preload=False, add_physician_reports=False, n_jobs=-1):
```
agramfort, line 203:
```suggestion
                 preload=False, add_physician_reports=False, n_jobs=1):
```

use n_jobs=1 by default everywhere please. It can create crazy issues on shared machines otherwise

agramfort:
@gemeinl or @hubertjb do you have a test script to test this feature on a real usecase?

gemeinl:
Runtime of reading TUHAbnormal files decreased by ~25% with n_jobs=2.
I figured, we might have similar benefits when loading preprocessed data, especially when using the TUH data with much more recordings. I therefore added parallelization to `load_concat_dataset` as well. So far, however, I cannot comment on runtime benefits of this, since I did not test it yet. We can postpone it / do this in a separate PR.

braindecode/datasets/tuh.py
gemeinl: Alright, thanks. Will use default of 1.

agramfort:
can you share a self contained code snippet for me to test on our systems?

eg using sleep data or TUH?

>

gemeinl:
Here is a very rudamentary script that loads the TUHAbnormal data, applies preprocessing, saves the data and reloads it.
```
import tempfile

from braindecode.datasets import TUHAbnormal
from braindecode.preprocessing.preprocess import preprocess, Preprocessor
from braindecode.datautil.serialization import load_concat_dataset


TUH_PATH = '/data/schirrmr/gemeinl/tuh-abnormal-eeg/raw/v2.0.0/edf/'
TMPDIR = tempfile.mkdtemp()
N_JOBS = 2

tuh_ab = TUHAbnormal(
    path=TUH_PATH,
    add_physician_reports=True,
    n_jobs=N_JOBS,
)
# do all kinds of preprocessing
preprocess(
    concat_ds=tuh_ab,
    preprocessors=[Preprocessor('crop', tmax=1)],
    # n_jobs=N_JOBS,  # PR277
)
# save preprocessed data
tuh_ab.save(TMPDIR)
# load preprocessed data to try different windoing parameters and for model training
load_concat_dataset(
    path=TMPDIR, 
    preload=False,
    n_jobs=N_JOBS,
)
```
The script fails on my machine with `RuntimeError: info["meas_date"] seconds must be between "(-2147483648, 0)" and "(2147483647, 0)", got "-2209161600"`. I had this error before and failed to reproduce, now I see it again.
More specifically, this is one recordings that fails due to a broken measurement date:
```
import mne
raw = mne.io.read_raw_edf("/data/schirrmr/gemeinl/tuh-abnormal-eeg/raw/v2.0.0/edf/eval/normal/01_tcp_ar/062/00006240/s001_2009_12_15/00006240_s001_t001.edf")
mne.io.meas_info._check_dates(raw.info)
```
Is there a way to fix the date?

agramfort:
@gemeinl you can fix your meas date issue with:

`[ds_.raw.set_meas_date() for ds_ in tuh_ab.datasets]`

it just removes all measurement dates from the data.

test/unit_tests/datautil/test_serialization.py
```diff
@@ -115,13 +108,71 @@ def test_load_multiple_concat_windows_dataset(setup_concat_windows_dataset,
...
     assert (2 * len(concat_windows_dataset.datasets) ==
             len(loaded_concat_windows_datasets.datasets))
     assert (2 * len(concat_windows_dataset.description) ==
             len(loaded_concat_windows_datasets.description))
+
+
+def test_save_concat_raw_dataset(setup_concat_raw_dataset, tmpdir):
+    concat_raw_dataset = setup_concat_raw_dataset
+    n_raw_datasets = len(concat_raw_dataset.datasets)
+    concat_raw_dataset.save(path=tmpdir, overwrite=False)
```
hubertjb, line 124: Should we add a test for `overwrite=True`? Also, we could also have a test for the case where raw data was saved first, followed by epoched data in the same directory (and vice versa).

braindecode/datasets/base.py
```diff
@@ -422,3 +428,71 @@ def set_description(self, description, overwrite=False):
         for key, value in description.items():
             for ds, value_ in zip(self.datasets, value):
                 ds.set_description({key: value_}, overwrite=overwrite)
+
+    def save(self, path, overwrite=False):
```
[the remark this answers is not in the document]
gemeinl: The code, however, belongs to the object, right? So for me it belongs to the base module.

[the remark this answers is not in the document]
gemeinl: I could move the `_outdated_save` until it can be removed, such that we will not have twice the amount of lines for saving.

braindecode/datasets/base.py
```diff
@@ -422,3 +428,71 @@ def set_description(self, description, overwrite=False):
...
+        """Save datasets to files by creating one subdirectory for each dataset.
+
```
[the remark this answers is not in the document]
gemeinl: Yes, good idea. Will add this.

braindecode/datasets/base.py
```diff
@@ -422,3 +428,71 @@ def set_description(self, description, overwrite=False):
...
+        Parameters
+        ----------
+        path : str
+            Directory to which .fif / -epo.fif and .json files are stored.
+        overwrite : bool
+            Whether to delete old files (.json, .fif, -epo.fif) in specified directory
+            prior to saving.
+        """
+        assert len(self.datasets) > 0, "Expect at least one dataset"
```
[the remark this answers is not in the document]
gemeinl: I don't think this was line was introduces in this PR but I'll change it to raising an error.

braindecode/datasets/base.py
```diff
@@ -422,3 +428,71 @@ def set_description(self, description, overwrite=False):
...
+        assert (hasattr(self.datasets[0], 'raw') + hasattr(
+            self.datasets[0], 'windows') == 1), (
+            "dataset should have either raw or windows attribute")
+        file_name_pattern = ["{}-raw.fif", "{}-epo.fif"]
+        description_file_name = os.path.join(path, 'description.json')
+        target_file_name = os.path.join(path, 'target_name.json')
+        if not overwrite:
```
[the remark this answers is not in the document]
gemeinl: It is always checked for `raw.fif` and `epo.fif`. If `overwrite=True`, all fif files will be deleted.

braindecode/datautil/serialization.py
```diff
@@ -109,3 +110,108 @@ def _load_signals(fif_file, preload, raws):
...
+    assert len(fif_files) == len(description_files)
+    # sort files by id of the datasets
+    fif_files = sorted(fif_files, key=lambda p: int(p.split(os.sep)[-2]))
+    description_files = sorted(
+        description_files, key=lambda p: int(p.split(os.sep)[-2]))
+    # optionally make a selection of files
+    if ids_to_load is not None:
+        fif_files = [fif_files[i] for i in ids_to_load]
+        description_files = [description_files[i] for i in ids_to_load]
+    raws = fif_files[0].endswith('-raw.fif')
```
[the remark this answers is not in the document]
gemeinl: Good catch, will update accordingly.

test/unit_tests/datautil/test_serialization.py
gemeinl: Yes, we should test it. It feels very error-prone.

braindecode/datasets/tuh.py
```diff
@@ -38,50 +39,57 @@ class TUH(BaseConcatDataset):
...
+        # disable joblib for tests. mocking seems to fail otherwise
+        if n_jobs == 1:
+            base_datasets = [self._create_dataset(
+                descriptions[i], target_name, preload, add_physician_reports)
+                for i in descriptions.columns]
+        else:
+            base_datasets = Parallel(n_jobs)(delayed(
+                self._create_dataset)(
+                descriptions[i], target_name, preload, add_physician_reports
+            ) for i in descriptions.columns)
```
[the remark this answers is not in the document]
gemeinl: It is a row in the end. concat_ds.description will be `n_recordings x n_description_values` I don't remember the exact reasons (maybe speed, or just convenience) twhy it is not rows at this point. Do you think it should always be rows also in code that is not exposed to users?

braindecode/datasets/tuh.py
```diff
@@ -38,50 +39,57 @@ class TUH(BaseConcatDataset):
...
         descriptions = _create_chronological_description(file_paths)
         # limit to specified recording ids before doing slow stuff
         if recording_ids is not None:
             descriptions = descriptions[recording_ids]
+        # this is the second loop (slow)
         # create datasets gathering more info about the files touching them
         # reading the raws and potentially preloading the data
-        base_datasets = self._create_datasets(
-            descriptions, target_name, preload, add_physician_reports)
+        # disable joblib for tests. mocking seems to fail otherwise
```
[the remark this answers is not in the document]
gemeinl: I don't remember the details of the proposal or why we did not do this yet, but I think it would be good to follow your suggestion.

braindecode/datasets/base.py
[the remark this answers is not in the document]
hubertjb: OK, works for me. No need to move `_outdated_save` I guess then.

braindecode/datasets/tuh.py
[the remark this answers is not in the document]
hubertjb: Right, the final attribute is already the right shape, it's just `descriptions` that is transposed. I don't think this makes a big difference for internal code - it just feels more natural to me to iterate over rows. Whatever you prefer :)

gemeinl:
> @gemeinl you can fix your meas date issue with:
> 
> `[ds_.raw.set_meas_date() for ds_ in tuh_ab.datasets]`
> 
> it just removes all measurement dates from the data.

Ah, great. Thank you!

gemeinl:
Best to wait for https://github.com/braindecode/braindecode/pull/277 to merge, we might need additional changes to `BaseConcatDataset.save()` to properly handle big datasets and serialization. I think we need something like an offset id, since the preprocessing function preprocesses one dataset at a time and saves it as a BaseConcatDataset with a single entry. So save will save to `0/0-raw.fif`, where we would like to add the offset id to account for the real position of the dataset in the original concat.

gemeinl:
Desired serialization logic in https://github.com/braindecode/braindecode/pull/277 will break with overwrite = True and I don't see a way to fix it. There are two possible ways to handle overwriting:
1) overwrite single files. When we attempt to write `save_dir/subdir/-raw.fif`, we will delete this file if overwrite and create this file again. However, in this case, we might end up with a mess of old and new files if we had more 'old' files than we save 'new' files.
2) overwrite the entire directory. When we attempt to write the first file, we will delete all the `-raw.fif`, `-epo.fif`, `description.json` and `target_name.json` in `save_dir`. That's what we do right now. However, if we want to process large datasets, where recordings are processed and serialized one at a time, the directory will be deleted over and over again and we will end up with a single fif file (the last in the list).
I think it would probably be easiest and best to abandon `overwrite` and force the user to write to an empty directory.

gemeinl:
We also get into trouble if the user attempts to overwrite raws and currently uses these raws with `preload=False`. If overwrite is True in that case, mne will crash, since it needs access to the raws on disk. If `preload=True`, there is no problem.

gemeinl:
Done from my side but probably not ready to merge yet. Should be discussed in more detail, maybe there are better options I don't see.

braindecode/datasets/base.py
```diff
@@ -438,3 +449,114 @@ def set_description(self, description, overwrite=False):
...
+        path : str
+            Directory to which -raw.fif | -epo.fif and .json files are stored.
+        overwrite : bool
+            Whether to delete old files (target_name.json, description.json,
+             -raw.fif, -epo.fif) in specified directory prior to saving.
+        offset : int | None
+            If provided the integer is added to the id of the datasets in the
+            concat. This is useful in the setting of very large datasets, where
+            one dataset has to be processed and saved at a time to account for
+            its original position.
```
robintibor, line 476: so just for my understanding, this way one can also save into the same parent directory step by step all the datasets? do we have a test for this?

braindecode/datautil/serialization.py
```diff
@@ -98,14 +101,107 @@ def _load_signals_and_description(path, preload, raws, ids_to_load=None):
...
+        'Number of descriptions and .fif files does not match.')
+    # sort files by id of the datasets
+    fif_files = sorted(fif_files, key=lambda p: int(p.split(os.sep)[-2]))
+    description_files = sorted(description_files,
+                               key=lambda p: int(p.split(os.sep)[-2]))
+    # optionally make a selection of files
+    if ids_to_load is not None:
+        fif_files = [fif_files[i] for i in ids_to_load]
+        description_files = [description_files[i] for i in ids_to_load]
+    is_raw = fif_files[0].endswith('-raw.fif')
```
robintibor, line 167: maybe very edge case, but what would happen if we have both raw and epochs in one folder? just for my understanding

robintibor:
overall seems quite sensible to me? Also add to what's new, and there are conflicts atm

hubertjb:
Concerning your comment above on `overwrite=True` @gemeinl

1. Overwriting single files. What if we added a check at the beginning of `preprocess` to see how many files currently exist under the provided directory, and raise an error if the number is not the same as the number of datasets in the input? This would avoid ending up with a mix of old and new files. Would that work with scenario 2 (sequential preprocessing)?

As for abandoning `overwrite` altogether, I agree that this would be much simpler code-wise, but I wonder whether this would be annoying from a user's perspective. In a typical pipeline where you (1) load data, (2) preprocess it, (3) window it, (4) preprocess the windows, this means you would have to provide four `save_dir`s and potentially simultaneously keep on disk four "versions" of the data. What do you think @gemeinl ?

Also, another option I proposed [here](https://github.com/braindecode/braindecode/pull/277#discussion_r666443800) to make serialization safer without asking the user to explicitly provide `save_dir` every time would be to automatically create a new directory with a similar name (e.g. by adding a numerical suffix to the original directory) and save the output there.

hubertjb:
> We also get into trouble if the user uses raws with `preload=False` and deletes them by overwriting or accidentally. In that case, mne will crash on save, since it needs access to the raws on disk. If `preload=True`, there is no problem.

@gemeinl can you provide an example of this? Could this be avoided by a check in `BaseConcatDataset.save()` that raises an error if the user's trying to overwrite raws with preload=False?

hubertjb:
One other edge case I just ran into is if you save a BaseConcatDataset under `save_dir` and then pass the same `save_dir` to `preprocess`. Right now I believe it will break because the two ways of saving files are conflicting (save_dir/0-raw.fif vs. save_dir/0/0-raw.fif). I'm not sure we need to address this right now, but we might need to put more thought into this.

braindecode/datautil/serialization.py
gemeinl: If the raws are not fif, then everything should work as expected (would have to verify this).

If raws are stored as fif, it should not be possible to create this situation through braindecode functionality. If the user did this himself, then
For saving:
- our code will crash if overwrite = False
- our code will delete all -raw.fif and -epochs.fif and description.json with overwrite = True

For loading:
- will crash if number of .fif files and description.json files does not match
- it will crash later since the reading function is determined by first fif file, so either we will try to read raws with mne.read_epochs or epochs with mne.io.read_raw_fif


We could add an explicit test somewhere before this happens. So check that not -raw.fif and -epo.fif are in the directory as well as only allow -raw.fif / -epo.fif and description.json files. Crash if anything else exists? Or how to check for other raw data files?

gemeinl:
> One other edge case I just ran into is if you save a BaseConcatDataset under `save_dir` and then pass the same `save_dir` to `preprocess`. Right now I believe it will break because the two ways of saving files are conflicting (save_dir/0-raw.fif vs. save_dir/0/0-raw.fif). I'm not sure we need to address this right now, but we might need to put more thought into this.

With this PR the user will get a warning if he loads a dataset that was saved in the old way with the prompt to convert it to the new way (https://github.com/braindecode/braindecode/pull/288/files#diff-ffab4ab4433ba5e07e98af8ac01e96b531cca0eb9e7d1ffe2b24663551be030eR142-R144). So yes, it will crash, but it is on the user if he does not take care of the warning?

gemeinl:
> > We also get into trouble if the user uses raws with `preload=False` and deletes them by overwriting or accidentally. In that case, mne will crash on save, since it needs access to the raws on disk. If `preload=True`, there is no problem.
> 
> @gemeinl can you provide an example of this? Could this be avoided by a check in `BaseConcatDataset.save()` that raises an error if the user's trying to overwrite raws with preload=False?

I will create a small snippet to demonstrate this.

braindecode/datasets/base.py
gemeinl: This was the idea but it is actually on https://github.com/braindecode/braindecode/pull/277#discussion_r666443800 to implement it.

gemeinl: Not a test strictly speaking, but https://github.com/braindecode/braindecode/pull/277#discussion_r666443800 uses it and would crash if it didn't work. Would probably be good to add a dedicated test.

gemeinl:
> Concerning your comment above on `overwrite=True` @gemeinl
> 
>     1. Overwriting single files. What if we added a check at the beginning of `preprocess` to see how many files currently exist under the provided directory, and raise an error if the number is not the same as the number of datasets in the input? This would avoid ending up with a mix of old and new files. Would that work with scenario 2 (sequential preprocessing)?

And if the number changes we ask the user to resolve this, right? So we could also just ask him to provide a proper (empty) directory in the first place?

> 
> As for abandoning `overwrite` altogether, I agree that this would be much simpler code-wise, but I wonder whether this would be annoying from a user's perspective. In a typical pipeline where you (1) load data, (2) preprocess it, (3) window it, (4) preprocess the windows, this means you would have to provide four `save_dir`s and potentially simultaneously keep on disk four "versions" of the data. What do you think @gemeinl ?

I don't see how you get to four, but probably our use cases are just so different? 
1) a raw data directory exists or one fetches data through moabb
2) provide a save_dir to preprocess or save manually afterwards
3) do not save again after windowing
4) do not save again after window preprocessing

How long does 3) + 4) take for you? Would it be possible to not save afterwards?

> Also, another option I proposed [here](https://github.com/braindecode/braindecode/pull/277#discussion_r666443800) to make serialization safer without asking the user to explicitly provide `save_dir` every time would be to automatically create a new directory with a similar name (e.g. by adding a numerical suffix to the original directory) and save the output there.

So this would be the same as forcing an empty directory, right? We just don't expect to get it from the user but create it internally?

gemeinl:
I just had another idea. The file path should be stored somewhere in the mne object, I guess. So what if we do not provide `overwrite` but something like `delete_old`? So we will require / create a new directory and write the dataset one dataset at a time. Simultaneously we delete the old file connected to the object. It might leave artifacts in the original directory, if the number of files changed, but it terms of disk space it is as little as one recording overhead.

gemeinl:
> > > We also get into trouble if the user uses raws with `preload=False` and deletes them by overwriting or accidentally. In that case, mne will crash on save, since it needs access to the raws on disk. If `preload=True`, there is no problem.
> > 
> > 
> > @gemeinl can you provide an example of this? Could this be avoided by a check in `BaseConcatDataset.save()` that raises an error if the user's trying to overwrite raws with preload=False?
> 
> I will create a small snippet to demonstrate this.
> Yes, I think it could but again adds more complexity.

```
import tempfile
from braindecode.datautil.serialization import load_concat_dataset

ds = MOABBDataset(
    dataset_name='BNCI2014001',
    subject_ids=1,
)
ds = ds.split([0])['0']
save_dir = tempfile.mkdtemp()
ds.save(save_dir)
ds = load_concat_dataset(save_dir, preload=False)  # preload = True to resolve
windows_ds = create_windows_from_events(ds)
# crashes since raw with preload = False is deleted before epochs is saved
windows_ds.save(save_dir, overwrite=True)
```

hubertjb:
> And if the number changes we ask the user to resolve this, right? So we could also just ask him to provide a proper (empty) directory in the first place?

I guess one likely scenario is you're working with a smallish dataset to see e.g. the effect of different preprocessing hyperparameters on learning. You might not want to use too much disk space, so ideally each time you run the experiment with different preprocessing settings you'd like to overwrite the preprocessed data from the previous experiment. It could be annoying to have to manually delete the data each time. The alternative would be to ask the user to write their own code to handle this. My preference would be to handle overwriting internally so it is very simple to do something like this.

After this PR is merged I could add a check that makes sure the number of files is the same. The function could be reused in BaseConcatDataset.save() too I guess.

> I don't see how you get to four, but probably our use cases are just so different?
> 
> How long does 3) + 4) take for you? Would it be possible to not save afterwards?

Ah right, it won't be 4, but actually 3 versions of the data. The reason it's 3 is that if your window preprocessing requires loading data into memory but you want to use preload=False, you need to provide a second `save_dir` so the data can be saved and reloaded. So even in the (likely) case that 3) + 4) is not that long to run as compared to 1) and 2) you might need to save a third version of the data.

hubertjb:
> I just had another idea: The file path should be stored somewhere in the mne object, I guess. So what if we do not provide `overwrite` but something like `delete_old`? So we will require / create a new directory and write one dataset at a time. Simultaneously, we delete the old file connected to the object. It might leave artifacts in the original directory, if the number of files changed, but in terms of disk space it is as little as one recording overhead. We could even rename / move the new folder to the first one, if the first one ended up empty.

Do you know where the file path would be recorded @gemeinl? I spent 5 minutes in the docs and couldn't find it. :P @agramfort

hubertjb:
Got it, thanks for the snippet @gemeinl ! I guess we'll need a check for that too then if we decide to keep support for overwriting, seems like an important use case...

hubertjb:
I think it would make sense to merge this and #277 soon so we can start using the parallelization/serialization features and see whether they fit our usage. Since it seems there's still quite a few potential edge cases with the overwriting feature in `preprocess` I'm OK with removing it for now and revisiting it later if it ends up being a pain point.

robintibor:
could be merged if whats_new is there and conflicts are resolved

gemeinl:
> Ah right, it won't be 4, but actually 3 versions of the data. The reason it's 3 is that if your window preprocessing requires loading data into memory but you want to use preload=False, you need to provide a second `save_dir` so the data can be saved and reloaded. So even in the (likely) case that 3) + 4) is not that long to run as compared to 1) and 2) you might need to save a third version of the data.
Got it, if you want to use preload=False you are actually forced to save, since there is no way to 'unload' the data again...

robintibor:
So regarding override, after some conversation with @gemeinl , here is my suggestion:

`override=False`:
* Crash when *subdirectory* (e.g. 0,1,2,etc) already exists
* Save if subdirectory didn't exist, independent of if other subdirectories exist inside
* Give warning if other subdirectories existed inside already

`override=True`:
* Override when  *subdirectory* (e.g. 0,1,2,etc) already exists
* Give warning if other subdirectories existed inside already (potentially only if there are subdirectories that won't be overriden

gemeinl:
In addition, we agreed to move all files to the `BaseDataset` level and to not have any files at the `BaseConcatDataset` level, right? So `save_dir` will not contain any files but only subdirectories. In the subdirectories there will be raw.fif / epo.fif, description.json, target_name.json, raw_preproc_kwargs.json, window_kwargs.json, and window_preproc_kwargs.json.

gemeinl:
`overwrite=True`:
...
* Give warning if other subdirectories existed inside already (potentially only if there are subdirectories that won't be overriden

In the context of https://github.com/braindecode/braindecode/pull/277 this will then create as many warnings as datasets.

hubertjb:
> So regarding override, after some conversation with @gemeinl , here is my suggestion ...

Very cool, I agree with your suggestion! Also I agree the repeated warning thing in `preprocess` might be annoying. One solution would be to suppress warnings in `_preprocess` directly...

agramfort:
@gemeinl can you rebase to fix conflicts and make the CIs green? 🙏

braindecode/datasets/base.py
```diff
@@ -368,60 +374,73 @@ def target_transform(self, fn):
...
-            "dataset should have either raw or windows attribute")
+        warnings.warn('This function only exists for backwards compatibility '
+                      'purposes. DO NOT USE!', UserWarning)
+        if len(self.datasets) == 0:
+            raise ValueError("Expect at least one dataset")
+        if not (hasattr(self.datasets[0], 'raw') or hasattr(
+                self.datasets[0], 'windows')):
+            raise ValueError("dataset should have either raw or windows "
+                             "attribute")
         file_names_ = ["{}-raw.fif", "{}-epo.fif"]
```
robintibor, line 400: I think this part becomes a bit confusing with `file_names_` `file_names` etc. suggest to rename:
`file_names_` to `file_templates` or `file_name_templates`. And also further below change `file_name` to `file_template` or `file_name_template` @gemeinl

braindecode/datasets/base.py
```diff
@@ -450,3 +469,115 @@ def set_description(self, description, overwrite=False):
...
+                target_name.json (if target_name is not None and dataset is raw)
+            ...
+
+        Parameters
+        ----------
+        path : str
+            Directory in which subdirectories are created to store
+             -raw.fif | -epo.fif and .json files to.
+        overwrite : bool
+            Whether to delete old subdirectories.
```
robintibor, line 498:
```suggestion
            Whether to delete old subdirectories that will be saved to in this call.
```

braindecode/datasets/base.py
```diff
@@ -450,3 +469,115 @@ def set_description(self, description, overwrite=False):
...
+            # save_dir/{i_ds+offset}/window_kwargs.json,
+            # save_dir/{i_ds+offset}/window_preproc_kwargs.json
+            self._save_kwargs(sub_dir, ds)
+            # save_dir/{i_ds+offset}/target_name.json
+            self._save_target_name(sub_dir, ds)
+        if overwrite:
+            # the following will be True for all datasets preprocessed and
+            # stored in parallel with braindecode.preprocessing.preprocess
+            if i_ds+1+offset != n_sub_dirs:
+                warnings.warn(f"The number of saved datasets ({i_ds+1+offset}) "
```
robintibor, line 542: great

braindecode/datasets/base.py
```diff
@@ -368,60 +374,73 @@ def target_transform(self, fn):
...
+                               'window_preproc_kwargs']:
                 kwarg_path = os.path.join(path, '.'.join([kwarg_name, 'json']))
                 if os.path.exists(kwarg_path):
                     os.remove(kwarg_path)
 
-        concat_of_raws = hasattr(self.datasets[0], 'raw')
-        file_name = file_names_[0] if concat_of_raws else file_names_[1]
-        if concat_of_raws:
+        is_raw = hasattr(self.datasets[0], 'raw')
+        file_name = file_names_[0] if is_raw else file_names_[1]
```
robintibor, line 425:
```suggestion
        file_name_template = file_names_[0] if is_raw else file_names_[1]
```

braindecode/datasets/base.py
```diff
@@ -368,60 +374,73 @@ def target_transform(self, fn):
...
+        if is_raw:
             # for checks that all have same target name and for
             # saving later
             target_name = self.datasets[0].target_name
         for i_ds, ds in enumerate(self.datasets):
             full_file_path = os.path.join(path, file_name.format(i_ds))
```
robintibor, line 431:
```suggestion
            full_file_path = os.path.join(path, file_name_template.format(i_ds))
```

robintibor (changes requested):
See in code @gemeinl

braindecode/datasets/base.py
```diff
@@ -368,60 +374,76 @@ def target_transform(self, fn):
...
             if (os.path.exists(description_file_name) or
                     os.path.exists(target_file_name)):
                 raise FileExistsError(
-                    f'{description_file_name} or {target_file_name} exist in {path}.')
+                    f'{description_file_name} or {target_file_name} exist in '
+                    f'{path}.')
         else:
-            for file_name in file_names_:
-                file_names = glob(os.path.join(path, f"*{file_name.lstrip('{}')}"))
+            for file_name_temple in file_name_templates:
```
robintibor, line 410: `temple` typo or intended?

gemeinl: typo =/

robintibor:
Amazing! The monster is merged! :D Great work! @gemeinl @hubertjb

MohammadJavadD:
> Here is a very rudimentary script that loads the TUHAbnormal data, applies preprocessing, saves the data and reloads it.
> 
> ```
> import tempfile
> 
> from braindecode.datasets import TUHAbnormal
> from braindecode.preprocessing.preprocess import preprocess, Preprocessor
> from braindecode.datautil.serialization import load_concat_dataset
> 
> 
> TUH_PATH = '/data/schirrmr/gemeinl/tuh-abnormal-eeg/raw/v2.0.0/edf/'
> TMPDIR = tempfile.mkdtemp()
> N_JOBS = 2
> 
> tuh_ab = TUHAbnormal(
>     path=TUH_PATH,
>     add_physician_reports=True,
>     n_jobs=N_JOBS,
> )
> # do all kinds of preprocessing
> preprocess(
>     concat_ds=tuh_ab,
>     preprocessors=[Preprocessor('crop', tmax=1)],
>     # n_jobs=N_JOBS,  # PR277
> )
> # save preprocessed data
> tuh_ab.save(TMPDIR)
> # load preprocessed data to try different windoing parameters and for model training
> load_concat_dataset(
>     path=TMPDIR, 
>     preload=False,
>     n_jobs=N_JOBS,
> )
> ```
> 
> The script fails on my machine with `RuntimeError: info["meas_date"] seconds must be between "(-2147483648, 0)" and "(2147483647, 0)", got "-2209161600"`. I had this error before and failed to reproduce, now I see it again. More specifically, this is one recordings that fails due to a broken measurement date:
> 
> ```
> import mne
> raw = mne.io.read_raw_edf("/data/schirrmr/gemeinl/tuh-abnormal-eeg/raw/v2.0.0/edf/eval/normal/01_tcp_ar/062/00006240/s001_2009_12_15/00006240_s001_t001.edf")
> mne.io.meas_info._check_dates(raw.info)
> ```
> 
> The problem does not show unless you try to save the recording. Is there a way to fix the date?

I'm getting this error as well! and the suggested solution (to remove dates) doesn’t work!

PR #291, opened 2021-06-30
[MRG] Adding robust scaler preprocessor

hubertjb:
Robust scaler to be used in the USleep example (#282).

### DIFF
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -249,19 +249,17 @@ def exponential_moving_demean(data, factor_new=0.001, init_block_size=None):
 
 def zscore(data):
     """Zscore normalize continuous or windowed data in-place.
-
     Parameters
     ----------
-    data: np.ndarray (n_channels, n_times) or (n_windows, n_channels, n_times)
-        continuous or windowed signal
-
+    data: np.ndarray
+        Continuous or windowed data of shape (n_channels, n_times) or
+        (n_windows, n_channels, n_times).
     Returns
     -------
-    zscored: np.ndarray (n_channels x n_times) or (n_windows x n_channels x
-    n_times)
-        normalized continuous or windowed data
-
-    ..note:
+    np.ndarray :
+        Normalized continuous or windowed data of shape (n_channels, n_times)
+        or (n_windows, n_channels, n_times).
+    .. note::
         If this function is supposed to preprocess continuous data, it should be
         given to raw.apply_function().
     """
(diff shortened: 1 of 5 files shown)


braindecode/preprocessing/preprocess.py
```diff
@@ -274,23 +275,53 @@ def zscore(data):
...
+
+    .. note::
+        If this function is supposed to preprocess continuous data, it should be
+        given to raw.apply_function().
+    """
+    scaled = data - np.median(data, keepdims=True, axis=-1)
+    q75, q25 = np.percentile(scaled, [75, 25], keepdims=True, axis=-1)
+    scaled = scaled / (q75 - q25)
+    if hasattr(data, '_data'):
+        data._data = scaled
```
agramfort, line 303: I don’t like that. Docsting says data is ndarray that will be modified inplace but inplace occurs only for Evoked and Epochs (not explained from docstring) and assume hack a private attribute only present if preload = True. You should be using apply_function here.

hubertjb: It was an artifact of the other preprocessors - I agree with you. I removed them with the last commit.

braindecode/preprocessing/preprocess.py
```diff
@@ -252,54 +252,66 @@ def zscore(data):
...
+    Returns
+    -------
+    np.ndarray :
+        Normalized continuous or windowed data of shape (n_channels, n_times)
+        or (n_windows, n_channels, n_times).
+    """
+    scaled = data - np.median(data, keepdims=True, axis=-1)
+    q75, q25 = np.percentile(scaled, [75, 25], keepdims=True, axis=-1)
+    scaled = scaled / (q75 - q25)
+    return scaled
```
agramfort, line 294: this is not operating in-place no?

gemeinl:
Is there a reason you don't just use `sklearn.preprocessing.robust_scale`?
Like so:
```
from sklearn.preprocessing import robust_scale
from braindecode.preprocessing import Preprocessor, preprocess
preprocess(
    dataset,
    preprocessors=[Preprocessor(robust_scale, chanel_wise=True)],
)
```
Should give you exactly the same results.

gemeinl:
Also `zscore` could potentially be removed and replaced by `sklearn.preprocessing.scale` if I am not mistaken.

hubertjb:
> Also `zscore` could potentially be removed and replaced by `sklearn.preprocessing.scale` if I am not mistaken.

Ah you're right @gemeinl, I didn't know sklearn already had an implementation... I guess then we should remove both `robust_scale` and `zscore` and replace them with sklearn's  `robust_scale` and `scale`. Should we also remove braindecode's `scale` and recommend the direct use of `np.multiply`?

I'll close this PR and open a new one to remove these functions and their tests, what do you think? @agramfort @gemeinl

robintibor:
sounds good, removing code always makes the world a bit more beautiful :D

agramfort:
+1000

hubertjb:
By the way, do you prefer we remove them entirely right now, or we start by deprecating them in 0.5.2?

agramfort:
>
> By the way, do you prefer we remove them entirely right now, or we start
> by deprecating them in 0.5.2?
>

theory says:
-  any released public code should be deprecated
- you don't deprecate with minor versions so you can only have deprecation
if next release is 0.6 (which we should do)

Issue #292, opened 2021-07-01
Add example of CNN LSTM


balandongiv:
Great work from the `braindecode` team. I believe this package will further help future researcher to dwell more into deep learning territory. 

If I can offer a suggestion to the team to add extra example.
Perhaps, the team can consider to add example the implementation of CNN+LSTM using the sleep public dataset. I notice, the type of framework has been explored by many researcher for quite some time such as [AutoSleepScorer](https://github.com/skjerns/AutoSleepScorer).

But, this also might help potential user understand how the can extend further the application of `braindecode` by integrating various other DL technique.

agramfort:
@balandongiv to add this model to our model zoo we would need to convert this model to PyTorch.

would you be willing to help?

PR #293, opened 2021-07-02
[MRG] Deprecating preprocessing functions zscore and scale

hubertjb:
Following discussion in #292, this PR deprecates functions `zscore` and `scale`, which can be replaced by `sklearn.preprocessing.scale` and `np.multiply`, respectively.

### DIFF
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -220,7 +220,6 @@ def exponential_moving_demean(data, factor_new=0.001, init_block_size=None):
     Deman the data point :math:`x_t` at time `t` as:
     :math:`x'_t=(x_t - m_t)`.
 
-
     Parameters
     ----------
     data: np.ndarray (n_channels, n_times)
@@ -265,6 +264,7 @@ def zscore(data):
         If this function is supposed to preprocess continuous data, it should be
         given to raw.apply_function().
     """
+    warn('zscore is deprecated. Use sklearn.preprocessing.scale instead.')
     zscored = data - np.mean(data, keepdims=True, axis=-1)
     zscored = zscored / np.std(zscored, keepdims=True, axis=-1)
     # TODO: the overriding of protected '_data' should be implemented in the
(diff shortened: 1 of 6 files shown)


braindecode/preprocessing/preprocess.py
```diff
@@ -265,6 +264,7 @@ def zscore(data):
         If this function is supposed to preprocess continuous data, it should be
         given to raw.apply_function().
     """
+    warn('zscore is deprecated. Use sklearn.preprocessing.scale instead.')
```
agramfort, line 267: you should warn  with DeprecationWarning and UserWarning

agramfort (commented):
also you could consider using such a decorator used in sklearn, mne and many other projects

https://github.com/mne-tools/mne-python/blob/main/mne/utils/docs.py#L2756-L2847

braindecode/preprocessing/preprocess.py
hubertjb: Ah nice, let's use `mne.utils.deprecated` then! However it looks like it will only warn with `DeprecationWarning`, not `UserWarning`?

agramfort:
yes only DeprecationWarning is needed

>

hubertjb:
> yes only DeprecationWarning is needed
> […](#)

I've just tested with `mne.utils.deprecated` and it looks like `DeprecationWarning` is not shown to the user by default. We could use `sklearn.utils.deprecation.deprecated` instead which throws a `FutureWarning` that is shown by default. Do you have a preference @agramfort ?

agramfort:
When you deprecate code you should specify the version number when it will be deprecated.

hubertjb:
> When you deprecate code you should specify the version number when it will be deprecated.

I wrote that they will be removed in 0.7.0, supposing next version will be 0.6.0 and we will want to wait for one minor version before removing these functions.

agramfort (approved):
@robintibor merge if happy

robintibor:
atm have some side effects from second shot, will look when feeling better

robintibor:
Great thanks merged

PR #294, opened 2021-07-06
[MRG] Self-ensembling for sequence-to-sequence models

hubertjb:
This PR introduces a utility function `aggregate_probas` to perform self-ensembling as described in [Phan et al. (2018)](https://ieeexplore.ieee.org/abstract/document/8502139). This works by aggregating the window-wise predictions of a seq2seq model when fed with overlapping sequences. This will be useful e.g. in #282.

To simplify its use, I also added an attribute `file_ids` to `SequenceSampler` so we can easily see which file each sequence comes from. 

In the end, the function can be used in the following way given a seq2seq model `clf` (e.g. an `EEGClassifier` that wraps `USleep`):

```python
y_true = np.array([valid_set[i][1] for i in range(len(valid_set))])
logits = clf.predict_proba(valid_set)
y_pred = np.concatenate(
    [aggregate_probas(logits[valid_sampler.file_ids == id]) 
     for id in np.unique(valid_sampler.file_ids)])
     
# Then, use `y_true` and `y_pred` to measure performance 
``` 

@Tgnassou @agramfort

### DIFF
--- a/braindecode/models/util.py
+++ b/braindecode/models/util.py
@@ -87,9 +87,9 @@ def _pad_shift_array(x, stride=1):
     Returns
     -------
     np.ndarray :
-        Array of shape (n_rows, n_classes, n_rows + n_windows - 1) where each
-        row is obtained by zero-padding the corresponding row in `x` before and
-        after in the last dimension.
+        Array of shape (n_rows, n_classes, (n_rows - 1) * stride + n_windows)
+        where each row is obtained by zero-padding the corresponding row in `x`
+        before and after in the last dimension.
     """
     if x.ndim != 3:
         raise NotImplementedError(
--- a/braindecode/models/util.py
+++ b/braindecode/models/util.py
@@ -4,6 +4,7 @@
 
 import torch
 import numpy as np
+from scipy.special import log_softmax
 
 
 def to_dense_prediction_model(model, axis=(2, 3)):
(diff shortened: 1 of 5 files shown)


braindecode/models/util.py
```diff
@@ -68,3 +69,69 @@ def get_output_shape(model, in_chans, input_window_samples):
...
+    Returns
+    -------
+    np.ndarray :
+        Array of shape (n_rows, n_classes, (n_rows - 1) * stride + n_windows)
+        where each row is obtained by zero-padding the corresponding row in `x`
+        before and after in the last dimension.
+    """
+    if x.ndim != 3:
+        raise NotImplementedError(
+            f'x must be of shape (n_rows, n_clases, n_windows), got {x.shape}')
```
agramfort, line 96:
```suggestion
            f'x must be of shape (n_rows, n_classes, n_windows), got {x.shape}')
```

agramfort (commented):
have you tried this on a real dataset already @hubertjb ?

test/unit_tests/models/test_util.py
```diff
@@ -20,3 +24,44 @@ def test_get_output_shape_2d_model():
...
+    n_outputs = (n_sequences - 1) * stride + n_windows
+    shifted_y = np.concatenate([
+        np.concatenate((
+            np.zeros((1, n_classes, i * stride)), dense_y[[i]],
+            np.zeros((1, n_classes, n_outputs - n_windows - i * stride))),
+            axis=2)
+        for i in range(n_sequences)], axis=0)
+    shifted_y2 = _pad_shift_array(dense_y, stride=stride)
+
+    assert (shifted_y == shifted_y2).all()
```
robintibor, line 45: are you just testing zero arrays here? might be good to have somethign that tests actual values as well, like 1-2 more explicit tests with predefined values

test/unit_tests/models/test_util.py
```diff
@@ -20,3 +24,44 @@ def test_get_output_shape_2d_model():
...
+    logits = OneHotEncoder(sparse=False).fit_transform(y_true.reshape(-1, 1))
+    logits = np.lib.stride_tricks.sliding_window_view(
+        logits, n_windows, axis=0)[::stride]
+
+    y_pred_probas = aggregate_probas(logits, n_windows_stride=stride)
+
+    assert y_pred_probas.ndim == 2
+    assert y_pred_probas.shape[0] == n_outputs
+    assert y_pred_probas.shape[1] == n_classes
+    assert (y_true == y_pred_probas.argmax(axis=1)).all()
```
robintibor, line 67: could you describe a bit more what you are testing here? like write a comment what is being tested? are values actually being tested? or just shapes?

test/unit_tests/models/test_util.py
hubertjb: It is testing against actual values - the concatenation only adds zeros around the values in `dense_y`.

test/unit_tests/models/test_util.py
hubertjb: I'll add a few comments to make this clearer. But yes, this tests both shapes and values!

robintibor:
maybe write a short line in whats new @hubertjb ? Then I can merge

robintibor:
Great, merged

PR #261, opened 2021-06-21
WIP: Discrete and synchronized targets support

sliwy:

PR #295, opened 2021-07-06
[MRG] Random sampler

Tgnassou:
Add a new sampler inspired by USleep article. We define a fixed number of sequence. For each sequence, a window is chosen for a random recording and a random class. The sequence is created around this window with a random position. Each class has the same probability to be chosen. So this method allow  to equitably represent the classes despite their difference in proportion.

@agramfort @hubertjb @l-omar-chehab

### DIFF
--- a/braindecode/samplers/base.py
+++ b/braindecode/samplers/base.py
@@ -148,7 +148,7 @@ class SequenceSampler(RecordingSampler):
             yield tuple(range(start_ind, start_ind + self.n_windows))
 
 
-class RandomSampler(RecordingSampler):
+class BalancedSequenceSampler(RecordingSampler):
     """Sample sequences of consecutive windows.
 
     Parameters
--- a/braindecode/samplers/base.py
+++ b/braindecode/samplers/base.py
@@ -65,7 +65,7 @@ class RecordingSampler(Sampler):
         # XXX docstring missing
         return self.rng.choice(self.n_recordings)
 
-    def sample_classe(self):
+    def sample_class(self):
         """Return a random recording index.
         """
         # XXX docstring missing
--- a/braindecode/samplers/base.py
+++ b/braindecode/samplers/base.py
@@ -3,6 +3,7 @@ Sampler classes.
 """
 
 # Authors: Hubert Banville <hubert.jbanville@gmail.com>
+#          Theo Gnassounou <>
 #
 # License: BSD (3-clause)
 
@@ -38,14 +39,27 @@ class RecordingSampler(Sampler):
         Series with MultiIndex index which contains the subject, session, run
         and window indices information in an easily accessible structure for
         quick sampling of windows.
+    n_recordings : int
+        Number of recordings available.
     """
-    # XXX attributes n_recordings missing
     def __init__(self, metadata, random_state=None):
         self.metadata = metadata
-        self._init_info()
+        self.info = self._init_info(metadata)
         self.rng = check_random_state(random_state)
 
-    def _init_info(self):
+    def _init_info(self, metadata, required_keys=None):
+        """Initialize ``info`` DataFrame.
+
+        Parameters
+        ----------
+        required_keys : list(str) | None
+            List of additional columns of the metadata DataFrame that we should
+            groupby when creating ``info``.
+
+        Returns
+        -------
+            See class attributes.
+        """
         keys = [k for k in ['subject', 'session', 'run']
                 if k in self.metadata.columns]
         if not keys:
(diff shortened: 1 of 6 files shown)


docs/whats_new.rst
```diff
@@ -33,7 +33,8 @@ Enhancements
 - Adding data :ref:`augmentation_api` module (:gh:`254` by `Cedric Rommel`_, `Alex Gramfort`_ and `Thomas Moreau`_)
 
 - Adding Mixup augmentation :class:`braindecode.augmentation.Mixup` (:gh:`254` by `Simon Brandt`_)
-
+- Adding the USleep sampler :class:`braindecode.samplers` (:gh:`` by `Theo Gnassounou`_)
```
agramfort, line 36:
```suggestion
- Adding the :class:`braindecode.samplers.BalancedSequenceSampler `  sampler proposed in USleep paper (:gh:`` by `Theo Gnassounou`_)
```

docs/api.rst
```diff
@@ -150,6 +150,7 @@ Samplers
    RecordingSampler
    SequenceSampler
    RelativePositioningSampler
+   RandomSampler
```
agramfort, line 153:
```suggestion
   BalancedSequenceSampler
```

braindecode/samplers/base.py
```diff
@@ -123,3 +133,84 @@ def __len__(self):
...
+        """Randomly compute sequence start indice.
+
+        Choose a window associated with a random recording
+        and a random class and place it randomly in a sequence.
+        The function returns the beginning of this sequence.
+
+        Returns
+        -------
+        start_ind : int
+            The indice of the first windows of possible sequences.
```
agramfort, line 180:
```suggestion
            The index of the first window of a possible sequence.
```
or make it plural is you return an array and not an int

braindecode/samplers/base.py
```diff
@@ -123,3 +133,84 @@ def __len__(self):
...
+
+        rec_ind : int
+            The random recording choosen for the indice of the f.
```
agramfort, line 183: indice -> index
when not plural

braindecode/samplers/base.py
```diff
@@ -123,3 +133,84 @@ def __len__(self):
     def __iter__(self):
         for start_ind in self.start_inds:
             yield tuple(range(start_ind, start_ind + self.n_windows))
+
+
+class RandomSampler(RecordingSampler):
```
agramfort, line 138:
```suggestion
class BalancedSequenceSampler(RecordingSampler):
```

braindecode/samplers/base.py
```diff
@@ -123,3 +133,84 @@ def __len__(self):
...
+    """Sample sequences of consecutive windows.
+
+    Parameters
+    ----------
+    metadata : pd.DataFrame
+        See RecordingSampler.
+    n_windows : int
+        Number of consecutive windows in a sequence.
+    seq_nbr : int
```
agramfort, line 147: n_sequences?

to be consistent?

braindecode/samplers/base.py
```diff
@@ -123,3 +133,84 @@ def __len__(self):
...
+        Number of sequences in the sampler.
+    random_state : np.random.RandomState | int | None
+        Random state.
```
agramfort, line 150: add a note section with the ref to USleep paper

agramfort (commented):
I am not sure where the balanced sampling is done.

have you tried this on a real problem / dataset already?

it not I would suggest you do it first to see how well it works. thx @Tgnassou

Tgnassou:
I tried my fonction on dataset. It returns sampling as i want. But i'm not sure that is what we want. 

I choose a random number of subject, I choose a random class (every class has the same probabilitie to be chosen) and I create the sequence with this window. Because all the class have the same probabilitie i thought it was balanced.

agramfort:
I mean does it enable you to get better performance on real data?
does it lead to a running time increase when learning?

basically is it useful according to you and if so is it a free lunch in
running time

>

braindecode/samplers/base.py
```diff
@@ -65,6 +65,12 @@ def sample_recording(self):
         # XXX docstring missing
         return self.rng.choice(self.n_recordings)
 
+    def sample_classe(self):
```
robintibor, line 68: I don't understand this one, what does it do exactly? Why is it called sample_classe? anyways should be either class or classes

braindecode/samplers/base.py
```diff
@@ -65,6 +65,12 @@ def sample_recording(self):
...
+        """Return a random recording index.
```
robintibor, line 69: This docstring not correct

braindecode/samplers/base.py
```diff
@@ -81,6 +87,10 @@ def __iter__(self):
     def n_recordings(self):
         return self.info.shape[0]
 
+    @property
```
robintibor, line 90: Might be actually safer to supply n_classes explicitly? Otherwise you may run into problems:
* if in some subset not all classes are present (imagine small evaluation subset in debugging experiment)
* targets somewhere else as in https://github.com/braindecode/braindecode/pull/261 may not work.

hubertjb:
I added tests and uniformized the implementation so it follows the other samplers' logic more closely. Next step is to test #282 with this sampler and make sure we get better results than with naive sampling of sequences.

braindecode/samplers/base.py
hubertjb: The first point is fixed in the latest commit. I'm not sure about the second point though - the sampler requires categorical targets in the BaseConcatDataset's `metadata` attribute. What use case exactly did you have in mind @robintibor ?

robintibor:
maybe rebase this on https://github.com/braindecode/braindecode/pull/319 then tests should be fine. is there more to do before this is ready on merge @hubertjb ?

hubertjb:
Thanks @robintibor ! The validation of the USleep architecture might lead to some minor changes, but otherwise I think this is ready to be merged.

robintibor:
Great, merged!

PR #303, opened 2021-07-14
[MRG] fix moabb datasets args and docs

gemeinl:
-removed args and kwargs from moabb datasets
-added subject_ids
-added explanation of subject_ids = None in docstrings


robintibor:
thanks

PR #304, opened 2021-07-14
[MRG] Too large windows

gemeinl:


gemeinl:
https://github.com/braindecode/braindecode/blob/150294c755181164595ff730c21a1c5ea9b5f9af/braindecode/preprocessing/windowers.py#L182-L185 were wrong and actually are still wrong. It does not matter if recordings/trials are of different length, unless the user wants to extract a single window per recording/trial. So the check is incomplete

gemeinl:
https://github.com/braindecode/braindecode/blob/150294c755181164595ff730c21a1c5ea9b5f9af/braindecode/preprocessing/windowers.py#L487-L490 is also incorrect. Error message does not match check above.

gemeinl:
@dcwil could you have another look at these checks?

dcwil:
> https://github.com/braindecode/braindecode/blob/150294c755181164595ff730c21a1c5ea9b5f9af/braindecode/preprocessing/windowers.py#L182-L185
> were wrong and actually are still wrong. It does not matter if recordings/trials are of different length, unless the user wants to extract a single window per recording/trial. So the check is incomplete

Should this block also be checking if the user wants to extract a single window per trial, ie `window_stride_samples` and `window_size_samples` are None?

gemeinl:
@dcwil Yes, correct. The warning only has to be raised if user attempts to extract one window per trial (in this case recording) and they are of different lengths. (Stride should not matter at all.) Usually, if window size is specified, or if trials are of the same length, all windows will always be of same length and therefore batch-able.

For the other check I don't know. I just saw that the error message does not match the check. But since the check is in an if else, I don't know whether something else has to be changed.

robintibor:
Great that we caught this bug

PR #310, opened 2021-07-24
[MRG] Remove duplicated info from TUH classes description

gemeinl:
As pointed out in https://github.com/braindecode/braindecode/pull/277#issuecomment-883742290, description contained entries that are basically duplicates from `Raw/Epochs.info`. Hence removed.


robintibor:
oki great always good to consider where information is stored best and if duplications can be avoided! :)

PR #267, opened 2021-06-22
[MRG] Enable multiple discrete targets per recording

gemeinl:
-base datasets now accept a list of strings as target name
-as a result base datasets and window datasets will return multiple values as y in `__getitem__`
-added tests to verify this
-added an example to demonstrate how to get from instantiation of a dataset class to pytorch dataloader

PR #312, opened 2021-07-26
[MRG] Predict trials

gemeinl:
I am not sure I fully understood the issue. Is this what you had in mind @robintibor ?
What would be the best location for `predict_trials`?

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -271,16 +271,23 @@ class EEGClassifier(NeuralNetClassifier):
         """
         return self.predict_proba(X).argmax(1)
 
-    def predict_trials(self, X):
-        """Return trialwise predictions and targets.
+    def predict_trials(self, X, return_targets=True):
+        """Create trialwise predictions (n_trials x n_classes x n_predictions),
+        and optionally also return trialwise labels (n_trials x n_targets).
 
         Parameters
         ----------
-        X : braindecode.datasets.BaseConcatDataset
+        X: braindecode.datasets.BaseConcatDataset
             A braindecode dataset to be predicted.
+        return_targets: bool
+            If True, additionally returns the trial targets.
 
         Returns
         -------
         trial_predictions, trial_labels: tuple(np.ndarray, np.ndarray)
         """
-        return predict_trials(self.module, X)
+        return predict_trials(
+            module=self.module,
+            dataset=X,
+            return_targets=return_targets,
+        )
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -4,6 +4,8 @@
 #
 # License: BSD (3-clause)
 
+import warnings
+
 import numpy as np
 from sklearn.metrics import get_scorer
 from skorch.callbacks import EpochTimer, BatchScoring, PrintLog, EpochScoring
@@ -272,8 +274,8 @@ class EEGClassifier(NeuralNetClassifier):
         return self.predict_proba(X).argmax(1)
 
     def predict_trials(self, X, return_targets=True):
-        """Create trialwise predictions (n_trials x n_classes x n_predictions),
-        and optionally also return trialwise labels (n_trials x n_targets).
+        """Create trialwise predictions and optionally also return trialwise
+        labels from cropped dataset.
 
         Parameters
         ----------
@@ -284,8 +286,24 @@ class EEGClassifier(NeuralNetClassifier):
 
         Returns
         -------
-        trial_predictions, trial_labels: tuple(np.ndarray, np.ndarray)
+        trial_predictions: np.ndarray
+            3-dimensional array (n_trials x n_classes x n_predictions), where
+            the number of predictions depend on the chosen window size and the
+            receptive field of the network.
+        trial_labels: np.ndarray
+            2-dimensional array (n_trials x n_targets) where the number of
+            targets depends on the decoding paradigm and can be either a single
+            value, multiple values, or a sequence.
         """
+        if not self.cropped:
+            warnings.warn(
+                "This method was designed to predict trials in cropped mode. "
+                "Calling it when cropped is False will give the same result as "
+                "'.predict'.", UserWarning)
+            preds = self.predict(X)
+            if return_targets:
+                return preds, X.get_metadata()['target'].to_numpy()
+            return preds
         return predict_trials(
             module=self.module,
             dataset=X,
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -1,5 +1,6 @@
 # Authors: Maciej Sliwowski <maciek.sliwowski@gmail.com>
 #          Robin Schirrmeister <robintibor@gmail.com>
+#          Lukas Gemein <l.gemein@gmail.com>
 #
 # License: BSD (3-clause)
 
@@ -10,7 +11,8 @@ from skorch.classifier import NeuralNet
 from skorch.classifier import NeuralNetClassifier
 from skorch.utils import train_loss_score, valid_loss_score, noop, to_numpy
 
-from .training.scoring import PostEpochTrainScoring, CroppedTrialEpochScoring
+from .training.scoring import (PostEpochTrainScoring, CroppedTrialEpochScoring,
+                               predict_trials)
 from .util import ThrowAwayIndexLoader, update_estimator_docstring
 
 
@@ -268,3 +270,28 @@ class EEGClassifier(NeuralNetClassifier):
 
         """
         return self.predict_proba(X).argmax(1)
+
+    def predict_trials(self, X):
+        """Return trialwise predictions and targets.
+
+        Parameters
+        ----------
+        X : input data, compatible with skorch.dataset.Dataset
+          By default, you should be able to pass:
+
+            * numpy arrays
+            * torch tensors
+            * pandas DataFrame or Series
+            * scipy sparse CSR matrices
+            * a dictionary of the former three
+            * a list/tuple of the former three
+            * a Dataset
+
+          If this doesn't work with your data, you have to pass a
+          ``Dataset`` that can deal with the data.
+
+        Returns
+        -------
+        trial_predictions, trial_labels: tuple(np.ndarray, np.ndarray)
+        """
+        return predict_trials(self.module, X)
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -276,19 +276,8 @@ class EEGClassifier(NeuralNetClassifier):
 
         Parameters
         ----------
-        X : input data, compatible with skorch.dataset.Dataset
-          By default, you should be able to pass:
-
-            * numpy arrays
-            * torch tensors
-            * pandas DataFrame or Series
-            * scipy sparse CSR matrices
-            * a dictionary of the former three
-            * a list/tuple of the former three
-            * a Dataset
-
-          If this doesn't work with your data, you have to pass a
-          ``Dataset`` that can deal with the data.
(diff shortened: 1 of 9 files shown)


braindecode/classifier.py
```diff
@@ -268,3 +270,17 @@ def predict(self, X):
 
         """
         return self.predict_proba(X).argmax(1)
+
+    def predict_trials(self, X):
+        """Return trialwise predictions and targets.
```
robintibor, line 275: Great, can you add one sentence how the expected output shape will be for trial_predictions and trial_labels, including meaning of dimensions for trial_predictions. That should be helpful

braindecode/classifier.py
```diff
@@ -268,3 +270,17 @@ def predict(self, X):
...
+        Parameters
+        ----------
+        X : braindecode.datasets.BaseConcatDataset
+            A braindecode dataset to be predicted.
+
+        Returns
+        -------
+        trial_predictions, trial_labels: tuple(np.ndarray, np.ndarray)
+        """
+        return predict_trials(self.module, X)
```
robintibor, line 286: Maybe also check if self.cropped is True and if not raise error?

braindecode/regressor.py
```diff
@@ -250,3 +252,17 @@ def predict_proba(self, X):
...
+        Parameters
+        ----------
+        X : braindecode.datasets.BaseConcatDataset
+            A braindecode dataset to be predicted.
+
+        Returns
+        -------
+        trial_predictions, trial_labels: tuple(np.ndarray, np.ndarray)
+        """
+        return predict_trials(self.module, X)
```
robintibor, line 268: same here for comments above...  Maybe we could also consider to have a superclass EEGNeuralNet for both to avoid duplication?

robintibor:
Thanks, great, yes that is what I had in mind. I was thinking if function name should somehow indicate that trial labels are being returned as well? Otherwise might be surprising? Or maybe just make return_labels:bool a parameter of the function? to return labels or not? Even with default true, would more explicitly indicate what is happening.

And of course needs tests + whats_new

braindecode/classifier.py
gemeinl: Well, should there be an error? We always have trials no matter if we do cropped decoding or not.

gemeinl:
Yes it is surprising but what you requested.
How about we rename to `get_trial_preds_and_labels`?
Or do you like the `return_labels` flag better?

braindecode/regressor.py
gemeinl: Yes, definitely. I just added one more duplicate to the large list of code duplicates in those classes...

robintibor:
I think return_labels flag is better, predict_trials is quite nice name

docs/whats_new.rst
```diff
@@ -30,6 +30,7 @@ Enhancements
 - Adding Mixup augmentation :class:`braindecode.augmentation.Mixup` (:gh:`254` by `Simon Brandt`_)
 - Adding saving of preprocessing and windowing choices in :func:`braindecode.preprocessing.preprocess`, :func:`braindecode.preprocessing.create_windows_from_events` and :func:`braindecode.preprocessing.create_fixed_length_windows` to datasets to facilitate reproducibility (:gh:`287` by `Lukas Gemein`_)
 - Adding :func:`braindecode.models.util.aggregate_probas` to perform self-ensembling of predictions with sequence-to-sequence models (:gh:`294` by `Hubert Banville`_)
+- Adding :func:`braindecode.training.scoring.predict_trials` to generate trialwise predictions (:gh:`312` by `Lukas Gemein`_)
```
robintibor, line 33:
```suggestion
- Adding :func:`braindecode.training.scoring.predict_trials` to generate trialwise predictions after cropped training (:gh:`312` by `Lukas Gemein`_)
```

To make even clearer

braindecode/classifier.py
robintibor: I think 
a) Right now code would fail without cropped decoding? There are a lot of assumptions inside coming from cropped decoding no? Or would it just run?
b) for trialwise decoding existing predict method would already give you trialwise predictions right? 

If that is correct either we assert we are in cropped mode (and maybe also add to name like predict_trials_from_cropped) or we just call regular predict function if not in cropped mode? We should check what happens atm when this function is called e.g. at end of trialwise decoding example

gemeinl: I am not so much familiar with the internals of the cropped decoding to know about the assumptions.
In my opinion `predict_trials` should always work, no matter if cropped or trialwise decoding, since we always have trials.
For trialwise decoding the output will then just be the same as calling `predict`. For cropped decoding it will be different.

braindecode/regressor.py
```diff
@@ -250,3 +252,24 @@ def predict_proba(self, X):
...
+        Parameters
+        ----------
+        X: braindecode.datasets.BaseConcatDataset
+            A braindecode dataset to be predicted.
+        return_targets: bool
+            If True, additionally returns the trial targets.
+
+        Returns
+        -------
+        trial_predictions, trial_labels: tuple(np.ndarray, np.ndarray)
```
agramfort, line 269: please specify the dimension of each array. This syntax `tuple(np.ndarray, np.ndarray)` is odd to me.

braindecode/training/scoring.py
```diff
@@ -295,3 +297,47 @@ def on_epoch_end(self, net, dataset_train, dataset_valid, **kwargs):
                 cached_net, dataset_train, self.y_trues_
             )
         self._record_score(net.history, current_score)
+
+
+def predict_trials(module, dataset, return_targets=True):
```
agramfort, line 302: why adding 2 ways of doing the same thing ie the method and this public function?

braindecode/training/scoring.py
```diff
@@ -295,3 +297,47 @@ def on_epoch_end(self, net, dataset_train, dataset_valid, **kwargs):
...
+    """Create trialswise predictions (n_trials x n_classes x n_predictions),
+    and optionally also return trialwise labels (n_trials x n_targets) from
```
agramfort, line 304: n_targets and n_predictions are conceptually different things? sorry but I get confused by n_trials, something I I would call n_crops etc.

gemeinl: `n_predictions` is in time domain and depends on `window_size` and size of the receptive field of the net.
`n_targets` can be very different. It can be a single value in classification / regression tasks, it can be multiple values in multiple discrete target classification as introduced with https://github.com/braindecode/braindecode/pull/267, and it can also be a sequence as in https://github.com/braindecode/braindecode/pull/261.

Why do you get confused by `n_trials`? That is the point, we do not have crops / compute windows at this point. These are actual trials. If you want to generate crop / compute window predictions you would call `.predict()` instead. In `predict_trials` we invert the creation of compute windows (removing potentially overlapping predictions) to obtain trial predictions.

braindecode/training/scoring.py
gemeinl: I guess `braindecode.training.scoring.predict_trials` exists because you don't need an EEGClassifier / EEGRegressor to make predictions. It is sufficient to have any kind of model given as PyTorch module as well as a braindecode dataset. 

However, you would expect your estimator to provide `predict..` right?

docs/whats_new.rst
gemeinl: I sitll disagree. For me this is not specific for cropped decoding. See comment above.

braindecode/regressor.py
gemeinl: Dimension are at the top of the docstring https://github.com/braindecode/braindecode/pull/312/files/0c3e6db742d0012eac23e35ae1ba47f74dda7c88#diff-049a3b69cc29c2439db003260fe1ca5f9908afd5f4de2498c5ff7d155cf96f60R257-R258. Should I move it to the Returns section or add it here again?

braindecode/training/scoring.py
agramfort: ok much clearer 🙏 . Maybe it's worth adding a glossary as we do with MNE https://mne.tools/stable/glossary.html ?

braindecode/regressor.py
agramfort: yes. I won't expect dimensions in the header line of a docstring. It should be in the Returns statement.

you should skim through https://www.python.org/dev/peps/pep-0257/

we should activate https://github.com/PyCQA/pydocstyle on the repo....

gemeinl: I see, thank you. I support the idea for more automated checks to ensure we follow conventions.

gemeinl:
`test_variable_length_trials_decoding` fails?

robintibor:
Worked on Rerun, so probably just need to increase tolerance even further

docs/whats_new.rst
gemeinl: @robintibor In `braindecode.training.scoring.predict_trials` we cannot know whether the model was trained in cropped fashion. We need an `EEGClassifier` / `EEGRegressor` for this. Is it save to add this function anyways? Or should we remove it and only have `EEGClassifier/EEGRegressor.predict_trials()` which will then check `self.cropped`?

gemeinl:
Done from my side unless @robintibor disagrees with current implementation regarding cropped / trialwise stuff...

braindecode/training/scoring.py
```diff
@@ -295,3 +297,59 @@ def on_epoch_end(self, net, dataset_train, dataset_valid, **kwargs):
...
+            2-dimensional array (n_trials x n_targets) where the number of
+            targets depends on the decoding paradigm and can be either a single
+            value, multiple values, or a sequence.
+    """
+    # we have a cropped dataset if there exists at least one trial with more
+    # than one compute window
+    cropped_data = sum(dataset.get_metadata()['i_window_in_trial'] != 0) > 0
+    if not cropped_data:
+        raise ValueError('This function was designed to predict trials from '
+                         'cropped datasets. This is a trialwise dataset.')
```
robintibor, line 331:
```suggestion
    more_than_one_window = sum(dataset.get_metadata()['i_window_in_trial'] != 0) > 0
    if not more_than_one_window:
        warnings.warn('This function was designed to predict trials from '
            'cropped datasets, which typically have multiple compute windows per trial .'
            'The given dataset has exactly one window per trial,')
```
Is it sure that this must be the case? I think this is not necessarily true. Maybe at least downgrade to a warning instead of an error?

robintibor:
Made one comment, what do you think @gemeinl ?

braindecode/training/scoring.py
gemeinl: I could not come up with a case where it does not hold. Isn't it the definition of trialwise decoding to have one window per trial?

docs/whats_new.rst
gemeinl: What about this here @robintibor

braindecode/training/scoring.py
robintibor: one can consider (and at least that's what I discussed with Tonio as well as far as I recall) trialwise decoding implies single window and *single prediction* per trial, whereas if you have single window but multiple predictions, than you can still consider it cropped decoding. and all the code should also run fine

docs/whats_new.rst
robintibor: I think we should have it so people can use this if they are not using skorch.

braindecode/training/scoring.py
gemeinl: Ok. Then I will update according to your suggestions

robintibor:
Great stuff, merged now.

PR #318, opened 2021-08-06
[MRG] TimeDistributed class to apply a feature extractor model on a sequence of input windows

hubertjb:
This PR adds the `TimeDistributed` class that was defined in the sleep staging example to `braindecode.models.util` so that it can be reused elsewhere.


gemeinl:
`test_variable_length_trials_decoding.py` sometimes randomly fails. I do not know why yet. I restarted the job and all tests passed.

robintibor:
> `test_variable_length_trials_decoding.py` sometimes randomly fails. I do not know why yet. I restarted the job and all tests passed.

Here also, rebasing now should prevent test from failing hopefully

hubertjb:
> Great idea, let's think about naming a bit. I would suggest to move to own file, not in util.
> For class name, how about `PerWindowExtractor` or something similar? `TimeDistributed` seems a little unclear to me what that could be @hubertjb

Do you think it would make more sense to move it under `braindecode.models.modules`?

I named it `TimeDistributed` because it's the term used in the [paper](https://arxiv.org/abs/1707.03321) the sleep staging example is based on, and also because it's conceptually similar to the `TimeDistributed` layer in keras: https://keras.io/api/layers/recurrent_layers/time_distributed/ Since it doesn't do the exact same thing I guess it might be better to use a different name as you suggest though. 

I'm not sure about `PerWindowExtractor` because it does more than just apply a feature extractor per window. What do you think of something like `SequenceAggregator` or `SequenceClassifier`?

braindecode/models/util.py
```diff
@@ -136,3 +138,60 @@ def aggregate_probas(logits, n_windows_stride=1):
...
+        Dropout to be applied before the linear layer. Ignored if ``clf`` is
+        provided.
+    clf : nn.Module | None
+        If provided, module that will receive the concatenated features and
+        produce a prediction. If None, a simple linear layer with dropout
+        is used, as defined using parameters ``feat_size``, ``n_windows``,
+        ``n_classes`` and ``dropout``.
+    """
+    def __init__(self, feat_extractor, feat_size=None, n_windows=None,
+                 n_classes=None, dropout=0.25, clf=None):
```
robintibor, line 176: How about instead if `clf=None`, you really just return features? this would also remove immediately three arguments and make the class substantially simpler...

robintibor:
Yeah `braindecode.models.modules` would make a lot more sense I think, if not even own fike, but modules could be fine. Didn't know about the keras class, but seems this one then is also different to the keras class, as the keras class really just applies the feature extractor (there called layer) right?

hubertjb:
Maybe then it would make more sense to do something like in keras, i.e. just applying the feature extractor on multiple windows, and returning the features, and call this `TimeDistributed`? I'm thinking I could then move some of the logic that is specific to the sleep staging example inside `braindecode.models.sleep_stager_chambon_2018.py`.

robintibor:
yeah sounds reasonable @hubertjb

hubertjb:
@robintibor I simplified the implementation, now it only does the feature extraction part.

robintibor (approved):

robintibor:
great makes sense to me now, merged

PR #321, opened 2021-08-11
[WIP] Runtime improvements

gemeinl:
Reduced runtime of `TUHAbonormal` from 2 min on master to 44s.
Reduced runtime of `TUH` from 58 min on master to 10 min.

master:
image
this PR:
image


agramfort:
can you also compare to torch.save ? I think it's what most pytorch
users would expect. And the .pt extension is clear of how to load such
files.

PR #326, opened 2021-08-30
Fix is_outdated_saved with large files split by mne

### LINKED ISSUE #324: `is_outdated_saved` recognize new dataset as outdated
When trying to load a dataset saved with new `save` method it is considered as `outdated_dataset`. The wrong behavior comes from this line:
https://github.com/braindecode/braindecode/blob/2dbd40cc7951934ad32892770536e360b6188920/braindecode/datautil/serialization.py#L220 which when True results in considering the dataset as outdated. This may not be the case when saved raw or epochs objects were too big and split into multiple files which results in a dataset with multiple .fif files in each folder:
```
path/
    0/
        0-raw.fif | 0-epo.fif
        0-raw-1.fif | 0-epo-1.fif and so on
        description.json
        raw_preproc_kwargs.json (if raws were preprocessed)
        window_kwargs.json (if this is a windowed dataset)
        window_preproc_kwargs.json  (if windows were preprocessed)
        target_name.json (if target_name is not None and dataset is raw)
    1/
        1-raw.fif | 1-epo.fif
        1-raw-1.fif | 1-epo-1.fif and so on
        description.json
        raw_preproc_kwargs.json (if raws were preprocessed)
        window_kwargs.json (if this is a windowed dataset)
        window_preproc_kwargs.json  (if windows were preprocessed)
        target_name.json (if target_name is not None and dataset is raw)
```
The idea to fix that would be to remove the condition on different number of description and fif_files. However I am not sure if it can be done because I don't know what was the previous way of saving (I tried to reproduce it but `_outdated_save` is not working for me anymore). Can anyone confirm that removing won't break the function or propose a new condition on multiple description files per fif file taking into account outdated dataset structure?

sliwy:
In some cases `_is_outdated_saved` incorrectly recognizes new datasets as saved in the deprecated way. Using `force_new_loading` users can enforce loading with new load function. It should be used  if user is sure about the dataset that it's new.

closes #324

### DIFF
--- a/braindecode/datautil/serialization.py
+++ b/braindecode/datautil/serialization.py
@@ -216,7 +216,7 @@ def _is_outdated_saved(path):
     or '-epo.fif' in path (no subdirectories) OR if there are more 'fif' files
     than 'description.json' files."""
     description_files = glob(os.path.join(path, '**/description.json'))
-    fif_files = glob(os.path.join(path, '**/*.fif'))
+    fif_files = glob(os.path.join(path, '**/*-raw.fif')) + glob(os.path.join(path, '**/*-epo.fif'))
     multiple = len(description_files) != len(fif_files)
     kwargs_in_path = any(
         [os.path.exists(os.path.join(path, kwarg_name))


agramfort:
can we avoid complexifying the public API for this that would be great.

sliwy:
so maybe let's have an environment variable e.g. `BRAINDECODE_ONLY_NEW_DATASETS` that can be set by user that needs that?

sliwy:
Finally there is an easier option to fix that. Now we count only files that were split by the old saving function and we ignore the files split by mne.

agramfort (approved):
much simpler :)

gemeinl:
Great stuff! Thanks a lot @sliwy !

robintibor:
Amazing stuff, great fix. wanna press a merge button  @gemeinl ? :D

Issue #325, opened 2021-08-30
`preprocess` fails for `n_jobs`>1


dcwil:
Running the latest master I run into issues with the parallelised preprocessing:

```
from braindecode.preprocessing import Preprocessor, preprocess, filterbank
from braindecode.datasets.moabb import MOABBDataset

ds = MOABBDataset(dataset_name='Schirrmeister2017',
                           subject_ids=[1], )
splits = ds.split([[i] for i in range(len(ds.datasets))])

fbank = [
            Preprocessor(
                apply_on_array=False,
                fn=filterbank,
                frequency_bands=[(0, 10)],
                verbose=False,
                n_jobs=1
            ),
        ]

for split_i, split_ds in splits.items():
    preprocess(concat_ds=split_ds, preprocessors=fbank, n_jobs=2)
```
gives the following output:

```
Creating RawArray with float64 data, n_channels=128, n_times=1225545
    Range : 0 ... 1225544 =      0.000 ...  2451.088 secs
Ready.
Creating RawArray with float64 data, n_channels=128, n_times=616535
    Range : 0 ... 616534 =      0.000 ...  1233.068 secs
Ready.
320 events found
Event IDs: [1 2 3 4]
160 events found
Event IDs: [1 2 3 4]
joblib.externals.loky.process_executor._RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/home/dan/Documents/AGBALL/RiemNetEnv/lib/python3.8/site-packages/joblib/externals/loky/process_executor.py", line 431, in _process_worker
    r = call_item()
  File "/home/dan/Documents/AGBALL/RiemNetEnv/lib/python3.8/site-packages/joblib/externals/loky/process_executor.py", line 285, in __call__
    return self.fn(*self.args, **self.kwargs)
  File "/home/dan/Documents/AGBALL/RiemNetEnv/lib/python3.8/site-packages/joblib/_parallel_backends.py", line 595, in __call__
    return self.func(*args, **kwargs)
  File "/home/dan/Documents/AGBALL/RiemNetEnv/lib/python3.8/site-packages/joblib/parallel.py", line 262, in __call__
    return [func(*args, **kwargs)
  File "/home/dan/Documents/AGBALL/RiemNetEnv/lib/python3.8/site-packages/joblib/parallel.py", line 262, in <listcomp>
    return [func(*args, **kwargs)
[...]

PR #327, opened 2021-09-06
[MRG] Fixing serialization of SleepPhysionet

hubertjb:
This PR fixes two minor bugs with the SleepPhysionet class that I came across with while running some tests for #325. First, the `exclude` attribute of the Raw object was a `dict_keys` object, which prevented the dataset from being serialized. Second, channels were renamed using `mne.rename_channels` instead of using Raw's `rename_channels` method, which led to `.ch_names` and `._orig_units` having different channel names. This created an issue when applying the filterbank preprocessor.


robintibor (approved):

robintibor:
great

PR #328, opened 2021-09-17
[MRG] Using recording date from file path instead of EDF header for TUH datasets

hubertjb:
This PR modifies the `TUH` dataset class to update the `raw.info['meas_date']` property with the recording date specified in the file path, instead of the one from the EDF header. This is helpful because some files seem to have corrupted recording dates. 

For instance, for TUAB, there are 79 mismatches between the EDF header date and the path-defined date, and 16 of those are visibly default values/errors (many start with 2000-01-01, or even 1899).

### DIFF
--- a/braindecode/datasets/tuh.py
+++ b/braindecode/datasets/tuh.py
@@ -11,6 +11,7 @@ import re
 import os
 import glob
 from unittest import mock
+from datetime import datetime, timezone
 
 import pandas as pd
 import numpy as np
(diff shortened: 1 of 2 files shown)


agramfort (approved):
@gemeinl @robintibor merge if happy.

thx @hubertjb

robintibor (approved):

gemeinl:
Great, thanks for fixing @hubertjb

PR #339, opened 2021-10-19
Fix rotation transform in case of learnable param

cedricrommel:
This PR will just catch correctly an error that occurs due to pytorch, within the SensorsRotation transform.
Indeed, when any of its parameters is a nn.Parameter, the functional would crash because, for some reason, you cannot pass a boolean mask made only of `True` to tensors that have `requires_grad = True` 🤷‍♂️ 

I've just added a commit where I fix how devices are handled. Indeed, I've realize by running the tests on `cuda` that I had broken device compatibility with my last PR. This is included in the tests (I was just running them locally and head no cuda available 😬 )

### DIFF
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -582,7 +582,6 @@ def frequency_shift(X, y, delta_freq, sfreq):
 
 def _torch_normalize_vectors(rr):
     """Normalize surface vertices."""
-    new_rr = rr.clone()
     norm = torch.linalg.norm(rr, axis=1, keepdim=True)
     mask = (norm > 0)
     norm[~mask] = 1  # in case norm is zero, divide by 1


robintibor:
Hi great stuff to catch bugs, but let's try to fully understand what is going on, also to prevent that there is maybe another undetected error. Maybe you can print the sizes of new_rr, rr, mask, size etc. for the different cases to make sure what is going on. Could also post it here... this error should really be understandable :)

cedricrommel:
I've also added a new commit where I fix how devices are handled (which had been broken in my last PR, as explained in the updated PR description).

agramfort (approved):
+1 for MRG if CIs are green.

thx @cedricrommel !

agramfort (approved):

agramfort:
@robintibor feel free to MRG if happy

robintibor:
So I had a bit more look inside as I still think we should have some code where we feel confident it works, not have some error solved while we don't understand what happened :)

I see at least three ways, 
1):

```python
new_rr = rr.clone()
size = torch.linalg.norm(rr, axis=1)
mask = (size > 0)
inds = mask.nonzero().flatten()
new_rr[mask] = rr.index_select(0, inds) / size.index_select(0, inds).unsqueeze(-1)
```

2) Simpler if you just want to divide by norm and prevent division by zero, you could clamp min to low value:
```python
new_rr = rr / torch.linalg.norm(rr, axis=1, keepdim=True).clamp_min(1e-12)
```
Might be also equivalent to:

```python
new_rr = torch.nn.functional.normalize(rr, eps=1e-12)
```

3) If you want to be sure only those with zero norm don't get renormalized:
```python
norm = torch.linalg.norm(rr, axis=1, keepdim=True)
mask = (norm > 0)
norm[~mask] = 1 # in case norm is zero, divide by 1
new_rr = rr / norm
```

Can you try these @cedricrommel and tell me which one you prefer?

agramfort (approved):

PR #340, opened 2021-10-19
[WIP] Fix violations of scikit-learn conventions

gemeinl:
- trying to avoid violating scikit-learn conventions

### DIFF
--- a/braindecode/regressor.py
+++ b/braindecode/regressor.py
@@ -162,7 +162,7 @@ class EEGRegressor(NeuralNetRegressor):
             epoch_cbs = []
             for name, cb in cbs:
                 if isinstance(cb, (CroppedTrialEpochScoring, CroppedTimeSeriesEpochScoring)) and (
-                        hasattr(cb, 'window_inds_')) and (not cb.on_train):
+                        hasattr(cb, '_window_inds_')) and (not cb.on_train):
                     epoch_cbs.append(cb)
             # for trialwise decoding stuffs it might also be we don't have
             # cropped loader, so no indices there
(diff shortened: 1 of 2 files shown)


sliwy:
@gemeinl sorry for this probably noob question but what is required by scikit-learn regarding the callbacks? I couldn't find a place where callbacks (or something similar) were used in scikit-learn. Did you mean skorch and its scikit-learn API?

For me it was quite convenient to have it automatically parsed inside the model class for standard scikit-learn metrics. The new way is not much worse but I feel that `parse_callbacks` fits quite well in the model class.

gemeinl:
@agramfort do you see a way of adding custom callbacks in the init of EEGRegressor while at the same time following scikit-learn conventions?

agramfort (approved):
diff looks great !

don't know about CI failure. But no objection to merge once CIs are green

thx heaps @gemeinl !

test/unit_tests/test_eeg_classifier.py
```diff
@@ -119,3 +120,23 @@ def test_predict_trials():
...
+def test_eeg_classifier_clonable():
+    preds = np.array(
+        [
+            [[0.2, 0.1, 0.1, 0.1], [0.8, 0.9, 0.9, 0.9]],
+            [[1.0, 1.0, 1.0, 1.0], [0.0, 0.0, 0.0, 0.0]],
+            [[1.0, 1.0, 1.0, 0.2], [0.0, 0.0, 0.0, 0.8]],
+            [[0.9, 0.8, 0.9, 1.0], [0.1, 0.2, 0.1, 0.0]],
+        ]
+    )
+    clf = EEGClassifier(
```
sliwy, line 134: Shouldn't we put here some callbacks as parameter to check whether we can clone something with callbacks? Here callbacks is None, so nothing will happen in the `_uniquely_named_callbacks`?

agramfort:
you should not overwrite an init param in the fit to follow sklearn
guidelines

>

PR #341, opened 2021-10-19
Adding Sleep Models

### LINKED ISSUE #334: Adding the AttnSleep Model and another CNN model for Sleep Scoring
I recently made use of 2 models (AttnSleep and another implemented from a paper) for sleep scoring [available here](https://github.com/Div12345/SleepStaging-TransferLearning), it should be possible to add them to Braindecode's model library. If it's of interest I could make an attempt at it and add them to the sleep staging [example](https://braindecode.org/auto_examples/plot_sleep_staging.html) already present in the docs.

Div12345:
Closes #334 

Added the AttnSleep Model and have the example tried once [here](https://colab.research.google.com/gist/Div12345/daf5664be45b8e680257481b53cf586b/testing-braindecode-attnsleep.ipynb). AttnSleep requires Single-Channel EEG Data. So I used a new windows_dataset with a picked channel. I wasn't sure whether to update the existing example because then it probably won't look good with 2 different datasets creations?


braindecode/models/attn_sleep.py
```diff
@@ -0,0 +1,395 @@
...
+
+        x, self.attn = attention(query, key, value, dropout=self.dropout)
+
+        x = x.transpose(1, 2).contiguous() \
+            .view(nbatches, -1, self.h * self.d_k)
+
+        return self.linear(x)
+
+
+class LayerNorm(nn.Module):
```
agramfort, line 310: why not using https://pytorch.org/docs/stable/generated/torch.nn.LayerNorm.html ?

braindecode/models/attn_sleep.py
```diff
@@ -0,0 +1,395 @@
...
+        super(SublayerOutput, self).__init__()
+        self.norm = LayerNorm(size)
+        self.dropout = nn.Dropout(dropout)
+
+    def forward(self, x, sublayer):
+        """Apply residual connection to any sublayer with the same size."""
+        return x + self.dropout(sublayer(self.norm(x)))
+
+
+def clones(module, N):
```
agramfort, line 340: private?

agramfort (commented):
@Div12345 you'll need to add test and update the documentation

ask if you need help

Div12345:
Regarding the docs, what all should be updated? From the USleep PR I see api.rst and whats_new.rst updated. Should these just be updated manually?

braindecode/models/attn_sleep.py
```diff
@@ -0,0 +1,373 @@
...
+from torch import nn
+import torch.nn.functional as F
+import numpy as np
+import math
+import copy
+from copy import deepcopy
+
+
+class AttnSleep(nn.Module):
+    """Sleep Staging Architecture from Eldele et al 2021.
```
agramfort, line 15: you add 2 publics models here no? 
to be consistent this could be:

SleepStagerEldele2021

?

agramfort (commented):
search for SleepStagerChambon2018 in the code base to see where the new models should be mentionned. Just add the new model names below

sliwy:
The problem then is the length (in time) of the window only? If we have `sfreq` dependent filters/pooling sizes the model can take any number of samples-however it has to correspond to 30 seconds of the signal? 

Actually I don't know whether setting windows to be always 30 seconds is a standard thing for sleep decoding and how much it can be a limiting factor in the future. 

@agramfort should we look for something working for different window lengths or can we just leave it as is with some warning message? I would go for the second option as it's easier and this is a kind of model property.

agramfort:
I think it's ok to suppose it's 30s at 100Hz so each window is 3000 samples
and raise a ValueError is it's not the case

robintibor:
yeah I think it is definitely good if by default users get a model the way it was developed/evaluated by original authors, to prevent that people use some suboptimal model and don't realize it. Whether one does this through issuing warnings in case the model is different or raises errors, I don't have a very strong opinion. But if it is the easiest solution to restrict to the paper-setting and raise errors otherwise, it is fine to me.

Div12345:
So in the paper, they've also tested with the SHHS dataset in which the EEG is epoched for 30sec at 125Hz. For this, they've used the d_model parameter as 100. I thought to conform to the paper, I could just add the information regarding this and just add that configuration to the test too.

So I tried testing for it and found an error that the output at the end of the MRCNN is 99 and not 100 for that config. d_model can also not be set to 99 since it isn't divisible by n_attn_heads. 
It does work however with 30-sec windows,  sfreq 126Hz, d_model 100.

Any suggestions? Should I just leave it out and just give the 100Hz, 30sec?

And as per the discussion above, I've added a warning that checks if the window size is not 30sec or sfreq is not 100Hz or d_model is not 80.

agramfort:
>
> So in the paper, they've also tested with the SHHS dataset in which the
> EEG is epoched for 30sec at 125Hz. For this, they've used the d_model
> parameter as 100. I thought to conform to the paper, I could just add the
> information regarding this and just add that configuration to the test too.
>
makes sense

> So I tried testing for it and found an error that the output at the end of
> the MRCNN is 99 and not 100 for that config. d_model can also not be set to
> 99 since it isn't divisible by n_attn_heads.
> It does work however with 30-sec windows, sfreq 126Hz, d_model 100.
>
> Any suggestions? Should I just leave it out and just give the 100Hz, 30sec?
>

if there is a bug in the paper contact the authors. In the mean time just
add support for 100Hz datasets.


> And as per the discussion above, I've added a warning that checks if the
> window size is not 30sec or sfreq is not 100Hz or d_model is not 80.
>

ok !

braindecode/models/sleep_stager_blanco_2020.py
```diff
@@ -0,0 +1,125 @@
...
+        self.feature_extractor.eval()
+        with torch.no_grad():
+            out = self.feature_extractor(
+                torch.Tensor(1, n_channels, 1, input_size))  # batch_size,n_channels,height,width
+        self.feature_extractor.train()
+        return len(out.flatten())
+
+    def forward(self, x):
+        """Forward pass.
+
```
robintibor, line 112:
```suggestion

```

robintibor:
have quickly fixed a whitespace flake8 error to see if CIs succeed then @Div12345

robintibor (approved):

robintibor:
Seems ready to merge for me, is it ready to merge from your side @Div12345 ?

Div12345:
Do you want me to add an example or add these to the existing examples?

robintibor:
ah good question. In general, I would like it that the example shows how to use the model in a state-of-the-art way. So on a dataset it performs well on (if publicly available and downloadable) and with good hyperparameters. And then only make it very fast for CI by reducing number of epochs and possibly selecting a subset of data. In our experience, users seeing an example will often assume that taking the code as is will give great performance, so we should have that in mind. With that in mind, what would be the best example from your point of view?

sliwy (approved):
The only question is if we can have an example script that uses this model?

Great job @Div12345!

Div12345:
Regarding the example - 
I shared the learning curve as a part of the existing example here - 

> 
> 
> The learning curve is available [here](https://colab.research.google.com/gist/Div12345/daf5664be45b8e680257481b53cf586b/testing-braindecode-attnsleep.ipynb#scrollTo=a07KxdkwAKTk&line=113&uniqifier=1) 
> Eldele2021 - image
> 
> Blanco2020 - image
> 
> The Eldele2021 learns fairly well whereas the Blanco2020 is pretty poor for this case.


In this I showed the decent performance of the Blanco2020 -
> 
> 
> In my use case [here](https://github.com/Div12345/SleepStaging-TransferLearning/blob/main/Blanco2020_Model_Sleep_TL.ipynb) it shows decent learning. But I understand, let me know if you want me to make any further specific tests to see if it's worth it or otherwise I can remove it.

sliwy:
I thought about some example similar to the ones you can find in the `examples/` in the braindecode. I've seen your learning curves and that's ok but that's the example in your repo, so not many people will spot it.

I don't feel like it's mandatory to have an example for every model but from what I see, people mainly use the models that have an example script showing how to use it (that's just my feeling after seeing some issues in the repo). Anyway, if we don't have an example ready, I would merge this. Maybe this problem can be solved in the future with benchmark script.

braindecode/models/sleep_stager_eldele_2021.py
```diff
@@ -167,7 +173,7 @@ def forward(self, x):
 
 
 class _MRCNN(nn.Module):
-    def __init__(self, after_reduced_cnn_size):
+    def __init__(self, after_reduced_cnn_size, shhs):
```
agramfort, line 176: I would not hard code anything for a specific dataset. Just check the value of sfreq? ok for you?

Div12345:
Hard coded in terms of the input arg for the private class? Meaning better
to check for the specific (sfreq, input_size, d_model) inside the mrcnn
rather than in the primary class? Or hard coded in terms of this change at
all in the kernel size for accomodating this use case for the specific
dataset?

agramfort:
>
> Hard coded in terms of the input arg for the private class? Meaning better
> to check for the specific (sfreq, input_size, d_model) inside the mrcnn
> rather than in the primary class? Or hard coded in terms of this change at
> all in the kernel size for accomodating this use case for the specific
> dataset?


do the checks as early as possible or just before it becomes a requirement.

what is specific so shhs? just sfreq? the network will break if you change
what?
sfreq? n_channels?

Div12345:
The network breaks due to that particular number of input samples (125Hz*30s) gives the mrcnn out as 99 rather than the 100 which the paper uses. As the first author said in the issue [here](https://github.com/emadeldeen24/AttnSleep/issues/6) - 

> 
>
> I guess it is a sample length issue that leads to kernel size and padding issues in MRCNN. That's why I included MRCNN_SHHS to overcome this issue in SHHS, where I just modified the kernel size of the last Conv1d layer. You should use that and modify d_model to 100.

agramfort:
ok so any dataset with sfreq=125 will have the same pb. It's not SHHS
specific.
that's why I suggest to check the sfreq and not the specific dataset;

makes sense?

>

Div12345:
@agramfort I do not check for the specific dataset as such. I do the check as per the input params [here](https://github.com/braindecode/braindecode/pull/341/files#diff-4e03e96560b897b4bf4f6353fff19ca630f66347eafd96f741a62d3da843537fR78). I just used the bool name to be shhs.. If that doesn't seem standard, I'll change it to something else.

agramfort:
yes I saw but you still mention shhs in the code (not just in a comment)
and you check a lot
more than sfreq == 125.

We should not hard code anything for one dataset.

my 2c

>

braindecode/models/sleep_stager_eldele_2021.py
```diff
@@ -0,0 +1,408 @@
...
+                nn.Conv1d(64, 128, kernel_size=6, stride=1, bias=False, padding=3),
+                nn.BatchNorm1d(128),
+                self.GELU,
+
+                nn.Conv1d(128, 128, kernel_size=6, stride=1, bias=False, padding=3),
+                nn.BatchNorm1d(128),
+                self.GELU,
+
+                nn.MaxPool1d(kernel_size=2, stride=2, padding=1)
+            )
```
robintibor, line 232: in any case here it does not become so clear what is the difference, most of the code is the same right? if it is only kernel sizes 6 instead of 7 you could just set these as variables before. Right now really hard to tell what is the difference

robintibor:
yeah @agramfort but then what do if orig author created separate version for a specific dataset as per https://github.com/emadeldeen24/AttnSleep/issues/6 ? I mean maybe we could solve more general case, checking if correct output size 100 is there and if not do the appropriate modification? Would this be easy or complicated @Div12345 ?

agramfort:
I would assume that if calling the model on the data does not crash it
should be allowed. We should advertise
one model that only works with one dataset. It's for me the whole point to
be able to more easily benchmark
the different architectures on different datasets.

>

Div12345:
> yes I saw but you still mention shhs in the code (not just in a comment) and you check a lot
more than sfreq == 125.

@agramfort I check for sfreq, input_size_s and d_model. The first 2 since together those cause the problem  and d_model for giving the warning to the user about violating the original design on this [Line](https://github.com/braindecode/braindecode/pull/341/files#diff-4e03e96560b897b4bf4f6353fff19ca630f66347eafd96f741a62d3da843537fR81)
I could also just check through input_size=sfreq*input_size_s - that would also make the adaptation slightly more generic, but a different sfreq but overall same number of samples, still mean that the mrcnn likely won't perform as intended to capture the intended frequency bands. I multipurpose the check to give the warning and call for the adaptation in the mrcnn sub-module. It may make sense to separate that out and retain the advantage of not causing the model to break whenever the number of samples is that much. I leave that to your call.

> I mean maybe we could solve more general case, checking if correct output size 100 is there and if not do the appropriate modification?

@robintibor If you mean checking the size at the end of the mrcnn and checking if that is equal to d_model, sure. I did that in one of the previous tries to solve the problem, so that can be added back easily. 

In more general terms, that would be checking for the failure in the architecture. Considering that we are not going to specifically make any modifications to the architecture though, it seems to me that checking the input and giving the warning for the intended design may be better than doing a check for the breakage and then making the adaptation just for the use case that d_model is 100 (since that is only adaptation we care considering that's already part of what the author shares with us). The way as it is now, if the input is "wrong", the model would anyway error out in the initialization but at least the user will have an idea about why that happens.

>I would assume that if calling the model on the data does not crash it should be allowed. 

@agramfort I agree, and that is why I think the pre-checking and simply giving a warning as it is now, is allowing that. 

> We should advertise one model that only works with one dataset.

I don't specifically understand the not hard-coding for one dataset part as you say, since we are just doing that hard coding for the extension as per the paper. On the whole, it can be said that this model is designed to work with 2 datasets and it would also be a "loss" for that dataset I guess even towards the benchmarking if we purposefully leave out the adaptation (especially since it is very minor)

robintibor:
> > I mean maybe we could solve more general case, checking if correct output size 100 is there and if not do the appropriate modification?
> 
> @robintibor If you mean checking the size at the end of the mrcnn and checking if that is equal to d_model, sure. I did that in one of the previous tries to solve the problem, so that can be added back easily.
> 
> In more general terms, that would be checking for the failure in the architecture. Considering that we are not going to specifically make any modifications to the architecture though, it seems to me that checking the input and giving the warning for the intended design may be better than doing a check for the breakage and then making the adaptation just for the use case that d_model is 100 (since that is only adaptation we care considering that's already part of what the author shares with us). The way as it is now, if the input is "wrong", the model would anyway error out in the initialization but at least the user will have an idea about why that happens.
> 

Could we do both, so checking in the beginning whether the input arguments are in the range that the authors have used and give a warning otherwise, later catch the breakage to be generic? maybe makes sense? would it be possible?

Div12345:
Sure, I can do that. Just to confirm, you mean catching the breakage whenever the output of the mrcnn is not equal to d_model, and giving an assertion error right? Do you want me to do the shhs adaptation on the basis of the initial param check or by making a subcase in this assertion to check for the d_model 100 case? Personally, I would prefer to leave the adaptation as it is than that.

robintibor:
> Sure, I can do that. Just to confirm, you mean catching the breakage whenever the output of the mrcnn is not equal to d_model, and giving an assertion error right? Do you want me to do the shhs adaptation on the basis of the initial param check or by making a subcase in this assertion to check for the d_model 100 case? Personally, I would prefer to leave the adaptation as it is than that.

Hm no to the assertion error, I wondered if given the input params one could set d model correctly to 99 or 100 so that it just wouldn't break in any case and there would be just warnings for cases that have never been used in this way by the original authors.

Div12345:
Oh nope, there are other parameters also dependent on d_model. For example, number of attention heads must be a factor of d_model. I tried that and reported it earlier [here](https://github.com/braindecode/braindecode/pull/341#issuecomment-958721030).

braindecode/models/sleep_stager_eldele_2021.py
```diff
@@ -0,0 +1,408 @@
...
+
+    def __init__(self, sfreq, n_tce=2, d_model=80, d_ff=120, n_attn_heads=5, dropout=0.1,
+                 input_size_s=30, n_classes=5, after_reduced_cnn_size=30, return_feats=False):
+        super(SleepStagerEldele2021, self).__init__()
+
+        input_size = np.ceil(input_size_s * sfreq).astype(int)
+
+        # exception case for shhs to make modification in mrcnn as per implementation in the paper
+        shhs = False
+        if input_size_s == 30 and sfreq == 125 and d_model == 100:
```
agramfort, line 78: all the sleep scoring models I know work with 30s windows. I think could even enforce this. Then it's just a matter of sfreq. If sfreq==100 do this and if 125 do that. If different crash. No need to mention shhs in the code. Just put the doctsring that if sfreq==125Hz then we will use the reference architecture from XXX et al. which was validated on SHHS dataset. This is just doc, no variable name etc.

makes sense?

robintibor: but from what I understood,f or the "shhs case" we still have to  perform the adaptation necessary for that case so that the model doesn't just crash (change kernel length). so we have to somehow do that right?

agramfort:
what do you mean do the "adaptation"? it's more than the "if else" I
suggested?

>

robintibor:
So @Div12345 is a bit busy atm and would take a bit time before he can continue. Should we merge this in its current form or make a braindecode release without this? Otherwise from my side release could be done. What do you think @agramfort @gemeinl @hubertjb ?

examples/plot_sleep_staging_eldele2021.py
```diff
@@ -0,0 +1,376 @@
...
+
+######################################################################
+# Preprocessing
+# ~~~~~~~~~~~~~
+#
+# Next, we preprocess the raw data. We convert the data to microvolts and apply
+# a lowpass filter. We omit the downsampling step of [1]_ as the Sleep
+# Physionet data is already sampled at a lower 100 Hz.
+
+from braindecode.preprocessing.preprocess import preprocess, Preprocessor, scale
```
agramfort, line 66: @robintibor @gemeinl this is not nice. Could we fix this so we read something like
```suggestion
from braindecode.preprocessing import preprocess, Preprocessor, scale
```

agramfort:
@Div12345 you need fix conflicts here

Div12345:
@agramfort I changed the name of the plot_sleep_staging example to plot_sleep_staging_chambon2018 to be more standardized like the other sleep staging examples. I think that is what is causing the conflict?

agramfort:
you need to merge the main branch into your branch. There has been changes
in this example
in parallel and this creates a conflict.

>

agramfort:
@Div12345 you need to `git merge main` into this PR to fix the conflict.

Div12345:
Just renaming the file seems to do it.

Div12345:
Gives `Already up to date`

agramfort:
I pushed a fix

please share the urls of the rendered examples in the doc build so we
can see how it looks

Div12345:
The Eldele example - [Link](https://1751-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_sleep_staging_eldele2021.html#sphx-glr-auto-examples-plot-sleep-staging-eldele2021-py) - Just have to change the title and references a bit. The renamed Chambon - [Link](https://1751-232335424-gh.circle-artifacts.com/0/dev/auto_examples/plot_sleep_staging_chambon2018.html#sphx-glr-auto-examples-plot-sleep-staging-chambon2018-py)

robintibor:
I would strongly be in favour of drastically reducing nr of epochs and/or possibly dataset size to speed up example on CI. now we know it would have good performance with reasonable epochs, so we can just cut to just 1 epoch from my point of view.. right now it has 11 min running time on CI just for this example... As before, let's just be clear to the reader that this is a modified example for faster runtime and be superclear how to run it for reasonable performance...

robintibor:
so let's merge this and have a release! how do you think @agramfort ?

robintibor:
Since there were no further comments, have commited the last suggestions inside and now it's merged! Thanks so much @Div12345 !! :)

agramfort:
thanks @Div12345

>

PR #345, opened 2021-10-21
Fix legacy break in WindowsDataset

cedricrommel:
In PR https://github.com/braindecode/braindecode/pull/261, `WindowsDataset` was modified to facilitate cropped-coding training. In the process, an important attribute (`y`) was removed from this class, which is one of the most used classes of the package I guess. Unfortunately, this might break existing code of many users (such as myself :) ).

Hence I propose in this PR to leave the attribute for legacy compliance reasons in the case where on-the-fly computation of the label is not required (`targets_from == 'metadata'`).

### DIFF
--- a/test/unit_tests/preprocessing/test_windowers.py
+++ b/test/unit_tests/preprocessing/test_windowers.py
@@ -76,8 +76,7 @@ def test_windows_from_events_n_jobs(lazy_loadable_dataset):
         assert pd.Series(ds1.windows.info).to_json() == \
                pd.Series(ds2.windows.info).to_json()
         assert ds1.description.equals(ds2.description)
-        assert np.array_equal(ds1.windows.metadata.loc[:, 'target'].to_numpy(),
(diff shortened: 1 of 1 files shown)


sliwy (commented):
thx @cedricrommel, we removed that only for consistency with the `channels` option.

The only problem I see with using `y` attribute is obtaining wrong labels when using `target_transform` in datasets. Thus, I would not recommend this way of obtaining labels to the users. Safer but slower (and less memory efficient way I guess due to need of loading the dataset into memory) is to iterate over the whole dataset and getting the second element of the tuple that should be the label. 
Maybe we can fix this by the way and make `y` correspond always to the labels you obtain while iterating over a dataset? If not we can just leave it as is and create an issue for that. Finally this is just to restore the attribute and it should not break anything. :)

robintibor:
So can you reopen this PR with this change @cedricrommel ?

PR #346, opened 2021-10-21
[WIP] EEGClassifier quick fix for cloning

sliwy:
Adding if statement that should resolve problem from #340 
@gemeinl if it works I can do the same for `EEGRegressor` or feel free to fix it there as well

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -60,7 +60,8 @@ class EEGClassifier(NeuralNetClassifier):
     def __init__(self, *args, cropped=False, callbacks=None,
                  iterator_train__shuffle=True, aggregate_predictions=True, **kwargs):
         self.cropped = cropped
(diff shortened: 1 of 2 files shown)


robintibor:
I don't think this will work in general, as scikit-learn expect all constructor arguments to remain unchanged, see
https://scikit-learn.org/stable/developers/develop.html 

> There should be no logic, not even input validation, and the parameters should not be changed. The corresponding logic should be put where the parameters are used, typically in fit. The following is wrong:
```python
def __init__(self, param1=1, param2=2, param3=3):
    # WRONG: parameters should not be modified
    if param1 > 1:
        param2 += 1
    self.param1 = param1
    # WRONG: the object's attributes should have exactly the name of
    # the argument in the constructor
    self.param3 = param2
```

Also trying to develop a fix atm

PR #347, opened 2021-10-21
[WIP] Parse callbacks on yield

robintibor:
Here is another way to solve https://github.com/braindecode/braindecode/pull/340

It passed tests locally, let's see if its all good


agramfort (approved):
whatever works ! :)

PR #348, opened 2021-10-21
[MRG] Incompatibility with `Skorch` master

jpaillard:
It seems that the use of skorch master triggers several conflicts. Though skorch developers announced that a new release will come shortly.

### DIFF
--- a/.github/workflows/tests-and-docs.yml
+++ b/.github/workflows/tests-and-docs.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        skorch_version: ['github', 'conda']
+        skorch_version: ['main', 'stable']
     defaults:
       run:
         shell: bash -l {0}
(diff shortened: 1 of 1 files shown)


sliwy:
Hi @lapaill, 
Let's wait for the release, so we don't have to change the skorch version in braindecode. We can start resolving some issues now as they plan to release it soon.

First step can be fixing `on_batch_end` methods in EEGRegressor and EEGClassifer that cause this errors:
```TypeError: on_batch_end() missing 2 required positional arguments: 'X' and 'y'``` as skorch won't unpack the data at this level (see more: https://github.com/skorch-dev/skorch/pull/807).

The solution would be to apply the fixes from https://skorch.readthedocs.io/en/latest/user/FAQ.html#migration-guide

@lapaill it would be great if you could fix that and then we can see what's more to fix. Let me know if you need any help with that. :)

robintibor:
Ah really great @sliwy that you already found the migration guide/fix! I think this change might actually allow us to get rid of the `ThrowAwayIndexLoader` 
https://github.com/braindecode/braindecode/blob/9033a0ab4f190a05fc3c44f9649428b1330de220/braindecode/util.py#L316-L340

the _last_window_inds_ stuff could possibly happen then directly in `on_batch_end`? Then question becomes more where to move this target conversion logic, which we apparently wanted to move to dataset at some point anyways... But these changes probably make more sense when skorch is released and we drop support for older skorch versions for master branch.

PR #353, opened 2021-10-29
Be less strict with window sizes exceeding trial

cedricrommel:
In PR https://github.com/braindecode/braindecode/pull/304, a new assertion check is added to `_compute_window_inds` (involved in trial-based windowing), to prevent trials which are too short compared to the requested window size. While the check is really useful for robustness, I wonder whether raising an error regardless of the severity of the inconsistency is not too harsh.

To illustrate what I mean, here is my personal example:
After rebasing onto braindecode/master, I encountered errors when loading the MASS dataset that I have been using for a while now. Probably due to `mne.resample` issues, it happened that one trial of one recording out of 64 had a slight misalignment, making its size be 3839 instead of 3840 (window size). And because of this, the whole dataset could not be loaded. So the first thing I did was find out what raw was causing the problem, what trial, go to the data saved locally, fix it, etc... But then I though that maybe this might happen quite often to other users as well and that warning the users (as it was done before PR https://github.com/braindecode/braindecode/pull/304) and just dropping the bad trial instead of raising an error and stopping all computation might be a good alternative.

What do you think?

### DIFF
--- a/braindecode/preprocessing/windowers.py
+++ b/braindecode/preprocessing/windowers.py
@@ -30,7 +30,7 @@ def create_windows_from_events(
         window_size_samples=None, window_stride_samples=None,
         drop_last_window=False, mapping=None, preload=False,
         drop_bad_windows=True, picks=None, reject=None, flat=None,
-        on_missing='error', n_jobs=1):
+        on_missing='error', accepted_bads_ratio=0.1, n_jobs=1):
     """Create windows based on events in mne.Raw.
 
     This function extracts windows of size window_size_samples in the interval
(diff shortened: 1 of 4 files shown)


robintibor:
I like the idea of this argument as an easier way to handle the case of too small recordings. I would prefer to keep the default to crash, i.e., `accepted_bads_ratio=0.` by default, and on crash put in the error message that this argument exists and can be used to prevent the crash (in the error message already displaying what ratio we have/what would need to be set to prevent crash. This way we ensure that it is not missed if some recording is too short (warnings can be more easily missed obviously).

robintibor:
great now only add something to whats_new missing @cedricrommel  :))

cedricrommel:
Sure, here it is :)

robintibor:
Amazing stuff, merged

PR #356, opened 2021-11-04
Fix incompatibility between augmentations and `skorch.NeuralNet.set_params`

cedricrommel:
Really sorry to open yet another PR @robintibor 😅 
We've realized with @lapaill that braindecode would crash if we instatiate a class inheriting from `skorch.NeuralNet` with an argument based on `Transform` objects and then try to set any other unrelated parameter with the method `set_params`. Indeed, the latter will inspect each attribute of the `NeuralNet` object (even if it was previously defined) and call their `get_params` method when the latter is implemented. The problem is the `Transform` implements such a method (for an unrelated reason) and outputs a tuple instead of a dict, which would make skorch raise an error.

Hence, this PR solves this incompatibility by replacing the tuples by dictionaries for every Transform. It is also cleaner this way I think.

### DIFF
--- a/braindecode/augmentation/transforms.py
+++ b/braindecode/augmentation/transforms.py
@@ -208,7 +208,7 @@ class ChannelsShuffle(Transform):
         Float setting the probability of applying the operation.
     p_shuffle: float | None, optional
         Float between 0 and 1 setting the probability of including the channel
-        in the set of permutted channels. Defaults to 0.2.
+        in the set of permuted channels. Defaults to 0.2.
     random_state: int | numpy.random.Generator, optional
         Seed to be used to instantiate numpy random number generator instance.
         Used to decide whether or not to transform given the probability
@@ -251,7 +251,7 @@ class ChannelsShuffle(Transform):
             Contains
             * p_shuffle : float
                 Float between 0 and 1 setting the probability of including the
-                channel in the set of permutted channels.
+                channel in the set of permuted channels.
             * random_state : numpy.random.Generator
                 The generator to use.
         """
--- a/braindecode/augmentation/transforms.py
+++ b/braindecode/augmentation/transforms.py
@@ -121,12 +121,14 @@ class FTSurrogate(Transform):
 
         Returns
         -------
-        phase_noise_magnitude : float
-            The magnitude of the transformation.
-        rng : numpy.random.Generator
-            The generator to use.
+        params : dict
+            Contains the magnitude of the transformation
+            (phase_noise_magnitude) and the random state.
         """
-        return self.phase_noise_magnitude, self.rng
+        return {
+            "phase_noise_magnitude": self.phase_noise_magnitude,
+            "random_state": self.rng,
+        }
 
 
 class ChannelsDropout(Transform):
@@ -178,13 +180,14 @@ class ChannelsDropout(Transform):
 
         Returns
         -------
-        p_drop : float
-            Float between 0 and 1 setting the probability of dropping each
-            channel.
-        rng : numpy.random.Generator
-            The generator to use.
+        params : dict
+            Contains the probability of dropping each channel (p_drop) and the
+            random state.
         """
-        return self.p_drop, self.rng
+        return {
+            "p_drop": self.p_drop,
+            "random_state": self.rng,
+        }
 
 
 class ChannelsShuffle(Transform):
@@ -237,13 +240,14 @@ class ChannelsShuffle(Transform):
 
         Returns
         -------
-        p_shuffle : float
-            Float between 0 and 1 setting the probability of including the
-            channel in the set of permutted channels.
-        rng : numpy.random.Generator
-            The generator to use.
+        params : dict
+            Contains the probability of including the channel in the set of
+            permutted channels (p_shuffle) and the random state.
         """
-        return self.p_shuffle, self.rng
+        return {
+            "p_shuffle": self.p_shuffle,
+            "random_state": self.rng,
+        }
 
 
 class GaussianNoise(Transform):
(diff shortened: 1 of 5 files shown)


robintibor:
```python
E               ValueError: This channel name (EOG horizontal) doesn't exist in info.
```

Strange do you know what happened there @cedricrommel ? Or is it unrelated to this PR, something changed somewhere else?

cedricrommel:
I might have a clue 🙃 : https://twitter.com/mne_news/status/1456240588421046279?s=20

cedricrommel:
I'll set the mne version to previous one for now, and we can open a new PR to ensure compatibility with new `mne` when we want to. What do you think?

robintibor:
oki good to know then we need to ensure compatibility for sure, as you said in separate PR

environment.yml
```diff
@@ -18,5 +18,5 @@ dependencies:
 - joblib
 - memory_profiler
 - pip:
-  - mne
+  - mne==0.23.0
```
robintibor, line 21: should be possible to remove this now after rebase @cedricrommel

braindecode/augmentation/transforms.py
```diff
@@ -178,13 +180,14 @@ def get_params(self, X, y):
...
         Returns
         -------
-        p_drop : float
-            Float between 0 and 1 setting the probability of dropping each
-            channel.
-        rng : numpy.random.Generator
-            The generator to use.
+        params : dict
+            Contains the probability of dropping each channel (p_drop) and the
+            random state.
```
agramfort, line 185: can you cleanly document what the dict contains ? key names and value types?

braindecode/augmentation/transforms.py
```diff
@@ -237,13 +240,14 @@ def get_params(self, X, y):
...
         Returns
         -------
-        p_shuffle : float
-            Float between 0 and 1 setting the probability of including the
-            channel in the set of permutted channels.
-        rng : numpy.random.Generator
-            The generator to use.
+        params : dict
+            Contains the probability of including the channel in the set of
+            permutted channels (p_shuffle) and the random state.
```
agramfort, line 245:
```suggestion
            permuted channels (p_shuffle) and the random state.
```

environment.yml
agramfort: +1

agramfort (commented):
please also check formatting. it does not look good:

https://1706-232335424-gh.circle-artifacts.com/0/dev/generated/braindecode.augmentation.SmoothTimeMask.html#braindecode.augmentation.SmoothTimeMask.get_params

test/unit_tests/augmentation/conftest.py
```diff
@@ -26,3 +33,24 @@ def random_batch(rng_seed, batch_size=5):
     rng = check_random_state(rng_seed)
     X = torch.from_numpy(rng.random((batch_size, 66, 51))).float().to(device)
     return X, torch.zeros(batch_size)
+
+
+class MockModule(torch.nn.Module):
```
cedricrommel, line 38: I'd remove it from `test_eeg_classifier` now that it has been moved here :)

test/unit_tests/augmentation/test_base.py
```diff
@@ -2,15 +2,16 @@
...
+import pytest
 import torch
-import mne
+from sklearn.utils import check_random_state
+from skorch.helper import predefined_split
 
-from braindecode.augmentation.base import (
-    Transform, Compose, AugmentedDataLoader
-)
+from braindecode.augmentation.base import AugmentedDataLoader
```
cedricrommel, line 12: Très propre 😄

test/unit_tests/augmentation/conftest.py
```diff
@@ -26,3 +33,24 @@ def random_batch(rng_seed, batch_size=5):
...
+        super().__init__()
+        self.preds = to_tensor(preds, device='cpu')
+        self.linear = torch.nn.Linear(5, 5)
+
+    def forward(self, x):
+        return self.preds
+
+
+@pytest.fixture
+def mock_clf():
```
cedricrommel, line 49: I would rename it into augmented_mock_clf as you have added a transform to it

test/unit_tests/augmentation/conftest.py
lapaill: It is still used in `test_eeg_classifier` to test `predict` and `clone` functions. These functions are not tested in `test_base` elsewhere.

test/unit_tests/augmentation/conftest.py
lapaill: Done

test/unit_tests/augmentation/conftest.py
cedricrommel: Ahh, this `conftest.py` concerns only augmentation. Sorry, I though it was a general file for all tests.

cedricrommel:
Ok, now the doc is clean @agramfort :)

agramfort (approved):
@robintibor ok for you?

test/unit_tests/augmentation/test_base.py
```diff
@@ -173,3 +174,13 @@ def test_dataset_with_transform(concat_windows_dataset):
     concat_windows_dataset.transform = transform
     transformed_X = concat_windows_dataset[0][0]
     assert torch.all(transformed_X == factor)
+
+
+def test_set_params(augmented_mock_clf, random_batch):
```
robintibor, line 179: Could you add a small comment here explaining what this is testing? Just that it doesn't crash? It's a bit hard to understand for me without any comment

robintibor (changes requested):

robintibor:
doc looks good to me, would be great if you can add the comment I mentioned above @cedricrommel than we could merge from my side

robintibor:
amazing stuff, great, good to merge this before release, thanks @cedricrommel

agramfort:
🙌

cedricrommel:
🙌

PR #363, opened 2021-11-21
add more subjects in plot_sleep...

agramfort:
in an attempt to improve the plot_sleep_staging.py example I added more subjects.
Doing so I get a model that does not learn at all...

@hubertjb @gemeinl @robintibor can you replicate?

as more and more people get started with sleep scoring using this example I'd
like to have this example as complete and working out of the box as possible...

### DIFF
--- a/examples/plot_sleep_staging.py
+++ b/examples/plot_sleep_staging.py
@@ -51,7 +51,6 @@ from braindecode.datasets.sleep_physionet import SleepPhysionet
 
 dataset = SleepPhysionet(
     subject_ids=list(range(50)), recording_ids=[2], crop_wake_mins=30)
-    # subject_ids=[0, 1], recording_ids=[2], crop_wake_mins=30)
 
 
 ######################################################################
--- a/examples/plot_sleep_staging.py
+++ b/examples/plot_sleep_staging.py
@@ -50,7 +50,8 @@ from numbers import Integral
 from braindecode.datasets.sleep_physionet import SleepPhysionet
 
 dataset = SleepPhysionet(
-    subject_ids=[0, 1], recording_ids=[2], crop_wake_mins=30)
+    subject_ids=list(range(50)), recording_ids=[2], crop_wake_mins=30)
+    # subject_ids=[0, 1], recording_ids=[2], crop_wake_mins=30)
 
 
 ######################################################################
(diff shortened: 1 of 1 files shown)


lapaill:
I could reproduce the issue: the model doesn't learn anything, the accuracy is constant and equal to 0.2. 
I solved it by parsing the argument `picks=['Fpz-Cz', 'Pz-Oz']` to `create_windows_from_events()`. Then the accuracy starts to exceed 0.2 (cf picture). 
I'm still trying to figure out why this is necessary since without this fix `n_channels = 2` already. 
image

sliwy:
maybe order of the channels between subjects differs?

agramfort:
for some reason it depends on which subjects I use. If I only take the
first subjects
everything works. I suspect some crappy data for certain subjects...

>

hubertjb:
Looks like both with and without specifying `picks` I get this on the first 50 subjects:
image

I think the culprit is actually the dropout rate. I had set it to a pretty low value of 50% to avoid pure overfitting with 2 subjects. With dropout=0.9 instead, I get the following (same 50 subjects):
image

I guess we could add a few sentences at the end of the example to recommend some hyperparameters for "heavier" testing. Increasing the sequence length and stride might also improve results without a big change in compute time.

hubertjb:
I was not expecting this, but it looks like increasing the dropout rate from 50 to 90% fixes this (same 50 subjects):
image

It could be that subjects 1 and 2 that are used in the current example are actually pretty similar and heavy regularization is not required, but some of the subjects here introduce a lot variability...

agramfort:
can you push a fix to my PR?

>

bruAristimunha:
Hi @agramfort,

I'm wondering, can I help in this PR?

bruAristimunha:
Hey @tgnassou,

What do you think to work in this code for the first day of the code sprint?

tgnassou:
Hey, the problem here is the sampler. Your sampler doesn't shuffle the sequences in a batch. So, the network always sees the same order of samples, which prevents the network from learning something. 

Two possibilities here:
1 - I added a `random`  boolean parameter to [`SequenceSampler`](https://braindecode.org/dev/generated/braindecode.samplers.SequenceSampler.html), which controls if the sequences are randomized or not.
2 - Use [`BalancedSequenceSampler`](https://braindecode.org/dev/generated/braindecode.samplers.BalancedSequenceSampler.html). This sampler randomizes the sequences and balances the number of classes seen in epochs. But this one is the one used in Usleep. Chambon et al. use a balanced sampling, but I still need to determine which kind.

tgnassou:
Let me know what do you think @agramfort @bruAristimunha

bruAristimunha:
It is look good to me @tgnassou. Can you open another PR?

PR #367, opened 2021-12-04
attempt to simplify some examples + fix stuff for recent mne

agramfort:
@hubertjb @robintibor @gemeinl 

can you have a look?

if we could simplify further before the next release it would be great

### DIFF
--- a/examples/plot_sleep_staging_eldele2021.py
+++ b/examples/plot_sleep_staging_eldele2021.py
@@ -248,14 +248,13 @@ if cuda:
 #
 
 from skorch.helper import predefined_split
-from skorch.callbacks import EpochScoring, EarlyStopping
+from skorch.callbacks import EpochScoring
 from braindecode import EEGClassifier
 
 lr = 1e-3
 batch_size = 32
 n_epochs = 3  # we use few epochs for speed and but more than one for plotting
 
-early_stopping = EarlyStopping(patience=10)
 train_bal_acc = EpochScoring(
     scoring='balanced_accuracy', on_train=True, name='train_bal_acc',
     lower_is_better=False)
@@ -263,7 +262,6 @@ valid_bal_acc = EpochScoring(
     scoring='balanced_accuracy', on_train=False, name='valid_bal_acc',
     lower_is_better=False)
 callbacks = [
-    ('patience', early_stopping),
     ('train_bal_acc', train_bal_acc),
     ('valid_bal_acc', valid_bal_acc)
 ]
--- a/examples/plot_sleep_staging_eldele2021.py
+++ b/examples/plot_sleep_staging_eldele2021.py
@@ -53,8 +53,10 @@ from numbers import Integral
 from braindecode.datasets import SleepPhysionet
 
 subject_ids = [0, 1]
+crop = (0, 30 * 400)  # we only keep 400 windows of 30s to speed example
 dataset = SleepPhysionet(
-    subject_ids=subject_ids, recording_ids=[2], crop_wake_mins=30)
+    subject_ids=subject_ids, recording_ids=[2], crop_wake_mins=30,
+    crop=crop)
 
 ######################################################################
 # Preprocessing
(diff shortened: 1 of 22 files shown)


agramfort:
@robintibor @hubertjb @gemeinl I think I am done here. See rendered doc at:

https://1786-232335424-gh.circle-artifacts.com/0/dev/index.html

let me know if you see anything weird.

robintibor:
Fixed small typo that I missed in PR before @agramfort 
In general I don't have objections except I am wondering about the early stopping on the validation set. Normally, I would expect there to be a train/val/test split if one  uses early stopping, now it looks like we are doing early stopping on the set that we report metrics on... might be quite misunderstandable no? @agramfort

examples/plot_sleep_staging_eldele2021.py
```diff
@@ -1,5 +1,5 @@
 """
-Sleep staging on the Sleep Physionet dataset usng Eldele2021
+Sleep staging on the Sleep Physionet dataset using Eldele2021
 ============================================================
```
agramfort, line 3:
```suggestion
=============================================================
```

robintibor:
ahya this with the `==`  I should really start to realize haha

agramfort:
>
> Fixed small typo that I missed in PR before @agramfort
> <https://github.com/agramfort>
>

thx


> In general I don't have objections except I am wondering about the early
> stopping on the validation set. Normally, I would expect there to be a
> train/val/test split if one uses early stopping, now it looks like we are
> doing early stopping on the set that we report metrics on... might be quite
> misunderstandable no? @agramfort <https://github.com/agramfort>
>

that's a very good point. So maybe we should update the examples to use a
train, valid and test sets?

robintibor:
Well seems like a bigger thing to do that? why did you want to introduce early stopping?
In general, can we maybe merge this as is, make release now and do this separately?

agramfort:
>
> Well seems like a bigger thing to do that? why did you want to introduce
> early stopping?
>

well see these examples are educational materials. People get started with
deep learning
on EEG from such examples. What do we want them to do? What would you
recommend
them?

> In general, can we maybe merge this as is, make release now and do this
> separately?
>

I can revert the changes on EarlyStopping if you want as I agree with you
that without
a valid / test sets it's not great...

robintibor:
> Well seems like a bigger thing to do that? why did you want to introduce early stopping?
> well see these examples are educational materials. People get started with deep learning on EEG from such examples. What do we want them to do? What would you recommend them?
> In general, can we maybe merge this as is, make release now and do this separately?
> I can revert the changes on EarlyStopping if you want as I agree with you that without a valid / test sets it's not great...

Yeah I would revert. I would say general deep learning experience as well as our experience suggests that in many settings, with the right training pipeline (e.g. right regularizations and hyperparameters), there is no reason for early stopping: Longer training just keeps increasing validation scores or at least does not decrease them. So I think training without early stopping is also something beginners should learn as the default way.

robintibor:
so if you would revert I would make the release :)

agramfort:
done

robintibor:
Also keep in mind both popular learning rate schedules like cosine annealing and some state-of-the-art techniques to further improve accuracies like exponential moving average are not that compatible with  early stopping https://pytorch.org/blog/how-to-train-state-of-the-art-models-using-torchvision-latest-primitives/

robintibor:
great, merged

agramfort:
>
> Also keep in mind both popular learning rate schedules like cosine
> annealing and some state-of-the-art techniques to further improve
> accuracies like exponential moving average are not that compatible with
> early stopping
> https://pytorch.org/blog/how-to-train-state-of-the-art-models-using-torchvision-latest-primitives/
>

which part of this page are you point me to?

robintibor:
This as an example of a very state of the art training pipeline for deep learning computer vision that does not use early stopping. Also, parts that are at least not really in the spirit of using early stopping from my view:
* cosine annealing learning rate schedule
* exponential moving average of weights

robintibor:
Also in general wanted to bring this pipeline to your attention, several of the methods there are fairly generic and could also be tried for our deep learning pipelines on EEG

agramfort:
we should have a chat ;)

Issue #379, opened 2022-05-04
 Linking source code button in braindecode doc


bruAristimunha:
Hello guys!

I really love this library! I have a simple issue related to the usability of the site. Furthermore, I don't know if this is the correct place to ask this or if I can ask for this, but it is possible to change the hyperlink for the source page in the braindecode documentation?

Nowadays, if I click on the documentation, it will lead to the page that generates it rather than the source code. I think it is just a parameter to change during the generation of the site.

For instance:

![image](https://user-images.githubusercontent.com/42702466/166643910-615ee9bf-d366-43b0-90ad-7a34ab2ad8d0.png)

After clicking on the View page source:

![image](https://user-images.githubusercontent.com/42702466/166644126-39b4bfb1-9571-4906-b7cf-2ac3a5ce4a28.png)

PR #380, opened 2022-05-04
Adding source code button for documentation.

bruAristimunha:
Adding an option in sphinx to generate page with source code. It solves the issue: #379

### DIFF
--- a/braindecode/preprocessing/preprocess.py
+++ b/braindecode/preprocessing/preprocess.py
@@ -8,14 +8,14 @@
 #
 # License: BSD (3-clause)
 
-from warnings import warn
-from functools import partial
-from collections.abc import Iterable
-
 import numpy as np
 import pandas as pd
-from sklearn.utils import deprecated
+from collections.abc import Iterable
+from functools import partial
 from joblib import Parallel, delayed
+from mne import create_info
+from sklearn.utils import deprecated
+from warnings import warn
 
 from braindecode.datasets.base import BaseConcatDataset, BaseDataset, WindowsDataset
 from braindecode.datautil.serialization import (
@@ -47,6 +47,7 @@ class Preprocessor(object):
     kwargs:
         Keyword arguments to be forwarded to the MNE function.
     """
+
     def __init__(self, fn, *, apply_on_array=True, **kwargs):
         if hasattr(fn, '__name__') and fn.__name__ == '<lambda>':
             warn('Preprocessing choices with lambda functions cannot be saved.')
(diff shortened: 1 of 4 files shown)


agramfort (approved):

agramfort:
Thanks 🙏 @bruAristimunha 

The ci failures are I think unrelated. Would you have time to look into it ? 🙏

bruAristimunha:
Yes professor @agramfort, I can look!

bruAristimunha:
Hello @agramfort!

I looked into the code. One error was related to a small difference in the numeric precision. It probably has been caused by the difference between OS or a macro updated in PyTorch or another library. I increased the threshold from 0.1E-3 to 0.5E-2.

The second and third errors involve implementing a filter bank in the module preprocessing.

In the line [469]( https://github.com/braindecode/braindecode/blob/5b4d8daee842c72fbd463b9d4528957fcd32f7be/braindecode/preprocessing/preprocess.py#L469) there is an overwriting of the info values of a mne object. This action tries to preserve the value of the high and low bands. This type of access is not allowed anymore, according to the documentation of the mne. 
image

I recreated all the information from the data object and redefined the value.

agramfort:
@bruAristimunha I let you investigate the CI issues. 🙏

bruAristimunha:
@agramfort, I think everything works.

robintibor:
Thanks for the effort @bruAristimunha , @gemeinl  what do you think about comment above? If we think it's fine we could merge I guess?

agramfort (approved):
Thx @bruAristimunha 

@robintibor @gemeinl merge if happy

bruAristimunha:
> Thanks for the effort @bruAristimunha , @gemeinl what do you think about comment above? If we think it's fine we could merge I guess?

I thank you for the library. If there's anything I can do to contribute more seriously to the library, I'm super available!

braindecode/preprocessing/preprocess.py
```diff
@@ -8,14 +8,14 @@
...
-
 import numpy as np
 import pandas as pd
-from sklearn.utils import deprecated
+from collections.abc import Iterable
+from functools import partial
 from joblib import Parallel, delayed
+from mne import create_info
+from sklearn.utils import deprecated
+from warnings import warn
```
agramfort, line 18: can you just revert the reordering of the imports? we use the convention to import standard lib first, then core packages (numpy, torch) and then local imports

agramfort:
@bruAristimunha what is your status? are you currently a student?

bruAristimunha:
I am a second year PhD student in Computer Science, I would love to collaborate with the library and the projects.

agramfort:
Thx @bruAristimunha 

Your next PRs will be more than welcome too 🙏

robintibor:
Great thanks so much @bruAristimunha :)

PR #381, opened 2022-05-10
[MRG] Fix predict_trials

gemeinl:
When the module is on GPU and data is on CPU, `predict_trials` will fail.
In the tests, everything is always on CPU, therefore it did not attract attention.
With this PR always move the data to the device of the module parameters.

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -319,5 +319,5 @@ class EEGClassifier(NeuralNetClassifier):
             dataset=X,
             return_targets=return_targets,
             batch_size=self.batch_size,
-            num_workers=self.iterator_valid__num_workers,
+            num_workers=clf.get_iterator(X, training=False).loader.num_workers,
         )
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -319,5 +319,5 @@ class EEGClassifier(NeuralNetClassifier):
             dataset=X,
             return_targets=return_targets,
             batch_size=self.batch_size,
-            num_workers=clf.get_iterator(X, training=False).loader.num_workers,
+            num_workers=self.get_iterator(X, training=False).loader.num_workers,
         )
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -318,5 +318,6 @@ class EEGClassifier(NeuralNetClassifier):
             module=self.module,
             dataset=X,
             return_targets=return_targets,
+            batch_size=self.batch_size,
             num_workers=self.iterator_valid__num_workers,
         )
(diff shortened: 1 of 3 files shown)


braindecode/training/scoring.py
```diff
@@ -430,6 +430,8 @@ def predict_trials(module, dataset, return_targets=True):
     all_preds, all_ys, all_inds = [], [], []
     with torch.no_grad():
         for X, y, ind in loader:
+            device = next(module.parameters()).device
+            X = X.to(device)
```
agramfort, line 434: I could consider this as black magic. What if my data is already on the device? this generates so extra computation for nothing? Could say it's the responsability of the user to pass a dataset that is on the right device.

gemeinl: I would disagree. The workflow is typically as follows:
A dataset is created, data is loaded onto CPU / not loaded at all.
A model is created on CPU and sent to GPU, if available / desired.
In the training loop, a batch is collected (loads data, if not already loaded) and put to the device compatible with the model.
Since we use skorch with the scikit-learn API `fit` and `predict`, the training loop is hidden and the device allocation should be / is handled by skorch.

`predict_trials` is now something special to braindecode, due to the cropped decoding. The function is __not__ called during training but only once after training has finished to create predictions on trial level instead of on compute window level as done during training. So we implement the data iteration ourselves and also the allocation to the appropriate device. The overhead of extra computations is also negligible.

robintibor: yes agree with Lukas, e.g. see
https://github.com/skorch-dev/skorch/blob/fe71e3d55a4ae5f5f94ef7bdfc00fca3b3fd267f/skorch/net.py#L208-L212

braindecode/classifier.py
```diff
@@ -318,4 +318,5 @@ def predict_trials(self, X, return_targets=True):
             module=self.module,
             dataset=X,
             return_targets=return_targets,
+            num_workers=self.iterator_valid__num_workers,
```
robintibor, line 321: `iterator_valid__num_workers` may not exist, so 
```suggestion
            getattr(clf, 'iterator_valid__num_workers', 0),
```
or 
```suggestion
            clf.get_iterator(X, training=False).loader.num_workers,
```

robintibor: Slightly prefer the second as I feel scared the first one will always return 0 even if variable names have changed etc... so even iif sth goes wrong we wouldn't notice

braindecode/regressor.py
```diff
@@ -300,6 +300,7 @@ def predict_trials(self, X, return_targets=True):
             module=self.module,
             dataset=X,
             return_targets=return_targets,
+            num_workers=self.iterator_valid__num_workers,
```
robintibor, line 303: Same here as above

robintibor: @gemeinl

agramfort (approved):
@robintibor merge if happy

thx @gemeinl

robintibor:
Great stuff thx @gemeinl

PR #384, opened 2022-05-13
[WIP} Move stored preds in callbacks to CPU

robintibor:
EpochScoring with caching by skorch keeps predictions in the GPU memory if model is on GPU. This can cause out of memory problems (it is a known problem: https://github.com/skorch-dev/skorch/issues/509)

This PR should fix that via a workaround on our side. Can you test it @AKiessner ? Basically callbacks should work fine again wrt to GPU memory.

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -12,6 +12,7 @@ from skorch.callbacks import EpochTimer, BatchScoring, PrintLog, EpochScoring
 from skorch.classifier import NeuralNet
 from skorch.classifier import NeuralNetClassifier
 from skorch.utils import train_loss_score, valid_loss_score, noop, to_numpy
+import torch
 
 from .training.scoring import (PostEpochTrainScoring,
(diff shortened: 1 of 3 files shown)


robintibor:
Also decided to directly call module forward in the `predict_with_window_inds_and_ys` functions to avoid further overhead, may also speed up scoring @gemeinl . We may wanna consider merging these functions and parts of the `predict_trials` function as well to avoid duplicate code (like a `predict_with_window_inds_and_ys`  function in scoring that gets iterator and model, is called by other functions)

agramfort (commented):
@robintibor I trust your judgement here but we should be careful not to slow things down for other use cases.

I would ask the skorch devs what they think about this...

robintibor:
Ya judging from discussion in issue linked above, they would need somebody to implement an option how to postprocess preds in epoch scoring. And until that is done, expect you to extend as we do in case it is not implemented there.
So merging this now as seems to work for @AKiessner  again

PR #385, opened 2022-05-21
Remove default drop_bad_windows=True

gemeinl:
Force user to decide on drop_bad_windows. 
Update docstring that this makes everything super slow.

```
Total time: 13.6696 s
File: /opt/conda/lib/python3.9/site-packages/braindecode/preprocessing/windowers.py
Function: _create_fixed_length_windows at line 325
```
```
Total time: 1.85751 s
File: /opt/conda/lib/python3.9/site-packages/braindecode/preprocessing/windowers.py
Function: _create_fixed_length_windows at line 325
```

### DIFF
--- a/braindecode/preprocessing/windowers.py
+++ b/braindecode/preprocessing/windowers.py
@@ -26,10 +26,10 @@ from ..datasets.base import WindowsDataset, BaseConcatDataset
 
 # XXX it's called concat_ds...
 def create_windows_from_events(
-        concat_ds, trial_start_offset_samples=0, trial_stop_offset_samples=0,
-        window_size_samples=None, window_stride_samples=None,
-        drop_last_window=False, mapping=None, preload=False,
-        drop_bad_windows=True, picks=None, reject=None, flat=None,
+        concat_ds, drop_bad_windows, trial_start_offset_samples=0,
+        trial_stop_offset_samples=0, window_size_samples=None,
+        window_stride_samples=None, drop_last_window=False, mapping=None,
+        preload=False, picks=None, reject=None, flat=None,
         on_missing='error', accepted_bads_ratio=0.0, n_jobs=1):
     """Create windows based on events in mne.Raw.
 
@@ -54,6 +54,13 @@ def create_windows_from_events(
     ----------
     concat_ds: BaseConcatDataset
         A concat of base datasets each holding raw and description.
+    drop_bad_windows: bool
+        If True, call `.drop_bad()` on the resulting mne.Epochs object. This
+        step allows identifying e.g., windows that fall outside of the
+        continuous recording. It is suggested to run this step here as otherwise
+        the BaseConcatDataset has to be updated as well. Caution: very slow!
+        Setting this to True will increase runtime of this function by a
+        factor of over 7.
     trial_start_offset_samples: int
         Start offset from original trial onsets, in samples. Defaults to zero.
     trial_stop_offset_samples: int
@@ -76,11 +83,6 @@ def create_windows_from_events(
         If True, preload the data of the Epochs objects. This is useful to
         reduce disk reading overhead when returning windows in a training
         scenario, however very large data might not fit into memory.
-    drop_bad_windows: bool
-        If True, call `.drop_bad()` on the resulting mne.Epochs object. This
-        step allows identifying e.g., windows that fall outside of the
-        continuous recording. It is suggested to run this step here as otherwise
-        the BaseConcatDataset has to be updated as well.
     picks: str | list | slice | None
         Channels to include. If None, all available channels are used. See
         mne.Epochs.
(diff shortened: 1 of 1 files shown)


gemeinl:
Data actually has to be loaded for lazy loading to work, which defeating the entire purpose.
To be able to use `mne.Epochs.get_data(item=index)` in lazy loading, mne forces us to call `nme.Epochs.drop_bad_windows`. For one, this is super slow as pointed out above. For another, this loads the data even if it is called with `reject=None, flat=None`.

Is there something we can do about this @agramfort ?

agramfort:
yes to know what epoch to drop you must load them.

now you could filter out the bad windows outside the data pretty easily as you know without loading how long it is.

gemeinl:
What if I know my data is clean and I don't want to drop anything. 
How can I make mne not drop and also not load the data?
As I said, it loads although I set both arguments to None.

agramfort:
if you don't pass any reject parameter, if you have no BAD annotation and no event outside of the data range, then I don't see the problem. Maybe we can talk during our meeting on Friday?

gemeinl:
Thanks, but somehow I cannot get it to work as I intend. Maybe you can help me figure out what I am missing in the snippet below. Friday is a bridge day in Germany and I will be on vacation.

```
import os
import tempfile

import mne
import numpy as np

# create fake data and export as raw
n_ch = 21
n_mins = 15
sfreq = 100
raw = mne.io.RawArray(
    data=np.random.rand(n_ch*n_mins*60*sfreq).reshape(n_ch, -1),
    info=mne.create_info(
        ch_names=[str(i) for i in range(n_ch)],
        sfreq=sfreq,
    ),
)
tmpdir = tempfile.mkdtemp()
tmpfile = os.path.join(tmpdir, 'raw.edf')
mne.export.export_raw(tmpfile, raw)

# read the raw with preload=False
raw = mne.io.read_raw_edf(
    input_fname=tmpfile,
    preload=False,
)

# create epochs that are exactly 1s = 100 samples duration
# apart from this, nothing should be done here
window_size_samples = 100
epochs = mne.Epochs(
    raw,
    events=[[0, 0, 0]],
    tmin=0,
    tmax=(window_size_samples-1)/sfreq,
    baseline=None,
    reject_by_annotation=False,
    reject=None,
    flat=None,
    proj=False,
    preload=False,
)
assert epochs.times.size == window_size_samples

# this raises:
# 'ValueError: item must be None in epochs.get_data() unless
# bads have been dropped. Consider using epochs.drop_bad().'
#epochs.get_data(item=0)

# this loads but is supposed to not load. it should only enable get_data(item=index)
# 'Loading data for 1 events and 10000 original time points ...
# 0 bad epochs dropped'
epochs.drop_bad(reject=None, flat=None)

# now this works
epochs.get_data(item=0)
```

gemeinl:
@agramfort Any tips on the snippet above?

agramfort:
this works:

```python
import os
import tempfile

import mne
import numpy as np

# create fake data and export as raw
n_ch = 21
n_mins = 15
sfreq = 100
raw = mne.io.RawArray(
    data=np.random.rand(n_ch*n_mins*60*sfreq).reshape(n_ch, -1),
    info=mne.create_info(
        ch_names=[str(i) for i in range(n_ch)],
        sfreq=sfreq,
    ),
)
tmpdir = tempfile.mkdtemp()
tmpfile = os.path.join(tmpdir, 'raw.edf')
mne.export.export_raw(tmpfile, raw)

# read the raw with preload=False
raw = mne.io.read_raw_edf(
    input_fname=tmpfile,
    preload=False,
)

# create epochs that are exactly 1s = 100 samples duration
# apart from this, nothing should be done here
window_size_samples = 100
epochs = mne.Epochs(
    raw,
    events=[[0, 0, 0]],
    tmin=0,
    tmax=(window_size_samples-1)/sfreq,
    baseline=None,
    reject_by_annotation=False,
    reject=None,
    flat=None,
    proj=False,
    preload=False,
)
assert epochs.times.size == window_size_samples

epochs[0].get_data()
```

does it help?

gemeinl:
Thanks @agramfort but no, it is about the lazy loading in braindecode which is realized with `get_data(item=index)`.

I'll try to explain more what is going on:
We want to train on the entire TUH EEG Corpus. Let's say it's 2 TB on disc with 60000 EDFs (it is not but it does not matter).
We internally agreed on some standard preprocessing which requires to load the data, for example resampling.
We have stored the preprocessed data back to disk which take up another 2 TB.

Now our steps look as follows:
- We read the raws from disk (this is quite slow, we might have to talk about this at another occasion, since mne makes up for 90% of that time).
- We apply additional preprocessing which does not require to load the data, for example setting tmin and tmax. This is reasonably fast.
- We create compute windows which does not require to load the data. In theory, this should also be fast, since it is only creating some indices like start and stop of a window, however, it is not due to `drop_bad()`. For one, it is super slow, for another it consumes a lot of memory, since it loads all the data (in our case with multiple workers in parallel). We still have to call it to afterwards be able to use lazy loading.
- We train with lazily loading mini-batches of compute windows from the Epochs objects.

We could do the windowing once and store the Epochs to disc again, however, this would consume another 2 TB of data. It also limits us to these specific windowing arguments. Additionally, braindecode serialization is suboptimal. For me, this is not a good option as I said, all intermediate steps should be fast enough.
So from my point of view `drop_bad()` is misbehaving in this setting, since it still loads the data, although all arguments were set to `None`. Why is that? Can we have a way to just declare all epochs as good, such that we can use lazy loading afterwards?
From a deep learning point of view we probably never want to drop bad epochs, we want the models to learn to ignore artifacts or whatever is going on.

agramfort:
@gemeinl ping me on the mne discord channel. it's easier to discuss. ok?

gemeinl:
Trying to improve the drop_bad situation on mne side:
https://github.com/mne-tools/mne-python/pull/10718

gemeinl:
Solved, due to changes implemented on mne side: https://github.com/mne-tools/mne-python/pull/10718

robintibor:
Only seeing now, thanks a lot for making this happen in mne @gemeinl

PR #391, opened 2022-06-22
added verbosity args to windowers

gemeinl:


gemeinl:
> running something like this `mne.set_config("MNE_LOGGING_LEVEL", "WARNING")` should otherwise silence mne verbosity

Unfortunately, for me, this does not seem to work. Also `mne.set_log_level('ERROR')` which I used before does not work.
I suspect that parallelization through joblib resets / removes / does not inherit the logging level?
My output is flooded while indexing individual compute windows with `Loading data for 1 events and 6000 original time points ...`

bruAristimunha:
@gemeinl, It doesn't work for me either, but only when I process in parallel. Is it your case?

bruAristimunha:
@agramfort, can you merge this option?

agramfort (approved):

Issue #378, opened 2022-05-03
Unified Validation Scheme


martinwimpff:
In my opinion this package still needs a unified way to evaluate DL models.

**Background**
As everyone knows, there are usually 3 different sets: training, validation, testing.
One trains on the training set, validates and tunes the model (i.e. the Hyperparameters) on the validation set and finally evaluates on the unseen test set.

As it is arguably the most popular BCI dataset I will make examples regarding the BCI competition IV 2a dataset. As a starting point I want to discuss WithinSubject Validation.
Regarding the BCI IV 2a dataset the train-test split is quite obvious: `session_T` for training and `session_E` for testing.
The problems come with the validation set.

**Examples**

1. [MOABB](https://github.com/NeuroTechX/moabb/blob/d4305c42380c9c4eb72b1b20b4bf180dc52d2b16/moabb/evaluations/evaluations.py#L420):
Here the dataset is splitted into session_T and session_E and the classifier gets trained on session_T and validated on session_E. There exists no test set (the validation set is the test set). This will lead to better results as the "test set" is used to tune the model (Hyperparamters, EarlyStopping, etc.). The final model therefore benefits from the test data during training and the final "test" result is positively biased. Further there is a 2-fold cross validation (same training is repeated with interchanged sessions).

2. [braindecode Example](https://github.com/braindecode/braindecode/blob/master/examples/plot_bcic_iv_2a_moabb_trial.py): same as in 1. without the 2-fold cross-validation.

3. [Schirrmeister et al. 2017 Appendix](https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.23730): Split data into train (session_T) and test (session_E). 2 training phases:
a. train set is splitted into train and validation set (probably a random split?). The model is trained on the train split and the hyperparameters are tuned via the validation set. The final training loss of the best model is saved.
b. The best model is trained with train and validation set (complete session_T) until the training loss reaches the previous best training loss to prevent overfitting. All results are then obtained via the unseen test set (session_E).

4. "Conventional" approach: same as in 3. but without the second training phase.

**Opinion** 
In my opinion either method 3 or 4 should be used to get reasonable results. The braindecode/BCI community would really benefit from a unified implementation of one of these (or both) methods. As method 3 adds additional complexity to method 4 it would be interesting to know, how big the performance boost of method 3 (over method 4) is (@robintibor: is it worth it?). 

What is your opinion on this topic?

agramfort:
[...]

PR #349, opened 2021-10-26
[WIP] Scikit-learn hyperparameter tuning example

gemeinl:
added a scikit-learn hyperparameter tuning example

PR #397, opened 2022-08-04
WIP: added unified validation scheme

martinwimpff:
Added a unified validation scheme example as discussed in  #378 
This tutorial shares some similarities with #349 
@bruAristimunha: let me know what your think about this first draft

### DIFF
--- a/examples/plot_how_train_test_and_tune.py
+++ b/examples/plot_how_train_test_and_tune.py
@@ -241,7 +241,7 @@ from matplotlib.patches import Patch
 
 def plot_simple_train_test(ax, windows_dataset, train_set, test_set):
     """Create a sample plot for training-testing split."""
-    braindecode_cmap = ["#3A6190", "#683E00", "#2196F3", "#DDF2FF"]
+    braindecode_cmap = ["#3A6190", "#683E00", "#DDF2FF", "#2196F3"]
 
     ax.scatter(
         range(len(windows_dataset)),
@@ -265,7 +265,7 @@ def plot_simple_train_test(ax, windows_dataset, train_set, test_set):
         yticks=[0.5, 3.5],
         yticklabels=["Train-Test\nSplit", "Original\nDataset"],
         xlabel="Number of samples.",
-        title="Training-Testing Split",
+        title="Train-Test Split",
     )
 
     ax.legend(
@@ -492,13 +492,13 @@ def plot_k_fold(cv, windows_dataset, X_train, y_train, test_set):
     braindecode_cmap = ["#3A6190", "#683E00", "#2196F3", "#DDF2FF"]
 
     mosaic = """
-      aaa
-      BBC
+      aa
+      BC
       """
 
     axes = plt.figure(figsize=(15, 7), constrained_layout=True).subplot_mosaic(
         mosaic,
-        gridspec_kw={"height_ratios": [1.5, 5], "width_ratios": [3.5, 3.5, 3.5]},
+        gridspec_kw={"height_ratios": [1.5, 5], "width_ratios": [3.5, 3.5]},
     )
 
     # Generate the training/testing visualizations for each CV split
--- a/examples/plot_how_train_test_and_tune.py
+++ b/examples/plot_how_train_test_and_tune.py
@@ -69,21 +69,29 @@ dataset = MOABBDataset(dataset_name="BNCI2014001", subject_ids=[subject_id])
 import numpy as np
 
 from braindecode.preprocessing import (
-    exponential_moving_standardize, preprocess, Preprocessor)
+    exponential_moving_standardize,
+    preprocess,
+    Preprocessor,
+)
 
-low_cut_hz = 4.  # low cut frequency for filtering
-high_cut_hz = 38.  # high cut frequency for filtering
+low_cut_hz = 4.0  # low cut frequency for filtering
+high_cut_hz = 38.0  # high cut frequency for filtering
 # Parameters for exponential moving standardization
 factor_new = 1e-3
 init_block_size = 1000
 
 preprocessors = [
-    Preprocessor('pick_types', eeg=True, meg=False, stim=False),  # Keep EEG sensors
-    Preprocessor(lambda data, factor: np.multiply(data, factor),  # Convert from V to uV
-                 factor=1e6),
-    Preprocessor('filter', l_freq=low_cut_hz, h_freq=high_cut_hz),  # Bandpass filter
-    Preprocessor(exponential_moving_standardize,  # Exponential moving standardization
-                 factor_new=factor_new, init_block_size=init_block_size)
+    Preprocessor("pick_types", eeg=True, meg=False, stim=False),  # Keep EEG sensors
+    Preprocessor(
+        lambda data, factor: np.multiply(data, factor),  # Convert from V to uV
+        factor=1e6,
+    ),
+    Preprocessor("filter", l_freq=low_cut_hz, h_freq=high_cut_hz),  # Bandpass filter
+    Preprocessor(
+        exponential_moving_standardize,  # Exponential moving standardization
+        factor_new=factor_new,
+        init_block_size=init_block_size,
+    ),
 ]
 
 # Transform the data
@@ -98,8 +106,8 @@ from braindecode.preprocessing import create_windows_from_events
 
 trial_start_offset_seconds = -0.5
 # Extract sampling frequency, check that they are same in all datasets
-sfreq = dataset.datasets[0].raw.info['sfreq']
-assert all([ds.raw.info['sfreq'] == sfreq for ds in dataset.datasets])
+sfreq = dataset.datasets[0].raw.info["sfreq"]
+assert all([ds.raw.info["sfreq"] == sfreq for ds in dataset.datasets])
 # Calculate the trial start offset in samples.
 trial_start_offset_samples = int(trial_start_offset_seconds * sfreq)
 
@@ -122,7 +130,7 @@ from braindecode.util import set_random_seeds
 from braindecode.models import ShallowFBCSPNet
 
 cuda = torch.cuda.is_available()  # check if GPU is available, if True chooses to use it
-device = 'cuda' if cuda else 'cpu'
+device = "cuda" if cuda else "cpu"
 if cuda:
     torch.backends.cudnn.benchmark = True
 seed = 20200220
@@ -137,7 +145,7 @@ model = ShallowFBCSPNet(
     n_chans,
     n_classes,
     input_window_samples=input_window_samples,
-    final_conv_length='auto',
+    final_conv_length="auto",
 )
 
 # Send model to GPU
@@ -167,9 +175,9 @@ if cuda:
 #    stage of training or tuning.
 #
 
-splitted = windows_dataset.split('session')
-train_set = splitted['session_T']
-test_set = splitted['session_E']
+splitted = windows_dataset.split("session")
+train_set = splitted["session_T"]
+test_set = splitted["session_E"]
 
 ######################################################################
 # Option 1: Simple Train-Test Split
@@ -209,7 +217,8 @@ clf = EEGClassifier(
     optimizer__weight_decay=weight_decay,
     batch_size=batch_size,
     callbacks=[
-        "accuracy", ("lr_scheduler", LRScheduler('CosineAnnealingLR', T_max=n_epochs - 1)),
+        "accuracy",
+        ("lr_scheduler", LRScheduler("CosineAnnealingLR", T_max=n_epochs - 1)),
     ],
     device=device,
 )
@@ -245,8 +254,8 @@ def plot_simple_train_test(ax, windows_dataset, train_set, test_set):
     ax.scatter(
         range(len(train_set) + len(test_set)),
         [0.5] * len(train_set) + [0.5] * len(test_set),
-        c=[braindecode_cmap[1]] * len(train_set) +
-          [braindecode_cmap[2]] * len(test_set),
+        c=[braindecode_cmap[1]] * len(train_set)
+        + [braindecode_cmap[2]] * len(test_set),
         marker="_",
         lw=50,
     )
@@ -256,24 +265,25 @@ def plot_simple_train_test(ax, windows_dataset, train_set, test_set):
         yticks=[0.5, 3.5],
         yticklabels=["Train-Test\nSplit", "Original\nDataset"],
         xlabel="Number of samples.",
-        title="Training-Testing Split"
+        title="Training-Testing Split",
     )
 
     ax.legend(
-        [Patch(color=braindecode_cmap[0]),
-         Patch(color=braindecode_cmap[1]),
-         Patch(color=braindecode_cmap[2])],
-        ["Original set",
-         "Training set",
-         "Testing set"],
+        [
+            Patch(color=braindecode_cmap[0]),
+            Patch(color=braindecode_cmap[1]),
+            Patch(color=braindecode_cmap[2]),
+        ],
+        ["Original set", "Training set", "Testing set"],
         loc=(1.02, 0.8),
     )
     return ax
 
 
 fig, ax = plt.subplots(figsize=(12, 5))
-plot_simple_train_test(ax=ax, windows_dataset=windows_dataset,
-                       train_set=train_set, test_set=test_set)
+plot_simple_train_test(
+    ax=ax, windows_dataset=windows_dataset, train_set=train_set, test_set=test_set
+)
 fig.tight_layout()
 ######################################################################
 # Option 2: Train-Val-Test Split
@@ -308,7 +318,9 @@ from skorch.helper import predefined_split, SliceDataset
 
 X_train = SliceDataset(train_set, idx=0)
 y_train = np.array([y for y in SliceDataset(train_set, idx=1)])
-train_indices, val_indices = train_test_split(X_train.indices_, test_size=0.2, shuffle=False)
+train_indices, val_indices = train_test_split(
+    X_train.indices_, test_size=0.2, shuffle=False
+)
 train_subset = Subset(train_set, train_indices)
 val_subset = Subset(train_set, val_indices)
 
@@ -329,7 +341,8 @@ clf = EEGClassifier(
     optimizer__weight_decay=weight_decay,
     batch_size=batch_size,
     callbacks=[
-        "accuracy", ("lr_scheduler", LRScheduler('CosineAnnealingLR', T_max=n_epochs - 1)),
+        "accuracy",
+        ("lr_scheduler", LRScheduler("CosineAnnealingLR", T_max=n_epochs - 1)),
     ],
     device=device,
 )
@@ -345,8 +358,8 @@ print(f"Test acc: {(test_acc * 100):.2f}%")
 # Let's visualize the second option with a util function.
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-def plot_train_valid_test(ax, windows_dataset, train_subset,
-                          val_subset, test_set):
+
+def plot_train_valid_test(ax, windows_dataset, train_subset, val_subset, test_set):
     """Create a sample plot for training, validation, testing."""
 
     braindecode_cmap = [
@@ -367,8 +380,8 @@ def plot_train_valid_test(ax, windows_dataset, train_subset,
         range(len(train_subset) + len(val_subset) + len(test_set)),
         [0.5] * len(train_subset) + [0.5] * len(val_subset) + [0.5] * len(test_set),
         c=[braindecode_cmap[1]] * len(train_subset)
-          + [braindecode_cmap[2]] * len(val_subset)
-          + [braindecode_cmap[3]] * len(test_set),
+        + [braindecode_cmap[2]] * len(val_subset)
+        + [braindecode_cmap[3]] * len(test_set),
         marker="_",
         lw=50,
     )
@@ -396,9 +409,13 @@ def plot_train_valid_test(ax, windows_dataset, train_subset,
 
 
 fig, ax = plt.subplots(figsize=(12, 5))
-plot_train_valid_test(ax=ax, windows_dataset=windows_dataset,
-                      train_subset=train_subset, val_subset=val_subset,
-                      test_set=test_set)
+plot_train_valid_test(
+    ax=ax,
+    windows_dataset=windows_dataset,
+    train_subset=train_subset,
+    val_subset=val_subset,
+    test_set=test_set,
+)
 fig.tight_layout()
 
 ######################################################################
@@ -441,17 +458,21 @@ clf = EEGClassifier(
     optimizer__weight_decay=weight_decay,
     batch_size=batch_size,
     callbacks=[
-        "accuracy", ("lr_scheduler", LRScheduler('CosineAnnealingLR', T_max=n_epochs - 1)),
+        "accuracy",
+        ("lr_scheduler", LRScheduler("CosineAnnealingLR", T_max=n_epochs - 1)),
     ],
     device=device,
 )
 
 train_val_split = KFold(n_splits=5, shuffle=False)
-fit_params = {'epochs': n_epochs}
+fit_params = {"epochs": n_epochs}
 cv_results = cross_val_score(
-    clf, X_train, y_train, scoring='accuracy', cv=train_val_split, fit_params=fit_params)
-print(f"Validation accuracy: {np.mean(cv_results * 100):.2f}"
-      f"+-{np.std(cv_results * 100):.2f}%")
+    clf, X_train, y_train, scoring="accuracy", cv=train_val_split, fit_params=fit_params
+)
+print(
+    f"Validation accuracy: {np.mean(cv_results * 100):.2f}"
+    f"+-{np.std(cv_results * 100):.2f}%"
+)
 
 ######################################################################
 # Let's visualize the third option with a util function.
@@ -475,20 +496,16 @@ def plot_k_fold(cv, windows_dataset, X_train, y_train, test_set):
       BBC
       """
 
-    axes = plt.figure(figsize=(15, 7),
-                      constrained_layout=True).subplot_mosaic(
+    axes = plt.figure(figsize=(15, 7), constrained_layout=True).subplot_mosaic(
         mosaic,
-        gridspec_kw={
-            "height_ratios": [1.5, 5],
-            "width_ratios": [3.5, 3.5, 3.5],
-        },
+        gridspec_kw={"height_ratios": [1.5, 5], "width_ratios": [3.5, 3.5, 3.5]},
     )
 
     # Generate the training/testing visualizations for each CV split
     for ii, (tr, tt) in enumerate(cv.split(X=X_train, y=y_train)):
         # Fill in indices with the training/test groups
 
-        axes['a'].scatter(
+        axes["a"].scatter(
             range(len(windows_dataset)),
             [3.5] * len(windows_dataset),
             c=braindecode_cmap[0],
@@ -502,7 +519,7 @@ def plot_k_fold(cv, windows_dataset, X_train, y_train, test_set):
         color_indices = list(map(encode_color, indices))
 
         # Visualize the results
-        axes['B'].scatter(
+        axes["B"].scatter(
             range(len(indices)),
             [ii + 0.5] * len(indices),
             c=color_indices,
@@ -512,7 +529,7 @@ def plot_k_fold(cv, windows_dataset, X_train, y_train, test_set):
             vmax=1.2,
         )
 
-        axes['C'].scatter(
+        axes["C"].scatter(
             range(len(test_set)),
             [ii + 0.5] * len(test_set),
             c=braindecode_cmap[3],
@@ -520,25 +537,22 @@ def plot_k_fold(cv, windows_dataset, X_train, y_train, test_set):
             lw=10,
         )
 
-    axes['a'].set(
-        yticklabels=[''],
+    axes["a"].set(
+        yticklabels=[""],
         xlim=[0, len(windows_dataset) + 1],
         ylabel="Original\nData",
         ylim=[3.4, 3.6],
     )
-    axes['a'].yaxis.get_label().set_fontsize(16)
+    axes["a"].yaxis.get_label().set_fontsize(16)
 
-    axes['C'].set(
-        yticks=np.arange(5) + 0.5,
-        yticklabels=[''] * 5,
-        xlim=[0, 300],
-        ylim=[5, -0.2],
+    axes["C"].set(
+        yticks=np.arange(5) + 0.5, yticklabels=[""] * 5, xlim=[0, 300], ylim=[5, -0.2],
     )
 
     # Formatting
     yticklabels = list(range(5))
 
-    axes['B'].set(
+    axes["B"].set(
         yticks=np.arange(5) + 0.5,
         yticklabels=yticklabels,
         ylabel="CV iteration",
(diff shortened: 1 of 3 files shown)


bruAristimunha:
Hi @martinwimpff,

Thank you so much for your PR!

Here is the render version: 
https://output.circle-artifacts.com/output/job/a34faa77-933e-41f7-8a6f-4905c1fe2ef4/artifacts/0/dev/auto_examples/plot_unified_validation_scheme.html#sphx-glr-auto-examples-plot-unified-validation-scheme-py

I am still revising it, but it is already pretty cool! I will try to add a figure to make it easier to differentiate the three schemes.

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+    optimizer__weight_decay=weight_decay,
+    batch_size=batch_size,
+    callbacks=[
+        "accuracy", ("lr_scheduler", LRScheduler('CosineAnnealingLR', T_max=n_epochs - 1)),
+    ],
+    device=device,
+)
+
+######################################################################
+# We will make use of the sklearn library to do the Hyperparameter
```
sliwy, line 268:
```suggestion
# We will make use of the sklearn library to do the hyperparameter
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+# Split dataset into train and valid
+# ----------------------------------
+#
+
+
+######################################################################
+# We can easily split the dataset using additional info stored in the
+# description attribute, in this case ``session`` column. We select
+# ``session_T`` for training and ``session_E`` for validation.
+# For other datasets you might have to choose another column.
```
bruAristimunha, line 117:
```suggestion
# We can easily split the dataset using additional info stored in the 
# description attribute, in this case, the  ``session`` column. We select 
# ``session_T``  for training and  ``session_E`` for validation.
# For other datasets, you might have to choose another column.
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+import pandas as pd
+from sklearn.model_selection import GridSearchCV, KFold, train_test_split
+from skorch.helper import SliceDataset
+
+X_train = SliceDataset(train_set, idx=0)
+y_train = np.array([y for y in SliceDataset(train_set, idx=1)])
+train_val_split = [tuple(train_test_split(X_train.indices_, test_size=0.2, shuffle=False))]
+
+######################################################################
+# Define the ``fit_params`` and the ``paramter_grid`` i.e. list all
```
sliwy, line 290:
```suggestion
# Define the ``fit_params`` and the ``parameter_grid`` i.e. list all
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+# Hyperparamters you want to include in your search.
```
sliwy, line 291:
```suggestion
# hyperparameters you want to include in your search.
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+# ---------------------------------
+#
+
+
+######################################################################
+# This is the easiest training scheme to use as the dataset is only
+# splitted in two distinct sets (``train_set`` and ``test_set``).
+# As this method uses no separate validation split it should only be
+# used for the final evaluation of the (previously!) found
+# Hyperparameter configuration.
```
bruAristimunha, line 175:
```suggestion
# This is the easiest training scheme to use as the dataset is only 
# split into two distinct sets (``train_set`` and ``test_set``).
# This scheme uses no separate validation split and should only be
#  used for the final evaluation of the (previously!) found
#  Hyperparameter configuration.
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+#
+# .. warning::
+#    If you make any use of the ``test_set`` during training
+#    (i.e. by using EarlyStopping) there will be data leakage
+#    which will make the reported generalization capability/decoding
+#    performance of your model less credible.
```
bruAristimunha, line 181:
```suggestion
#    If you use the  ``test_set`` during training
#    (i.e. by using EarlyStopping), there will be data leakage
#    which will make your model's reported generalization capability/decoding
#    performance less credible.
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+# By averaging over the k validation scores afterwards, you get a
+# very reliable estimate of how the model would perform on unseen
+# data (test set).
+# The implementation is straight forward as we only have to change one
+# line of code compared to Option 2. The ``KFold`` class will split
+# the ``train_set`` into k folds. We will use 5 folds to get a 80-20
+# train-val split.
+#
+
+train_val_split = KFold(n_splits=5, shuffle=False)
```
sliwy, line 344: maybe you can also show how to evaluate model on cross-validation with `cross_val_score` - should be one line somewhere, just to show that it can be also used as performance evaluation method and not only hyperparams search

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+
+low_cut_hz = 4.  # low cut frequency for filtering
+high_cut_hz = 38.  # high cut frequency for filtering
+# Parameters for exponential moving standardization
+factor_new = 1e-3
+init_block_size = 1000
+
+preprocessors = [
+    Preprocessor('pick_types', eeg=True, meg=False, stim=False),  # Keep EEG sensors
+    Preprocessor(scale, factor=1e6, apply_on_array=True),  # Convert from V to uV
```
sliwy, line 73: I would use `np.multiply` here to avoid `DeprecationWarning` in the doc

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+# the best (or at least a suitable) Hyperparameter configuration makes
+# up a substantial part of the developement process.
+# As stated above, it is not suitable to use the ```test_set``
+# for this Hyperparamter search. Therefore we need a third split, the
+# so called validation set which is a Subset of the ``train_set``.
+# This second option splits the original ``train_set`` only once
+# (instead of k times as in Option 3) to speed up the tuning process.
+# This method should only be preferred over Option 3 if either the
+# training duration is very long or the Hyperparamter search space is
+# very large.
```
bruAristimunha, line 232:
```suggestion
# Usually, when developing a new deep learning model/method finding 
# the best (or at least a suitable) hyperparameter configuration makes 
# up a substantial part of the development process. 
# 
#As stated above, using the ```test_set`` 
# for this hyperparameter search is unsuitable. Therefore we need a third split, 
# the so-called validation set, which is a Subset of the train_set. # This second scheme splits the original train_set only once 
# (instead of k times as in scheme 3) to speed up the tuning process. 
#
# This method should only be preferred over scheme 3 if the 
# training duration is very long, or the Hyperparameter search space # is very large.
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
+"""
+Unified Validation Scheme
+========================================
+
+This tutorial shows you how to properly train, tune and test your deep learning
+models with Braindecode. We will use the BCIC IV 2a dataset as a showcase example,
+however this scheme holds for any supervised trial-based decoding setting.
+As this tutorial will include additional parts of code like loading and preprocessing,
+defining a model, etc. which are not exclusive to this page (compare `Cropped Decoding
+Tutorial <./plot_bcic_iv_2a_moabb.html>`__), feel free to skip these parts.
```
sliwy, line 10: I would link to trialwise example. This also fixes the link to the file.
```suggestion
defining a model, etc. which are not exclusive to this page (compare `Trialwise Decoding
Tutorial <./plot_bcic_iv_2a_moabb_trial.html>`__), feel free to skip these parts.
```

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+# To compensate this, one can run a k-fold Cross Validation, where
+# every sample of the training set is in the validation set once.
+# By averaging over the k validation scores afterwards, you get a
+# very reliable estimate of how the model would perform on unseen
+# data (test set).
+# The implementation is straight forward as we only have to change one
+# line of code compared to Option 2. The ``KFold`` class will split
+# the ``train_set`` into k folds. We will use 5 folds to get a 80-20
+# train-val split.
+#
```
bruAristimunha, line 342:
```suggestion
######################################################################
# As mentioned above, using only one validation split might not be
# sufficient, as there might be a shift in the data distribution.
# To compensate for this, one can run a k-fold Cross Validation, where
# every sample of the training set is in the validation set once.
# After averaging over the k validation scores, you get a
# very reliable estimate of how the model would perform on unseen
# data (test set).
# The implementation is straightforward as we only have to change one
# line of code compared to Option 2. The ``KFold`` class will split
# the ``train_set`` into k folds. We will use 5 folds to get an 80-20
# train-validation split.
#
```

sliwy (changes requested):
Great job @martinwimpff @bruAristimunha! To me the description is clear and I hope people will read it before doing ML for EEG. My comments are mainly typos and some reformulations.

A visualization of the CV schemes would be highly appreciated.

One more thing, now this tutorial and [`plot_hyperparameter_tuning_with_scikit-learn.py` ](https://github.com/braindecode/braindecode/blob/66ddda17875bb4d5b788228dd7a3b0b6117290b8/examples/plot_hyperparameter_tuning_with_scikit-learn.py) are quite similar. Maybe we can add `hyperparameters search` to the title of this one and just keep this tutorial? What do you think?

bruAristimunha (commented):
Is very good! I tried to give some small suggestions, I tried not to change the text. Feel free to accept or not the suggestions. I'll try to put some image to illustrate the options.

bruAristimunha:
Hi @martinwimpff,

At a tutorial, I was wondering if we could add:

schema_train_test

It's a vertorized image, so possibly it will be bigger in the tutorial.

examples/plot_unified_validation_scheme.py
martinwimpff: Good point - I will add that

examples/plot_unified_validation_scheme.py
martinwimpff: Agree

examples/plot_unified_validation_scheme.py
```diff
@@ -0,0 +1,344 @@
...
+
+######################################################################
+# Usually when developing a new Deep Learning model/method finding
+# the best (or at least a suitable) Hyperparameter configuration makes
+# up a substantial part of the developement process.
+# As stated above, it is not suitable to use the ```test_set``
+# for this Hyperparamter search. Therefore we need a third split, the
+# so called validation set which is a Subset of the ``train_set``.
+# This second option splits the original ``train_set`` only once
+# (instead of k times as in Option 3) to speed up the tuning process.
```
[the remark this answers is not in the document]
martinwimpff: We can do that but Robins method was a little different. He used Option 2 for tuning and Option 1 for the final evaluation (to use the full training set).

examples/plot_unified_validation_scheme.py
martinwimpff: Do you know how to properly do that? The problem with `np.multiply` is that x2 (the factor) is a positional argument and not a keyword argument. Therefore
`Preprocessor(np.multiply, x2=1e6, apply_on_array=True)`
does not work. My current workaround would look like this:
`Preprocessor(lambda data, factor: np.multiply(data, factor), factor=1e6)`

martinwimpff:
@sliwy: first of all, thanks a lot for your review!
> One more thing, now this tutorial and [`plot_hyperparameter_tuning_with_scikit-learn.py` ](https://github.com/braindecode/braindecode/blob/66ddda17875bb4d5b788228dd7a3b0b6117290b8/examples/plot_hyperparameter_tuning_with_scikit-learn.py) are quite similar. Maybe we can add `hyperparameters search` to the title of this one and just keep this tutorial? What do you think?

I agree with that, there shouldn't be two tutorials. The new title could be something like: How to train, test and tune your model. But I am open for your suggestions ;)

As you suggested I added additional parts for option 2 and 3 that do not use GridSearch.

@bruAristimunha: thanks for your suggestions as well, I added them as well.
Regarding the visualization, I also agree with both of you, we should include that. The image you posted looks really nice but in my opinion, something simpler would probably make the point clearer. Here is a quick example of what I had in mind:

CV

agramfort:
I would produce images with matplotlib like sklearn does to demo the cross validation generators

examples/plot_unified_validation_scheme.py
sliwy: good question! :smile:  I just followed the DeprecationWarning message. I think we should remove the DeprecationWarning and abandon the plan of removing `scale` function because in our flow `np.multiply` can't replace it. At least I don't see an easy way as we use keyword arguments in preprocessing (the lambda you've proposed does exactly the same as current `scale` function). 

We may remove the DeprecationWarning here or in another PR if everyone agrees that it won't work with `np.multiply` @robintibor @hubertjb @bruAristimunha

sliwy:
@martinwimpff thanks for the corrections!

`How to train, test and tune your model` - works for me :)

Having visualization done with matplotlib and sklearn could be also awesome.

Can we find some names for CV options (it is easier to use outside of tutorial context), my proposition:
1. Option 1 - train-test split
2. Option 2 - train-valid-test split
3. Option 3 - k-fold blockwise CV
In the case inside k-fold blockwise CV there is another one for hyperparams this could be indicated with adding 'nested'?

@martinwimpff For the option 3 visualization I would add `test_set` as a supplementary option because in some articles CV result is the final one while hyperparams are optimized in a nested CV. 

I took a while (quite short, so I just scanned paper for CV method) to prepare summary of evaluation methods used for models that are currently available in braindecode. Maybe it can be useful for discussion and showing how it is used in papers. Feel free to include it in the tutorial if you think it may be useful. I feel that it would be nice if we can express all the methods from this table in terms of our evaluation tutorial terminology. What do you think?

|          Article          |            Models             |    Hyperparameters optimization    |               Final evaluation                |
| :-----------------------: | :---------------------------: | :--------------------------------: | :-------------------------------------------: |
| Schirrmeister et al. 2017 | ShallowFBSCPNet<br />Deep4Net |              Option 2              |                   Option 1                    |
|    Lawhern et al. 2018    |            EEGNet             |                n/a                 | Option 3 (hold-out test only for SMR dataset) |
|    Chambon et al. 2018    |    SleepStagerChambon2018     |              Nested option 3              |        Option 3, no hold-out test set         |
|    Eldele et al. 2021     |     SleepStagerEldele2021     | Option 3 (without hold-out test)  |       Option 3 (without hold-out test)        |
|      Perslev et al.       |            USleep             |              Option 1              |           Option 1 on hold-out test           |



Schirrmeister, R.T., Springenberg, J.T., Fiederer, L.D.J., Glasstetter, M., Eggensperger, K., Tangermann, M., Hutter, F., Burgard, W. and Ball, T. (2017), Deep learning with convolutional neural networks for EEG decoding and visualization. Hum. Brain Mapp., 38: 5391-5420. https://doi.org/10.1002/hbm.23730

Lawhern, V.J., Solon, A.J., Waytowich, N.R., Gordon, S.M., Hung, C.P., & Lance, B. (2018). EEGNet: a compact convolutional neural network for EEG-based brain–computer interfaces. *Journal of Neural Engineering, 15*.

Chambon, S., Galtier, M.N., Arnal, P.J., Wainrib, G., & Gramfort, A. (2018). A Deep Learning Architecture for Temporal Sleep Stage Classification Using Multivariate and Multimodal Time Series. *IEEE Transactions on Neural Systems and Rehabilitation Engineering, 26*, 758-769.

Eldele, E., Chen, Z., Liu, C., Wu, M., Kwoh, C., Li, X., & Guan, C. (2021). An Attention-Based Deep Learning Approach for Sleep Stage Classification With Single-Channel EEG. *IEEE Transactions on Neural Systems and Rehabilitation Engineering, 29*, 809-818.

Perslev, M., Darkner, S., Kempfner, L., Nikolic, M., Jennum, P.J., & Igel, C. (2021). U-Sleep: resilient high-frequency sleep staging. *NPJ Digital Medicine, 4*.

martinwimpff:
> ``How to train, test and tune your model`` - works for me :)

👍 

> Can we find some names for CV options (it is easier to use outside of tutorial context), my proposition:
>
>1. Option 1 - train-test split
>2. Option 2 - train-valid-test split
>3. Option 3 - k-fold blockwise CV
In the case inside k-fold blockwise CV there is another one for hyperparams this could be indicated with adding 'nested'?

Agree. I am not sure about the nested CV. What we do in option 3 is a k-fold blockwise CV (with a fixed Hold-Out set, the untouched test set). As far as I understand this should be used with BCI research as the test set should always be from an unseen session or subject. Further, most datasets provide this two-fold split anyways. 
With nested CV one would throw everything in one data set and run two loops: one outer loop for splitting out the test set and one inner loop for the validation set. I don't really see a benefit from doing that for BCI research, as one would have to be extremely careful with how to design the splitters and the improvement is [quite marginal](https://scikit-learn.org/stable/auto_examples/model_selection/plot_nested_cross_validation_iris.html#sphx-glr-auto-examples-model-selection-plot-nested-cross-validation-iris-py).

> @martinwimpff For the option 3 visualization I would add test_set as a supplementary option because in some articles CV result is the final one while hyperparams are optimized in a nested CV.

I actually had the same thought while creating the image. Maybe we have to make the **division between HP search and "normal" training with 3 splits clearer**. For HP tuning the test set should never be used. 
For option 2 it may be used if you need to use the validation set for i.e. EarlyStopping. Else (if all HPs are fixed) option 1 should be used. 
The same holds for option 3 with the addition that the final test score is the average over k test scores.

We can include the table that you created in something like an additional section that we call `Reporting` or something like that? There we could also highlight the (large) differences between the publications + the occasional lack of documentation regarding the tuning/evaluation to further motivate why this page was created.

agramfort:
To avoid confusing the community I would aim to follow sklearn naming conventions and refer to sklearn documentation as much as possible. Maintaining a good documentation takes energy so let’s put our energy where we are very unique

examples/plot_unified_validation_scheme.py
bruAristimunha: Sure @sliwy. After finishing this tutorial we will do a PR to update the warning message.

bruAristimunha:
> I would produce images with matplotlib like sklearn does to demo the cross validation generators

Something like that @agramfort?

https://scikit-learn.org/stable/auto_examples/model_selection/plot_cv_indices.html

bruAristimunha:
> gree. I am not sure about the nested CV. What we do in option 3 is a k-fold blockwise CV (with a fixed Hold-Out set, the untouched test set). As far as I understand this should be used with BCI research as the test set should always be from an unseen session or subject. Further, most datasets provide this two-fold split anyways. With nested CV one would throw everything in one data set and run two loops: one outer loop for splitting out the test set and one inner loop for the validation set. I don't really see a benefit from doing that for BCI research, as one would have to be extremely careful with how to design the splitters and the improvement is [quite marginal](https://scikit-learn.org/stable/auto_examples/model_selection/plot_nested_cross_validation_iris.html#sphx-glr-auto-examples-model-selection-plot-nested-cross-validation-iris-py).
> 
> > @martinwimpff For the option 3 visualization I would add test_set as a supplementary option because in some articles CV result is the final one while hyperparams are optimized in a nested CV.
> 
> I actually had the same thought while creating the image. Maybe we have to make the **division between HP search and "normal" training with 3 splits clearer**. For HP tuning the test set should never be used. For option 2 it may be used if you need to use the validation set for i.e. EarlyStopping. Else (if all HPs are fixed) option 1 should be used. The same holds for option 3 with the addition that the final test score is the average over k test scores.
> 
> We can include the table that you created in something like an additional section that we call `Reporting` or something like that? There we could also highlight the (large) differences between the publications + the occasional lack of documentation regarding the tuning/evaluation to further motivate why this page was created.

bruAristimunha:
> > `How to train, test and tune your model` - works for me :)
> 
+2. I love the new title =)
> 
> > Can we find some names for CV options (it is easier to use outside of tutorial context), my proposition:
> > 
> > 1. Option 1 - train-test split
> > 2. Option 2 - train-valid-test split
> > 3. Option 3 - k-fold blockwise CV
> >    In the case inside k-fold blockwise CV there is another one for hyperparams this could be indicated with adding 'nested'?
> 
> Agree. I am not sure about the nested CV. What we do in option 3 is a k-fold blockwise CV (with a fixed Hold-Out set, the untouched test set). As far as I understand this should be used with BCI research as the test set should always be from an unseen session or subject. Further, most datasets provide this two-fold split anyways. With nested CV one would throw everything in one data set and run two loops: one outer loop for splitting out the test set and one inner loop for the validation set. I don't really see a benefit from doing that for BCI research, as one would have to be extremely careful with how to design the splitters and the improvement is [quite marginal](https://scikit-learn.org/stable/auto_examples/model_selection/plot_nested_cross_validation_iris.html#sphx-glr-auto-examples-model-selection-plot-nested-cross-validation-iris-py).

> > @martinwimpff For the option 3 visualization I would add test_set as a supplementary option because in some articles CV result is the final one while hyperparams are optimized in a nested CV.
> 
> I actually had the same thought while creating the image. Maybe we have to make the **division between HP search and "normal" training with 3 splits clearer**. For HP tuning the test set should never be used. For option 2 it may be used if you need to use the validation set for i.e. EarlyStopping. Else (if all HPs are fixed) option 1 should be used. The same holds for option 3 with the addition that the final test score is the average over k test scores.
> 
> We can include the table that you created in something like an additional section that we call `Reporting` or something like that? There we could also highlight the (large) differences between the publications + the occasional lack of documentation regarding the tuning/evaluation to further motivate why this page was created.

I love the idea of a reporting subsection to discuss a little the potential difference that we can find in metrics reported in articles and during the replication moment. I think you just passed this with deep4net, right @martinwimpff?

I would love this table in my thesis, but not in the library. With more deep learning methods, it may become unfeasible to maintain the table.

I wonder if we're putting too much information into a tutorial. I would suggest that we don't spend too much energy on definitions, as indicated by @agramfort, and close with a simple matplotlib view, like https://scikit-learn.org/stable/auto_examples/model_selection/plot_cv_indices.html. What do you think, @martinwimpff, @sliwy and @agramfort?

martinwimpff:
> To avoid confusing the community I would aim to follow sklearn naming conventions and refer to sklearn documentation as much as possible. Maintaining a good documentation takes energy so let’s put our energy where we are very unique

I think that this is a valid point and it might be better if we reorganize the content of this tutorial, also regarding the merge with the current hyperparameter tuning tutorial.

So, as mentioned above, I would like to make the division between "normal" training and HP tuning clearer. Therefore I would propose the following struture:

1. Introduction (motivate the topic)
2. How to train and evaluate your model
2.1 Simple Train-Test Split
2.2 Train-Val-Test Split
2.3 (blockwise) k-Fold CV (with Holdout test set)
3. Hyperparameter Tuning
3.1 Train-Val(-Test) Split
3.2 (blockwise) k-Fold CV
4. Reporting

As @agramfort mentioned, we should reference the sklearn library as often as possible (especially in sections 2 and 3). Further, the explanations of the schemes should be minimal and rather focus on the specifics of BCI data (e.g. test set should be from another session or subject, train-val split should be blockwise, is there anything else?). Is this what you meant with too much information @bruAristimunha ?

bruAristimunha:
Yes @martinwimpff

sliwy:
@bruAristimunha I agree, the table may be too much for the tutorial. The matplotlib simple plots look good to me! :)

@martinwimpff I like the new structure of the tutorial. Just for point 2.3 I would mention that it can be used as well without hold-out test set.

martinwimpff:
I just committed a new version with the restructuring mentioned above.
@sliwy @bruAristimunha what do you think of this?
@bruAristimunha How should we proceed with the visualizations? Have you already had a look into it?

bruAristimunha:
Hello @martinwimpff and @sliwy,
I have included three visualizations. Can you look and review it?

bruAristimunha (approved):

bruAristimunha (approved):

martinwimpff:
Great job @bruAristimunha!

I have two small suggestions for the visualizations:
1. I would keep the colors consistent throughout the images (currently the test set color changes from 1 to 2&3)
2. I would keep the scaling for option 3 (x-axis ratio) s.t. its clearer that the test set (size) remains the same in all 3 options. You probably wanted to make the kFold visualization as big as possible + clearly separate the test set. So this one is probably a matter of taste ;)

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+        marker="_",
+        lw=50,
+    )
+
+    ax.set(
+        ylim=[-1, 5],
+        yticks=[0.5, 3.5],
+        yticklabels=["Train-Test\nSplit", "Original\nDataset"],
+        xlabel="Number of samples.",
+        title="Training-Testing Split",
```
sliwy, line 268:
```suggestion
        title="Train-Test Split",
```

sliwy (approved):
Great job @martinwimpff @bruAristimunha! Thanks for that!

The only small thing is that the code for generating the visualization complicates the tutorial a little bit, but I don't have any idea how to fix that apart from moving those functions to some module. I am not a big fan of this option either.

To me, after the small changes in the visualization mentioned by @martinwimpff it is ready to be merged! :)

bruAristimunha:
I have incorporated all comments. For me, we can apply merge, @agramfort.

sliwy:
@bruAristimunha I almost forgot, do we finally remove the [plot_hyperparameter_tuning_with_scikit-learn.py](https://github.com/braindecode/braindecode/blob/66ddda17875bb4d5b788228dd7a3b0b6117290b8/examples/plot_hyperparameter_tuning_with_scikit-learn.py)?

bruAristimunha:
Can we handle this in another pull request?

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+models with Braindecode. We will use the BCIC IV 2a dataset as a showcase example.
+
+The methods shown can be applied to any standard supervised trial-based decoding setting.
+This tutorial will include additional parts of code like loading and preprocessing,
+defining a model, and other details which are not exclusive to this page (compare
+`Cropped Decoding Tutorial <./plot_bcic_iv_2a_moabb_trial.html>`__). Therefore we
+will not further elaborate on these parts and you can feel free to skip them.
+
+In general we distinguish between "usual" training and evaluation and hyperparameter search.
+The Tutorial is therefore split into two parts, one for the three different training schemes
```
agramfort, line 15:
```suggestion
The tutorial is therefore split into two parts, one for the three different training schemes
```

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+and one for the two different hyperparameter tuning methods.
+
+"""
+
+######################################################################
+# Why should I care about model evaluation?
+# -----------------------------------------
+# Short answer: To produce reliable results!
```
agramfort, line 23:
```suggestion
# Short answer: Yes to produce reliable results!
```

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+# data into two parts, training and testing sets. It sounds like a
+# simple division, right? But the story does not end here.
+#
+# While developing a ML model you
+# usually have to adjust and tune hyperparameters of your model or
+# pipeline (e.g., number of layers, learning rate, number of epochs).
+# Deep learning models usually have many free parameters; they could
+# be considered complex models with many degrees of freedom.
+# If you kept using the test dataset to evaluate your adjustment
+# you would run into data leakage.
```
agramfort, line 35: can you format this to leverage the full authorized line width (80c ?)

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+#
+# This means that if you use the test set to adjust the hyperparameters
+# of your model, the model implicitly learns or memorizes the test set.
+# Therefore, the trained model is no longer independent of the test set
+# (even though they were never used for backpropagation!).
```
agramfort, line 40:
```suggestion
# (even though they were never used for training!).
```

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+# If you perform any Hyperparameter tuning, you need a third split,
```
agramfort, line 41:
```suggestion
# If you perform any hyperparameter tuning, you need a third split,
```

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+cuda = torch.cuda.is_available()  # check if GPU is available, if True chooses to use it
+device = "cuda" if cuda else "cpu"
+if cuda:
+    torch.backends.cudnn.benchmark = True
+seed = 20200220
+set_random_seeds(seed=seed, cuda=cuda)
+
+n_classes = 4
+# Extract number of chans and time steps from dataset
+n_chans = windows_dataset[0][0].shape[0]
```
agramfort, line 141: for later but n_channels would be nicer for me

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+
+from torch.utils.data import Subset
+from sklearn.model_selection import train_test_split
+from skorch.helper import predefined_split, SliceDataset
+
+X_train = SliceDataset(train_set, idx=0)
+y_train = np.array([y for y in SliceDataset(train_set, idx=1)])
+train_indices, val_indices = train_test_split(
+    X_train.indices_, test_size=0.2, shuffle=False
+)
```
agramfort, line 323: this is neat but a bit convoluted for an intro tutorial.

I would consider exposing in train_set indices or something that can simply
be passed to train_test_split

sliwy:
@bruAristimunha sure!

Once again, thanks for this great tutorial! 👏

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+#    This k-Fold Cross Validation can be used without a separate
+#    (holdout) test set. If there is no test set available, e.g. in a
+#    competition, this scheme is highly recommended to get a reliable
+#    estimate of the generalization performance.
+#
+# To implement this, we will make use of sklearn function
+# `cross_val_score <https://scikit-learn.org/stable/modules/generated/
+# sklearn.model_selection.cross_val_score.html>`__ and the `KFold
+# <https://scikit-learn.org/stable/modules/generated/sklearn.model_
+# selection.KFold.html>`__. cross-validator.
```
agramfort, line 445: its not called a cross-validator in sklearn but a cross-validation splitter

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+
+plot_k_fold(
+    cv=train_val_split,
+    windows_dataset=windows_dataset,
+    X_train=X_train,
+    y_train=y_train,
+    test_set=test_set,
+)
+######################################################################
+# How to tune your Hyperparameters
```
agramfort, line 588:
```suggestion
# How to tune your hyperparameters
```

examples/plot_how_train_test_and_tune.py
```diff
@@ -0,0 +1,669 @@
...
+#
+
+######################################################################
+# To perform a full k-Fold CV just replace ``train_val_split`` from
+# above with the ``KFold`` cross-validator from sklearn.
+
+train_val_split = KFold(n_splits=5, shuffle=False)
+
+######################################################################
+# .. note::
```
agramfort, line 662: this is something that should appear at the top and maybe as a warning as it's not a detail.

examples/plot_how_train_test_and_tune.py
martinwimpff: One could use
```
train_test_split(np.array(len(train_set)), test_size=0.2, shuffle=False)
```
instead, but `X_train` and `y_train` would still be necessary for `cross_val_score, search.fit()` and the visualizations.

examples/plot_how_train_test_and_tune.py
martinwimpff: I had the same thought about that one as they call it CV generator or CV splitter in general but cross-validator
In the [KFold doc](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html).
The class itself also inherits from `BaseCrossValidator`. 
How should we call it?

agramfort:
arff I would use the word CV splitter

examples/plot_how_train_test_and_tune.py
sliwy: maybe we can use `copy.deepcopy(model)` in every `EEGClassifier` init, so we won't modify the weights of original model?

martinwimpff: I prefer the warning option. If we use `deepcopy` we would have to include a small note on that as well to avoid confusion.

martinwimpff:
I included the comments from @agramfort.

I did not change this part:
```python
from torch.utils.data import Subset
from sklearn.model_selection import train_test_split
from skorch.helper import predefined_split, SliceDataset

X_train = SliceDataset(train_set, idx=0)
y_train = np.array([y for y in SliceDataset(train_set, idx=1)])
train_indices, val_indices = train_test_split(
    X_train.indices_, test_size=0.2, shuffle=False
)
```
as `X_train` and `y_train` are used multiple times later in the tutorial. 

I changed the final note to a warning and moved it to the start as I personally prefer this option for the tutorial. I also like the `deepcopy` option in general (since it it a very common thing to do) but I think that it might confuse people in this tutorial.

agramfort (approved):
looks good !

https://output.circle-artifacts.com/output/job/218ce5d3-5b86-4094-a596-ddce84388a1a/artifacts/0/dev/auto_examples/plot_how_train_test_and_tune.html

agramfort:
thanks @martinwimpff @bruAristimunha 🙏

bruAristimunha:
A huge thank you to you, @martinwimpff. For building the entire tutorial! Welcome to the braindecode contributor group, and we look forward to your next contributions! Always feel welcome to contribute and interact with the library.

robintibor:
Also wanted to say a big thank you to @martinwimpff  @bruAristimunha and also @agramfort the tutorial really looks incredible, really a valuable addition to the library. I had been a bit absent and feel super happy to see such great work once I found the time to look haha :)

PR #398, opened 2022-08-11
Fix torch.linalg issue in rotation augmentations

cedricrommel:
This commit fixes an issue encountered in the implementation of
augmentations using the `sensors_rotation` functional. The latter comes
from a bug that appeared in `torch.linalg.pinv` function since its
reimplementation for v1.8.0 release. The v1.8.0 implementation indeed
led too often in the failing of computation of the pseudo-inverse.
As a results, RandomSensorsRotations augmentations were incorrect at times
and even irreproducible (since pinv fails don't depend on controllable PRNGs).

The fix I propose in this commit follows recommendations in [this torch issue](https://github.com/pytorch/pytorch/issues/75494) and consists in:
1. trying to use regular inverse whenever possible (since the matrix C
   in `_torch_make_interpolation_matrix` is supposed to be invertible
   quite often).
2. falling back to pinv if C is not invertible, but doing the
   computation in cpu (which is more reliable according to the issue
   above).

### DIFF
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -840,7 +840,6 @@ def _torch_make_interpolation_matrix(pos_from, pos_to, alpha=1e-5):
     try:
         C_inv = torch.linalg.inv(C)
     except torch._C._LinAlgError:
-        print("Had to fallback to pinv for one batch.")
         # There is a stability issue with pinv since torch v1.8.0
         # see https://github.com/pytorch/pytorch/issues/75494
         C_inv = torch.linalg.pinv(C.cpu()).to(device)
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -836,7 +836,14 @@ def _torch_make_interpolation_matrix(pos_from, pos_to, alpha=1e-5):
                 torch.ones((1, n_from), device=device),
                 torch.as_tensor([[0]], device=device)])
         ])
-    C_inv = torch.linalg.pinv(C)
+
+    try:
+        C_inv = torch.linalg.inv(C)
+    except torch._C._LinAlgError:
+        print("Had to fallback to pinv for one batch.")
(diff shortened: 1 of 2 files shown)


agramfort (approved):

bruAristimunha:
Maybe this is related to a test I had to remove in another PR. [Commit Link](https://github.com/braindecode/braindecode/commit/337d4c807c4f46b11107ffcf19601f57599091b6#diff-b7d7f2126b8c604634b1867a65dbb878330b239642060143a8ee369c9fe13a02)

In the file: test/unit_tests/augmentation/test_transforms.py
`(SensorsZRotation, 15, False, True),`

cedricrommel:
> Maybe this is related to a test I had to remove in another PR. [Commit Link](https://github.com/braindecode/braindecode/commit/337d4c807c4f46b11107ffcf19601f57599091b6#diff-b7d7f2126b8c604634b1867a65dbb878330b239642060143a8ee369c9fe13a02)
> 
> In the file: test/unit_tests/augmentation/test_transforms.py `(SensorsZRotation, 15, False, True),`

Indeed, just checked and it was that :) Just pushed a commit adding the test back. Thanks @bruAristimunha

cedricrommel:
I really don't understand why the test is not passing on GHActions. I error is related to torch.autograd. I have exactly the same version of torch and when running tests on CPU they all pass on my machine:
Capture d’écran 2022-09-01 à 17 52 30

cedricrommel:
Given that I don't currently have time to investigate why that test passes on my machine and not on GH and that it is a "nice to have" test, I am just reverting the last commit.

agramfort (approved):
@cedricrommel does it slow down significantly computations for you?

cedricrommel:
> @cedricrommel does it slow down significantly computations for you?

What do you mean @agramfort ? I did not notice slower computations, but I did not look for it either.

Issue #377, opened 2022-04-13
TUHAbnormal Needs to be updated in braindecode/examples/benchmark_lazy_eager_loading.py


vamsi231297:
In the file **braindecode/examples/benchmark_lazy_eager_loading.py**. The TUHAbnormal class initialization needs to be updated

```
ds = TUHAbnormal(
        TUH_PATH, subject_ids=subject_ids, target_name='pathological',
        preload=preload)```

In the latest TUHAbnormal version there is no **subject_ids** argument

gemeinl:
Hi @kuwar81523,
thanks for reporting.
Would you mind creating a PR to fix this?
Thanks.

bruAristimunha:
Hi @kuwar81523,

I was wondering if you had the opportunity to consider @gemeinl' proposal.

kuwar81523:
Yeah sure @bruAristimunha :)

kuwar81523:
Sure @bruAristimunha

bruAristimunha:
We look forward to your pull request, @kuwar81523  =) If you need any help please let me know.

PierreGtch:
PR merged

PR #402, opened 2022-08-27
 Removing *subject_ids* argument in TUHAbnormal example.

bruAristimunha:
Related with the issue #377.

Removing a deprecated parameter in the TUHAbnormal example.

### DIFF
--- a/examples/benchmark_lazy_eager_loading.py
+++ b/examples/benchmark_lazy_eager_loading.py
@@ -90,7 +90,7 @@ def load_example_data(preload, window_len_s, n_subjects=10):
     """
     subject_ids = list(range(n_subjects))
     ds = TUHAbnormal(
-        TUH_PATH, subject_ids=subject_ids, target_name='pathological',
+        TUH_PATH, target_name='pathological',
         preload=preload)
 
     fs = ds.datasets[0].raw.info['sfreq']
--- a/examples/benchmark_lazy_eager_loading.py
+++ b/examples/benchmark_lazy_eager_loading.py
@@ -88,7 +88,6 @@ def load_example_data(preload, window_len_s, n_subjects=10):
         sampling rate. The following assumes that the files have already been
         resampled to a common sampling rate.
     """
-    subject_ids = list(range(n_subjects))
     ds = TUHAbnormal(
         TUH_PATH, target_name='pathological',
         preload=preload)
--- a/examples/benchmark_lazy_eager_loading.py
+++ b/examples/benchmark_lazy_eager_loading.py
@@ -68,7 +68,7 @@ torch.set_num_threads(N_JOBS)  # Sets the available number of threads
 # Each one of these steps will be timed, so we can report the total time taken
 # to prepare the data and train the model.
 
-def load_example_data(preload, window_len_s, n_recordings=[10]):
+def load_example_data(preload, window_len_s, n_recordings=10):
     """Create windowed dataset from subjects of the TUH Abnormal dataset.
 
     Parameters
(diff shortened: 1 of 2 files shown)


robintibor:
So is this correct @gemeinl ?

gemeinl:
No, actually not correct. 
The function `load_example_data` has an argument `subject_ids` which does not have any effect anymore, due to the changes introduced here. Therefore, the entire dataset will be loaded instead of just 10 recs.
We still need this argument, however, its meaning has changed. It is not subjects we are refering to but recordings. I suggest to rename the argument to `n_recordings` and give this as input to `TUHAbnormal` which has the same argument (see https://github.com/braindecode/braindecode/blob/884a78b2fb306047d444352ac4dcd9b503324fca/braindecode/datasets/tuh.py#L207-L208).

bruAristimunha:
Many thanks for the explication @gemeinl. I was wondering, can you review it again to make sure everything is correct?

examples/benchmark_lazy_eager_loading.py
```diff
@@ -68,15 +68,17 @@
 # Each one of these steps will be timed, so we can report the total time taken
 # to prepare the data and train the model.
 
-def load_example_data(preload, window_len_s, n_subjects=10):
+def load_example_data(preload, window_len_s, n_recordings=[10]):
```
gemeinl, line 71: `n_recordings=10`

examples/benchmark_lazy_eager_loading.py
```diff
@@ -89,7 +91,8 @@ def load_example_data(preload, window_len_s, n_subjects=10):
         resampled to a common sampling rate.
     """
```
gemeinl, line 92: `recording_ids=list(range(n_recordings))`

examples/benchmark_lazy_eager_loading.py
```diff
@@ -89,7 +91,8 @@ def load_example_data(preload, window_len_s, n_subjects=10):
...
     ds = TUHAbnormal(
-        TUH_PATH, target_name='pathological',
+        TUH_PATH, recording_ids=n_recordings,
```
gemeinl, line 94: `recording_ids=recording_ids`

gemeinl:
I suggest to stick as close to the original code as possible.
So change the input argument `n_subject` to `n_recordings` in `load_example_data` and create the `recorind_ids` for `TUHAbnormal` inside as before.

bruAristimunha:
Done @gemeinl.

gemeinl (approved):
Great, thanks a lot @bruAristimunha !

examples/benchmark_lazy_eager_loading.py
```diff
@@ -216,7 +221,7 @@ def run_training(model, dataloader, loss, optimizer, n_epochs=1, cuda=False):
 # Next, we define the different hyperparameters that we want to compare:
 
 PRELOAD = [True, False]  # True -> eager loading; False -> lazy loading
-N_SUBJECTS = [10]  # Number of recordings to load from the TUH Abnormal corpus
+N_RECORDINGS = [10]  # Recording ids to load from the TUH Abnormal corpus
```
robintibor, line 224: Is this now recording ids or number of recordings?

bruAristimunha: Thank you, @robintibor. You are correct. I updated the comment.

bruAristimunha: Number of recordings

bruAristimunha:
Can you merge? @robintibor

cedricrommel:
Could you solve the conflicts @bruAristimunha before I merge?

bruAristimunha:
Done @cedricrommel.

Issue #404, opened 2022-09-05
Augmentation on GPU


martinwimpff:
The Transforms of [AugmentedDataloader](https://github.com/braindecode/braindecode/blob/5a9208666bd2ebf99539ae1dab94c58bc56985d2/braindecode/augmentation/base.py#L153) currently use `X.device` to determine which device to use.
The problem with that is, that X/the dataset is often not on the GPU when getting passed to the Dataloader.
Therefore the augmentations would be processed on the CPU.

Changing [_make_collateable](https://github.com/braindecode/braindecode/blob/5a9208666bd2ebf99539ae1dab94c58bc56985d2/braindecode/augmentation/base.py#L145) from
```python
def _make_collateable(transform):
    def _collate_fn(batch):
        collated_batch = default_collate(batch)
        X, y = collated_batch[:2]
        return (*transform(X, y), *collated_batch[2:])
    return _collate_fn
```
to
```python
def _make_collateable(transform, device="cpu"):
    def _collate_fn(batch):
        collated_batch = default_collate(batch)
        X, y = collated_batch[:2]
        X = X.to(device)
        y = y.to(device)
        return (*transform(X, y), *collated_batch[2:])
    return _collate_fn
```
+adding the device Keyword to [AugmentedDataloader](https://github.com/braindecode/braindecode/blob/5a9208666bd2ebf99539ae1dab94c58bc56985d2/braindecode/augmentation/base.py#L153) would solve this problem.
What do you think about that @bruAristimunha ?

bruAristimunha:
Hi @martinwimpff!

For me, looks nice to have this parameter control! But, I asked for the @cedricrommel review since he was responsible for this part of the code.

bruAristimunha:
Hello @martinwimpff,

I created a PR to make it easier for @cedricrommel to review, could you take a look?

martinwimpff:
@bruAristimunha looks fine for me 👍

PR #406, opened 2022-09-12
Adding a new parameter in the class AugmentedDataloader to send transformation to the device.

bruAristimunha:
As suggested by @martinwimpff, this PR tries to close issue https://github.com/braindecode/braindecode/issues/404.

### DIFF
--- a/braindecode/augmentation/base.py
+++ b/braindecode/augmentation/base.py
@@ -1,6 +1,7 @@
 # Authors: Cédric Rommel <cedric.rommel@inria.fr>
 #          Alexandre Gramfort <alexandre.gramfort@inria.fr>
-#
+#          Bruno Aristimunha <b.aristimunha@gmail.com>
+#          Martin Wimpff <martin.wimpff@iss.uni-stuttgart.de>
 # License: BSD (3-clause)
 
 from typing import List, Tuple, Any
@@ -142,10 +143,14 @@ class Compose(Transform):
         return X, y
 
 
-def _make_collateable(transform):
+def _make_collateable(transform, device="cpu"):
+    """Wraps a transform to make it collateable.
+        with device control."""
     def _collate_fn(batch):
         collated_batch = default_collate(batch)
         X, y = collated_batch[:2]
+        X = X.to(device)
+        y = y.to(device)
         return (*transform(X, y), *collated_batch[2:])
     return _collate_fn
 
@@ -163,7 +168,7 @@ class AugmentedDataLoader(DataLoader):
         keyword arguments to pass to standard DataLoader class.
     """
 
-    def __init__(self, dataset, transforms=None, **kwargs):
+    def __init__(self, dataset, transforms=None, device="cpu", **kwargs):
         if "collate_fn" in kwargs:
             raise ValueError(
                 "collate_fn cannot be used in this context because it is used "
@@ -172,11 +177,11 @@ class AugmentedDataLoader(DataLoader):
         if transforms is None or (
             isinstance(transforms, list) and len(transforms) == 0
         ):
-            self.collated_tr = _make_collateable(IdentityTransform())
+            self.collated_tr = _make_collateable(IdentityTransform(), device=device)
         elif isinstance(transforms, (Transform, nn.Module)):
-            self.collated_tr = _make_collateable(transforms)
+            self.collated_tr = _make_collateable(transforms, device=device)
         elif isinstance(transforms, list):
-            self.collated_tr = _make_collateable(Compose(transforms))
+            self.collated_tr = _make_collateable(Compose(transforms), device=device)
         else:
             raise TypeError("transforms can be either a Transform object" +
                             " or a list of Transform objects.")
--- a/braindecode/augmentation/base.py
+++ b/braindecode/augmentation/base.py
@@ -143,15 +143,20 @@ class Compose(Transform):
         return X, y
 
 
-def _make_collateable(transform, device="cpu"):
-    """Wraps a transform to make it collateable.
-        with device control."""
+def _make_collateable(transform, device=None):
+    """ Wraps a transform to make it collateable.
+        with device control. """
+
     def _collate_fn(batch):
         collated_batch = default_collate(batch)
         X, y = collated_batch[:2]
-        X = X.to(device)
-        y = y.to(device)
+
+        if device is not None:
+            X = X.to(device)
+            y = y.to(device)
+
         return (*transform(X, y), *collated_batch[2:])
+
     return _collate_fn
 
 
@@ -164,18 +169,20 @@ class AugmentedDataLoader(DataLoader):
         The dataset containing the signals.
     transforms : list | Transform, optional
         Transform or sequence of Transform to be applied to each batch.
+    device : str | torch.device | None, optional
+        Device on which to transform the data. Defaults to None.
     **kwargs : dict, optional
         keyword arguments to pass to standard DataLoader class.
     """
 
-    def __init__(self, dataset, transforms=None, device="cpu", **kwargs):
+    def __init__(self, dataset, transforms=None, device=None, **kwargs):
         if "collate_fn" in kwargs:
             raise ValueError(
                 "collate_fn cannot be used in this context because it is used "
                 "to pass transform"
             )
         if transforms is None or (
-            isinstance(transforms, list) and len(transforms) == 0
+                isinstance(transforms, list) and len(transforms) == 0
         ):
             self.collated_tr = _make_collateable(IdentityTransform(), device=device)
         elif isinstance(transforms, (Transform, nn.Module)):
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -34,7 +34,7 @@ Bugs
 
 API changes
 ~~~~~~~~~~~
-
+- Adding new parameter into :class:`AugmentedDataloader` to allow the user to choose which device to use during the transformation (:gh:`406` by `Martin Wimpff`_ and `Bruno Aristimunha`_)
 - Renaming the method `get_params` to `get_augmentation_params` in augmentation classes. This makes the Transform module compatible with scikit-learn cloning mechanism (:gh:`388` by `Bruno Aristimunha`_ and `Alex Gramfort`_)
 
 .. _changes_0_6_0:
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -29,7 +29,7 @@ Enhancements
 - Adding tutorial with GridSearchCV to exemplify how to tune hyperparameters, for instance with the learning rate (:gh:`349` by `Lukas Gemein`_ and by `Bruno Aristimunha`_)
 - Adding tutorial with a Unified Validation sheme (:gh:`378` by `Bruno Aristimunha`_ and `Martin Wimpff`_)
 - Adding `verbose` parameter to :func:`braindecode.preprocessing.create_windows_from_events`, :func:`braindecode.preprocessing.create_windows_from_target_channels`, and :func:`braindecode.preprocessing.create_fixed_length_windows` (:gh:`391` by `Lukas Gemein`_)
-- Adding new parameter into :class:`AugmentedDataloader` to allow the user to choose which device to use during the transformation (:gh:`406` by `Martin Wimpff`_ and `Bruno Aristimunha`_)
+- Enable augmentation on GPU within :class:`AugmentedDataloader` via a new `device` parameter (:gh:`406` by `Martin Wimpff`_, `Bruno Aristimunha`_ and `Cedric Rommel`_)
 
 Bugs
 ~~~~
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -28,13 +28,14 @@ Enhancements
 - Adding tutorial with GridSearchCV for data augmentation on the BCIC IV 2a with module `braindecode.augmentation` (:gh:`389` by `Bruno Aristimunha`_ and `Cedric Rommel`_)
 - Adding tutorial with GridSearchCV to exemplify how to tune hyperparameters, for instance with the learning rate (:gh:`349` by `Lukas Gemein`_ and by `Bruno Aristimunha`_)
 - Adding tutorial with a Unified Validation sheme (:gh:`378` by `Bruno Aristimunha`_ and `Martin Wimpff`_)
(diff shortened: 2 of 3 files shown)


bruAristimunha:
I agree with you @robintibor! I incorporated this suggestion into the code and improved the documentation a bit. See if it's good for the merge.

agramfort:
@bruAristimunha from your experience how much speed up do you observe with this? can you give us some details on practical impact? 🙏

martinwimpff:
> @bruAristimunha from your experience how much speed up do you observe with this? can you give us some details on practical impact? 🙏

One additional thing that I observed for the GaussianNoise augmentation is that there is no speed up on the GPU if we keep using the numpy rng (from sklearns `check_random_state`). This rng always generates the random numbers on the cpu and then transfers them to the gpu.

```python
rng = check_random_state(random_state)
if isinstance(std, torch.Tensor):
    std = std.to(X.device)
noise = torch.from_numpy(
      rng.normal(
          loc=np.zeros(X.shape),
          scale=1
      ),
  ).float().to(X.device) * std
transformed_X = X + noise
return transformed_X, y
```

Using something like
```python
noise = torch.normal(mean=torch.zeros_like(X, device=X.device), std=torch.as_tensor(std, device=X.device), generator=rng)
```
instead would result in a substantial speed up. The rng would have to be a [torch.Generator](https://pytorch.org/docs/stable/generated/torch.Generator.html).
So it might be good to also change the rng class to benefit from this change in augmentations with large random tensors.

cedricrommel:
> > @bruAristimunha from your experience how much speed up do you observe with this? can you give us some details on practical impact? 🙏
> 
> One additional thing that I observed for the GaussianNoise augmentation is that there is no speed up on the GPU if we keep using the numpy rng (from sklearns `check_random_state`). This rng always generates the random numbers on the cpu and then transfers them to the gpu.
> 
> ```python
> rng = check_random_state(random_state)
> if isinstance(std, torch.Tensor):
>     std = std.to(X.device)
> noise = torch.from_numpy(
>       rng.normal(
>           loc=np.zeros(X.shape),
>           scale=1
>       ),
>   ).float().to(X.device) * std
> transformed_X = X + noise
> return transformed_X, y
> ```
> 
> Using something like
> 
> ```python
> noise = torch.normal(mean=torch.zeros_like(X, device=X.device), std=torch.as_tensor(std, device=X.device), generator=rng)
> ```
> 
> instead would result in a substantial speed up. The rng would have to be a [torch.Generator](https://pytorch.org/docs/stable/generated/torch.Generator.html). So it might be good to also change the rng class to benefit from this change in augmentations with large random tensors.

Good point @martinwimpff!

But we are potentially touching a point involving way more modifications to the code, because this remark involves all augmentations :)
Indeed, the `torch.from_numpy` is there because, when these augmentations were created, torch did not have the `torch.Generator` class yet and we wanted to be able to control the RNG object passed to the augmentation. But of course, pytorch has evolved and we should take advantage of it.

However, it is still probably worth seeing whether there are indeed any speed ups before doing the change.

cedricrommel:
> > > @bruAristimunha from your experience how much speed up do you observe with this? can you give us some details on practical impact? 🙏
> > 
> > 
> > One additional thing that I observed for the GaussianNoise augmentation is that there is no speed up on the GPU if we keep using the numpy rng (from sklearns `check_random_state`). This rng always generates the random numbers on the cpu and then transfers them to the gpu.
> > ```python
> > rng = check_random_state(random_state)
> > if isinstance(std, torch.Tensor):
> >     std = std.to(X.device)
> > noise = torch.from_numpy(
> >       rng.normal(
> >           loc=np.zeros(X.shape),
> >           scale=1
> >       ),
> >   ).float().to(X.device) * std
> > transformed_X = X + noise
> > return transformed_X, y
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Using something like
> > ```python
> > noise = torch.normal(mean=torch.zeros_like(X, device=X.device), std=torch.as_tensor(std, device=X.device), generator=rng)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead would result in a substantial speed up. The rng would have to be a [torch.Generator](https://pytorch.org/docs/stable/generated/torch.Generator.html). So it might be good to also change the rng class to benefit from this change in augmentations with large random tensors.
> 
> Good point @martinwimpff!
> 
> But we are potentially touching a point involving way more modifications to the code, because this remark involves all augmentations :) Indeed, the `torch.from_numpy` is there because, when these augmentations were created, torch did not have the `torch.Generator` class yet and we wanted to be able to control the RNG object passed to the augmentation. But of course, pytorch has evolved and we should take advantage of it.
> 
> However, it is still probably worth seeing whether there are indeed any speed ups before doing the change.

bruAristimunha:
Hello @agramfort, @martinwimpff and @cedricrommel,

I conducted a little analysis for each augmentation method listed below, performing a runtime analysis with five repetitions, passing either the GPU or the CPU as the device to the AugmentedDataLoader class.

- BandstopFilter, 
- ChannelsDropout, 
- ChannelsShuffle, 
- ChannelsSymmetry, 
- FrequencyShift, 
- FTSurrogate, 
- GaussianNoise, 
- SensorsXRotation, 
- SignFlip, 
- SmoothTimeMask, 
- TimeReverse

The code to do the analysis is in the end. I studied using the entire BCI competition 4a dataset to stress memory consumption and see how the model behaves in a scenario closer to the real one. I ran my research computer with i7-9700K CPU @ 3.60GHz and GPU: GeForce RTX 2080 Ti 11Gb. The expected is increase the the time when the augmentation is sent to the GPU.

The improvement ratio can be seen below:
image

It seems that for some raises, there is a good gain in general, except for Sensor*Rotation. I think this is related to Martin's discussion with Cedric. Some methods need to be GPU optimized, possibly in another PR.

Code:

```python
import time
from functools import partial
from itertools import product

import matplotlib.pyplot as plt
import numpy as np
import pandas as pd
import seaborn as sns
import mne
import torch

from tqdm import tqdm
from memory_profiler import memory_usage
from skorch.callbacks import LRScheduler
from skorch.helper import predefined_split

from braindecode.augmentation import AugmentedDataLoader, BandstopFilter, ChannelsDropout, ChannelsShuffle, \
    ChannelsSymmetry, FrequencyShift, FTSurrogate, GaussianNoise, SensorsXRotation, SignFlip, SmoothTimeMask, \
    TimeReverse
from braindecode.preprocessing import create_windows_from_events, exponential_moving_standardize, preprocess, \
    Preprocessor, scale
from braindecode.util import set_random_seeds

mne.set_config("MNE_LOGGING_LEVEL", "ERROR")

subject_id = None
dataset = MOABBDataset(dataset_name="BNCI2014001", subject_ids=subject_id)

low_cut_hz = 4.  # low cut frequency for filtering
high_cut_hz = 38.  # high cut frequency for filtering
# Parameters for exponential moving standardization
factor_new = 1e-3
init_block_size = 1000

preprocessors = [
    Preprocessor('pick_types', eeg=True, meg=False, stim=False),  # Keep EEG sensors
    Preprocessor(scale, factor=1e6, apply_on_array=True),  # Convert from V to uV
    Preprocessor('filter', l_freq=low_cut_hz, h_freq=high_cut_hz, verbose=False),  # Bandpass filter
    Preprocessor(exponential_moving_standardize,  # Exponential moving standardization
                 factor_new=factor_new, init_block_size=init_block_size)
]

dataset = preprocess(dataset, preprocessors)

trial_start_offset_seconds = -0.5
# Extract sampling frequency, check that they are same in all datasets
sfreq = dataset.datasets[0].raw.info['sfreq']
assert all([ds.raw.info['sfreq'] == sfreq for ds in dataset.datasets])
# Calculate the trial start offset in samples.
trial_start_offset_samples = int(trial_start_offset_seconds * freq)

# Create windows using braindecode function for this. It needs parameters to
# define how trials should be used.
windows_dataset = create_windows_from_events(
    dataset,
    trial_start_offset_samples=trial_start_offset_samples,
    trial_stop_offset_samples=0,
    preload=True,
    verbose=False)

splitted = windows_dataset.split('session')
train_set = splitted['session_T']
valid_set = splitted['session_E']

cuda = torch.cuda.is_available()  # check if GPU is available, if True chooses to use it
device = 'Cuda' if Cuda else 'CPU'
if cuda:
    torch.backends.cudnn.benchmark = True

# Set random seed to be able to roughly reproduce results
# Note that with cudnn benchmark set to True, GPU indeterminism
# may still make results substantially different between runs.
# To obtain more consistent results at the cost of increased computation time,
# you can set `cudnn_benchmark=False` in `set_random_seeds`
# or remove `torch.backends.cudnn.benchmark = True`
seed = 20200220
set_random_seeds(seed=seed, cuda=cuda)

n_classes = 4

# Extract number of chans and time steps from dataset
n_channels = train_set[0][0].shape[0]
input_window_samples = train_set[0][0].shape[1]

model = ShallowFBCSPNet(
    n_channels,
    n_classes,
    input_window_samples=input_window_samples,
    final_conv_length='auto',
)

transforms = []
probability = 1
random_state = 0

MONTAGE_10_20 = ['Fz',
                 'FC3',
                 'FC1',
                 'FCz',
                 'FC2',
                 'FC4',
                 'C5',
                 'C3',
                 'C1',
                 'Cz',
                 'C2',
                 'C4',
                 'C6',
                 'CP3',
                 'CP1',
                 'CPz',
                 'CP2',
                 'CP4',
                 'P1',
                 'Pz',
                 'P2',
                 'POz']

transform = TimeReverse(probability=probability, random_state=random_state)
transforms.append(transform)

transform = SignFlip(probability=probability, random_state=random_state)
transforms.append(transform)

transform = FTSurrogate(probability=probability,
                        phase_noise_magnitude=1, random_state=0)
transforms.append(transform)

transform = ChannelsDropout(probability=probability,
                            p_drop=probability, random_state=random_state)
transforms.append(transform)

transform = ChannelsShuffle(probability=probability,
                            p_shuffle=probability, random_state=random_state)
transforms.append(transform)

transform = GaussianNoise(probability=probability, std=0.1, random_state=random_state)
transforms.append(transform)

transform = ChannelsSymmetry(probability=probability,
                             ordered_ch_names=MONTAGE_10_20, random_state=random_state)
transforms.append(transform)

transform = SmoothTimeMask(probability=probability, random_state=random_state)
transforms.append(transform)

transform = BandstopFilter(probability=probability, sfreq=sfreq, random_state=random_state,
                           bandwidth=0.2)
transforms.append(transform)

transform = FrequencyShift(probability=probability, sfreq=sfreq,
                           random_state=random_state, max_delta_freq=2.)
transforms.append(transform)

# transform = SensorsZRotation(probability=probability,
#                             ordered_ch_names=MONTAGE_10_20, random_state=random_state)
# transforms.append(transform)

# transform = SensorsYRotation(probability=probability,
#                             ordered_ch_names=MONTAGE_10_20, random_state=random_state)
# transforms.append(transform)

transform = SensorsXRotation(probability=probability,
                             ordered_ch_names=MONTAGE_10_20, random_state=random_state)
transforms.append(transform)


def train_fit(AugmentedDataLoader, transformation, train_set, valid_set):
    lr = 0.0625 * 0.01
    weight_decay = 0

    batch_size = 64
    n_epochs = 3

    clf = EEGClassifier(
        model,
        iterator_train=AugmentedDataLoader,  # This tells EEGClassifier to use a custom DataLoader
        iterator_train__transforms=transformation,  # This sets the augmentations to use
        criterion=torch.nn.NLLLoss,
        optimizer=torch.optim.AdamW,
        train_split=predefined_split(valid_set),  # using valid_set for validation
        optimizer__lr=lr,
        optimizer__weight_decay=weight_decay,
        batch_size=batch_size,
        callbacks=[
            "accuracy",
            ("lr_scheduler", LRScheduler('CosineAnnealingLR', T_max=n_epochs - 1)),
        ],
        device=device,
    )
    # Model training for a specified number of epochs. `y` is None as it is already
    # supplied in the dataset.
    clf.verbose = False
    clf.fit(train_set, y=None, epochs=n_epochs)
    return clf


AugmentedDataLoader_gpu = partial(AugmentedDataLoader, device='cuda')
AugmentedDataLoader_cpu = partial(AugmentedDataLoader, device='cpu')

n_repets = 5  # Number of repetitions
augmentation_type = [AugmentedDataLoader_gpu, AugmentedDataLoader_cpu]
results = list()
products_list = list(product(range(n_repets), augmentation_type, transforms))

for n_rep, cpu_gpu, trans in tqdm(products_list, total=len(products_list)):
    start = time.time()
    mem = max(memory_usage(proc=(train_fit, [cpu_gpu, trans, train_set, valid_set], {})))
    time_taken = time.time() - start

    results.append({
        'idx_rep': n_rep,
        'max_mem': mem,
        'time': time_taken,
        'transf': str(trans),
        'cpu_or_gpu': str(cpu_gpu)})

data_type = pd.DataFrame(results)
data_type.to_csv("raw_data.csv")

data_type['cpu_or_gpu'] = [type_cpu_gpu[-6:-2].replace("'", "")
                           for type_cpu_gpu in data_type['cpu_or_gpu']]

analyse = data_type.groupby(["transf", "cpu_or_gpu"])['time'].apply(np.mean).unstack()
analyse = analyse.reset_index()
analyse['gpu_improving'] = 100 - ((analyse['cuda'] / analyse['cpu']) * 100)

sns.set_theme(style="white", context="talk")
f, ax = plt.subplots(1, 1, figsize=(15, 10))

sns.barplot(x=analyse.transf, y=analyse.gpu_improving, palette="icefire", ax=ax)
ax.axhline(0, color="k", clip_on=False)
plt.ylabel("Ratio gpu/cpu time.")
plt.xticks(rotation=90)
plt.tight_layout(h_pad=0)
plt.savefig("gpu_improving.png", dpi=300)
```

bruAristimunha:
It was necessary to test MixUp, as it has a slightly different logic, I could not run with this code.

agramfort:
thx @bruAristimunha 

@cedricrommel I let you merge if you're happy

cedricrommel:
I'm happy :) Thanks @bruAristimunha 
Indeed, not only does SensorsRotation not benefit from GPU, it is also the worst transform currently in terms of computation. One day I'll take some time and optimize it haha

LGTM, merging!

Let's leave the rng stuff for another PR as proposed. Not obvious either whether the proposed change is worth in terms of computation.

Issue #408, opened 2022-09-21
Augmentation Enhancements


martinwimpff:
This issue is a continuation of #404 and #406.
As mentioned in the Issue/PR above, there is room for improvement regarding the augmentation pipeline.

Before going in one direction or another, we should clarify one important question:
Do we want to continue applying augmentations batch-wise or do we want to change to sample-wise augmentations?

In the vision domain [the transforms](https://pytorch.org/vision/stable/transforms.html) are defined per sample and get loaded by multiple CPU cores in parallel. This is done by using standard dataloaders with multiple workers and implementing the transform part in the `__getitem__ `function of the dataset. 
If we want to keep doing batch-wise transforms, augmenting on the GPU would probably make sense. For batch-wise transforms there are two options: apply the augmentation with batch-wise similar augmentation parameters or apply the augmentation with sample-wise different augmentation parameters. E.g. `FTSurrogate `does the first one, and `GaussianNoise `does the second one.

The core question in my opinion would be two-fold: 
Model performance: Do sample-wise augmentations improve the model performance? By how much?
Computational Efficiency: How do the three methods compare to each other (sample-wise (CPU), batch-wise similar (GPU), batch-wise different (GPU)

I guess that the batch-wise different (GPU) method would in general be the best trade-off between performance and speed (faster than CPU, better than batch-wise GPU) but that's just a wild guess. The problem I see with this method is that some augmentations could be very hard to implement without a for loop which would affect the speed. One other thing to keep in mind is that the implementation of the sample-wise (CPU) would be very easy and seems to be the way to go looking at the torchvision library. So it might also be sufficiently fast.

@bruAristimunha @cedricrommel what are your thoughts on this one?

bruAristimunha:
Hello @martinwimpff,

Many thanks for the issue. I'm thinking about the points... I believe that we want to keep the augmentation in batch; this is useful for a self-supervised approach. However, we can create the augmentation outside of the batch, a sample-wise augmentation using something close to transformation torchvision, with some simple interface. I have the intuition that probably the best solution is combining both approaches.

Example: http://proceedings.mlr.press/v136/mohsenvand20a/mohsenvand20a.pdf

I will probably need some time to think more, but I like the suggestion of the three benchmarks.

Also, what do you think, @agramfort and @sliwy?

cedricrommel:
Hello @martinwimpff, thanks for the interest in the data augmentation submodule :)

[...]

PR #409, opened 2022-09-22
bugfix ft_surrogate

martinwimpff:
Bugfix as discussed in #408.

### DIFF
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -171,7 +171,10 @@ def _ft_surrogate(x=None, f=None, eps=1, random_state=None):
         device=device,
         random_state=random_state
     )
-    random_phase = torch.permute(torch.tile(random_phase, (f.shape[-2], 1, 1)), (1, 0, 2))
+    random_phase = torch.permute(
+        torch.tile(random_phase, (f.shape[-2], 1, 1)),
+        (1, 0, 2)
+    )
     if isinstance(eps, torch.Tensor):
         eps = eps.to(device)
     f_shifted = f * torch.exp(eps * random_phase)
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -74,27 +74,27 @@ def sign_flip(X, y):
     return -X, y
 
 
-def _new_random_fft_phase_odd(batch_size, n, device, random_state):
+def _new_random_fft_phase_odd(batch_size, c, n, device, random_state):
     rng = check_random_state(random_state)
     random_phase = torch.from_numpy(
-        2j * np.pi * rng.random((batch_size, (n - 1) // 2))
+        2j * np.pi * rng.random((batch_size, c, (n - 1) // 2))
     ).to(device)
     return torch.cat([
-        torch.zeros((batch_size, 1), device=device),
+        torch.zeros((batch_size, c, 1), device=device),
         random_phase,
         -torch.flip(random_phase, [-1])
     ], dim=-1)
 
 
-def _new_random_fft_phase_even(batch_size, n, device, random_state):
+def _new_random_fft_phase_even(batch_size, c, n, device, random_state):
     rng = check_random_state(random_state)
     random_phase = torch.from_numpy(
-        2j * np.pi * rng.random((batch_size, n // 2 - 1))
+        2j * np.pi * rng.random((batch_size, c, n // 2 - 1))
     ).to(device)
     return torch.cat([
-        torch.zeros((batch_size, 1), device=device),
+        torch.zeros((batch_size, c, 1), device=device),
         random_phase,
-        torch.zeros((batch_size, 1), device=device),
+        torch.zeros((batch_size, c, 1), device=device),
         -torch.flip(random_phase, [-1])
     ], dim=-1)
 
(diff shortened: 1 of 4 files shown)


braindecode/augmentation/functional.py
```diff
@@ -166,10 +166,12 @@ def _ft_surrogate(x=None, f=None, eps=1, random_state=None):
         device = f.device
     n = f.shape[-1]
     random_phase = _new_random_fft_phase[n % 2](
+        f.shape[0],
         n,
         device=device,
         random_state=random_state
     )
+    random_phase = torch.permute(torch.tile(random_phase, (f.shape[-2], 1, 1)), (1, 0, 2))
```
cedricrommel, line 174: I just wonder whether we want the same phase perturbation for all channels or not... Because this augmentation is based on the premise that Fourier phases don't matter, so in principle there should be no need to preserve phase dependence among channels. However, it might be good to only act on frequency information and leave spatial correlations intact. We can keep it like this for now. I'll run a few tests and we can create another PR it it seems better the other way around.

cedricrommel (approved):

cedricrommel:
I just changed the linting in a long line, but everything LGTM :) Thanks a lot @martinwimpff !
I'll carry some tests regarding the comment above and let you know.

martinwimpff:
@cedricrommel I had the same thought but I think this version makes more "physical" sense: if you interprete the phase perturbation as a simple time shift in the time domain, the perturbation should be the same across all channels. I would guess that this would not really affect the performance but I am looking forward to your test results ;)

cedricrommel:
So, results are that if we sample phase perturbations independently for each channel we break cross-channels correlations, which are instrumental in tasks such as BCI. So we see on BCI IV 2a for example that augmenting as you were doing @martinwimpff (`FTSurrogate`) is way better than doing it with fully random phases (`FTSurrogateIND`):
image


However, for tasks where cross-channel correlations don't matter that much like sleep staging, we obtain slightly better improvements if we randomize as much as possible (i.e. independently for each channel): 
image

So I've added an argument to control this and set the default to do it your way @martinwimpff since it is safer.

Thanks a lot for the bug fix and very insightful discussion !! :)

If the modifications I've made look good to you @martinwimpff  and @bruAristimunha , I'll merge the PR as soon as it passes the tests!

bruAristimunha:
looks wonderful to me :)

bruAristimunha:
looks wonderful to me :)

braindecode/augmentation/functional.py
```diff
@@ -144,6 +150,10 @@ def _ft_surrogate(x=None, f=None, eps=1, random_state=None):
     eps: float, optional
         Float between 0 and 1 setting the range over which the phase
         pertubation is uniformly sampled: [0, `eps` * 2 * `pi`]. Defaults to 1.
+    channel_indep : bool, optional
+        Whether to sample phase perturbations independently for each channel or
+        not. It is advised to set it to False when spatial information is
+        important for the task, like in BCI. Default False.
     random_state: int | numpy.random.Generator, optional
         By default None.
```
robintibor, line 158: One question, why do we even need comment for private function, especially if public function has same comment again? Can't we just remove it here to avoid duplicate docs? Also do we need default argument for channel_indep here? like a bit confusing for me there is a default here but not at the public one

bruAristimunha: I don't see a problem in the comment in a private function. This is useful for the developer who proposes a new augmentation or even to explain the library logic to a student. 

In that case, I prefer to err on the side of excess. Also, some augmentations take inspiration from other libraries.

But I don't know if there is any rule about it.

About the default value, I'm not sure.

braindecode/augmentation/functional.py
```diff
@@ -167,22 +177,27 @@ def _ft_surrogate(x=None, f=None, eps=1, random_state=None):
...
 
-def ft_surrogate(X, y, phase_noise_magnitude, random_state=None):
+def ft_surrogate(
+    X,
+    y,
+    phase_noise_magnitude,
+    channel_indep,
+    random_state=None
+):
     """FT surrogate augmentation of a single EEG channel, as proposed in [1]_.
```
martinwimpff, line 201: This line should be adapted as the new implementation is for multiple channels

braindecode/augmentation/functional.py
martinwimpff: I am not sure about the additional comment, it is more a matter of taste to me.

About the default value: I would say that either both functions should have a default value or neither of them. As a user typically only uses the `Transforms` anyway, I don't think that this decision matters too much.

To me, both "problems" raise the same question: why do we even need a private function?

robintibor: I think that in any case duplication should be avoided where possible, especially comment duplication as duplicate comments require extra work to keep in sync and can easily go out of sync as well

bruAristimunha: What do you think @cedricrommel?

cedricrommel: TBH I agree with @robintibor and @martinwimpff although I implemented the private function :) At some point it made sense because I allowed to isolate the official code from the authors, but it makes no sense anymore, so we should not have the private version anymore. I'll fix this ASAP.

braindecode/augmentation/functional.py
cedricrommel: This was intentional, the rationale being to set defaults at Transform level and force the user to take a decision for functionals (which are "lower-level"). The same designs is adopted for other transforms and functionals.

robintibor (approved):

robintibor:
seems good so is this ready to merge?

agramfort (approved):

Issue #16, opened 2020-01-14
Ensure small training batches are dropped


robintibor:
Our training loop should behave like setting drop_last to True here in dataloader https://pytorch.org/docs/stable/data.html

robintibor:
Test for Lukas @gemeinl

robintibor:
We still didn't do this I think we should set it as default in https://github.com/braindecode/braindecode/blob/49b770ef5226c64c28a1bdea6c6dcfde2567a736/braindecode/classifier.py#L45-L46 and https://github.com/braindecode/braindecode/blob/master/braindecode/regressor.py#L44-L45

robintibor:
So set similar to `iterator_train__shuffle` 
You may have to adjust acceptance tests, update the numbers there, make it easy  for yourself by using the existing testcode and printing new values 
```python
print(history_without_dur) 
```
or 
```python
import pprint
pprint(history_without_dur)
```

should be the format of the code in the test I hope.

PR #411, opened 2022-09-23
[WIP] Drop last in train

robintibor:
Closes https://github.com/braindecode/braindecode/issues/16 
This PR would by default drop the last batch of the training set during training if it is smaller than the given batch size (happens if #examples in training set is not exactly divisible by the batch size). This would affect existing training code, i.e., might change results (hence tests had to be updated), but I think it is such a sensible default it should be on by default... 
If you give a batch size that is larger than the entire training set, the behavior might be somewhat unexpected (training loop is simply skipped in skorch, but it does not crash, just no training happens)... But might be such a rare case we can live with it? otherwise we would have to catch that on call of fit I guess..
What do you think @bruAristimunha @agramfort @gemeinl @sliwy ?


agramfort:
what experiment on your end did motivate you change this default? did you observe issues without?

robintibor:
So actually when writing original HBM paper (https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.23730), this issue caused the longest "bughunt" I ever had (~about 2 weeks). One subject had a dataset size that led to the final training batch having size 1, which would cause the loss to randomly diverge after some random number of epochs (let's say 200 epochs everything fine, then suddenly diverge). 
As even today, even very state-of-the-art deep learning papers encounter mysterious training instabilities, I think anything that is likely to increase training stability without costing anything should be done. 
E.g., a current paper with mysterious instabilities related to precise batches being used: https://arxiv.org/abs/2204.02311
image

So should have maybe done this much earlier, but it is what it is.

agramfort:
Ok I am convinced :)

robintibor:
ok then just still to consider whether to catch the case with batch_size < first batch size, so case where this would mean no training. In general I think in this case anyways one should be made aware that the requested batch size is impossible as dataset too small just question how to solve this nicely

braindecode/classifier.py
```diff
@@ -59,13 +59,15 @@ class name. Name conflicts are resolved by appending a count suffix
     __doc__ = update_estimator_docstring(NeuralNetClassifier, doc)
 
     def __init__(self, *args, cropped=False, callbacks=None,
```
agramfort, line 61:
```suggestion
    def __init__(self, *args, *, cropped=False, callbacks=None,
```

I think we should enforce this but maybe in another PR

cedricrommel:
> what experiment on your end did motivate you change this default? did you observe issues without?

We were getting this in our experiments if you remember well @agramfort :)
We were doing the same in our exp code and I think it makes total sense to make it the default.

robintibor:
so with regard to giving the user some error/warning in case batch size < min batch size, should we do this? I think it is not super trivial, probably on call to fit(), one would need to check the lengths of the train loader iterator and then give some warning (or error?) if it is 0... should we still do this now or defer to 0.8 even? Skorch doesn't raise an error in that case and runs without training and without computing training metrics...

bruAristimunha:
> so with regard to giving the user some error/warning in case batch size < min batch size, should we do this? I think it is not super trivial, probably on call to fit(), one would need to check the lengths of the train loader iterator and then give some warning (or error?) if it is 0... should we still do this now or defer to 0.8 even? Skorch doesn't raise an error in that case and runs without training and without computing training metrics...

It looks like this is more unexpected behaviour for the skorch team than it does for us. For me, we don't do anything in version 0.7 and try to solve it together (braindecode and skorch) for version 0.8 with a warning.

braindecode/classifier.py
bruAristimunha: Hi @agramfort,

I find this parameter change very interesting, but I agree with you that this has to be addressed in another pull request. I think that with the proposed code change at the moment, we maybe already have a change in several places in the library, or not.

agramfort: 👌

bruAristimunha:
Can we merge @robintibor?

robintibor:
yes feel free to merge then  @bruAristimunha

Issue #412, opened 2022-09-27
About deprecated functions expected in version 0.7


bruAristimunha:
We have some functions that need to be deprecated in version 0.7.

> [MNEPreproc](https://github.com/braindecode/braindecode/blob/8b3435b24d36cec20a8137287f7b0b9ec739f16f/braindecode/preprocessing/preprocess.py#L89), [NumpyPreproc](https://github.com/braindecode/braindecode/blob/8b3435b24d36cec20a8137287f7b0b9ec739f16f/braindecode/preprocessing/preprocess.py#L107), [zscore](https://github.com/braindecode/braindecode/blob/8b3435b24d36cec20a8137287f7b0b9ec739f16f/braindecode/preprocessing/preprocess.py#L376) and [scale](https://github.com/braindecode/braindecode/blob/8b3435b24d36cec20a8137287f7b0b9ec739f16f/braindecode/preprocessing/preprocess.py#L404).

@sliwy and I commented on this question in PR #397. It looks like the deprecated warning in the preprocessing.process.scale is with the wrong message.

It gives the impression that we only need to replace the scale function with multiply from numpy, but in reality, you need to establish a lambda function that will use numpy.multiply.

The message currently is:

> will be removed in 0.7.0. Use numpy.multiply instead.

And what we are expecting from the user is that it does something like this:

```python
    Preprocessor(
        lambda data, factor: np.multiply(data, factor),
        factor=1e6,
    ),
```

I don't think it's very reasonable for someone new to braindecode. Another problem also related to this function is that it is present in several tutorials, as listed below:

1. https://braindecode.org/master/auto_examples/plot_hyperparameter_tuning_with_scikit-learn.html
2. https://braindecode.org/master/auto_examples/plot_data_augmentation.html
3. https://braindecode.org/master/auto_examples/plot_sleep_staging_eldele2021.html
4. https://braindecode.org/master/auto_examples/plot_data_augmentation_search.html
5. https://braindecode.org/master/auto_examples/plot_data_augmentation_search.html
6. https://braindecode.org/master/auto_examples/plot_bcic_iv_2a_moabb_trial.html
7. https://braindecode.org/master/auto_examples/plot_bcic_iv_2a_moabb_cropped.html
8. https://braindecode.org/master/auto_examples/plot_relative_positioning.html

I propose to delay this deprecating to version 0.8 only for the scale function and replace the scale function in all the tutorials above. The other functions I think are okay to deprecate in version 0.7

[...]

PR #413, opened 2022-09-30
Updated depreciation

bruAristimunha:
As discussed in issue #412, this PR defers the deprecation of the scale function, and updates tutorials where the function is used.

### DIFF
--- a/examples/plot_bcic_iv_2a_moabb_cropped.py
+++ b/examples/plot_bcic_iv_2a_moabb_cropped.py
@@ -90,10 +90,12 @@ high_cut_hz = 38.  # high cut frequency for filtering
 # Parameters for exponential moving standardization
 factor_new = 1e-3
 init_block_size = 1000
+# Factor to convert from V to uV
+factor = 1e6
 
 preprocessors = [
     Preprocessor('pick_types', eeg=True, meg=False, stim=False),  # Keep EEG sensors
-    Preprocessor(lambda data, factor: multiply(data, factor), factor=1e6),  # Convert from V to uV
+    Preprocessor(lambda data: multiply(data, factor)),  # Convert from V to uV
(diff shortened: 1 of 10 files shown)


examples/plot_bcic_iv_2a_moabb_cropped.py
```diff
@@ -92,7 +93,7 @@
 
 preprocessors = [
     Preprocessor('pick_types', eeg=True, meg=False, stim=False),  # Keep EEG sensors
-    Preprocessor(scale, factor=1e6, apply_on_array=True),  # Convert from V to uV
+    Preprocessor(lambda data, factor: multiply(data, factor), factor=1e6),  # Convert from V to uV
```
robintibor, line 96:
```suggestion
    Preprocessor(partial(multiply, 1e6)),  # Convert from V to uV
```
should also work? Maybe it is nicer?
Also wondered if we could also forward *args to the function be applied and this way write something like
```
Preprocessor(multiply,2)
```
What do you think @bruAristimunha ?

bruAristimunha: I like the first option, using the partial, but this is definitely not a strong opinion.

bruAristimunha: whichever option you choose, I'm happy

bruAristimunha: What do you think @agramfort?

agramfort: No strong feeling. Partial is more pythonic but maybe harder for new comers. Now our custom API is not standard either. So as you prefer

bruAristimunha: Let's go for simple option, okay @robintibor?

agramfort (commented):
guys I don't have much time these days. I let you approve and marge here.

PR #414, opened 2022-10-04
[WIP] New return for sleep architecture

tgnassou:
The PR proposes a new way to choose which layers of the architecture is return. 

The old architecture had a bool parameters called "return_feats" allowing to choose to return the prediction for the classifier if it was false and to return the features (i.e. the output of the features extractor) if it was True.

I propose to change this parameters by a string called "layers_returned". This string allows to return the prediction for the classifier if it equals to "classifier", to return the features (i.e. the output of the features extractor) if it equals to "features" and to return both prediction and features if it equals to "all".

My changes are motivated by the application domain adaptation methods to sleep staging which requires both prediction and features.

@agramfort


braindecode/models/sleep_stager_blanco_2020.py
```diff
@@ -50,7 +51,7 @@ class SleepStagerBlanco2020(nn.Module):
 
     def __init__(self, n_channels, sfreq, n_conv_chans=20, input_size_s=30,
                  n_classes=5, n_groups=2, max_pool_size=2, dropout=0.5, apply_batch_norm=False,
-                 return_feats=False):
+                 layers_returned="classifier"):
```
agramfort, line 54: @tgnassou you cannot do this. This will break backward compatibility. Please avoid changes that can break existing code.

braindecode/models/sleep_stager_blanco_2020.py
```diff
@@ -36,10 +36,11 @@ class SleepStagerBlanco2020(nn.Module):
...
         If True, apply batch normalization after both temporal convolutional
         layers.
-    return_feats : bool
-        If True, return the features, i.e. the output of the feature extractor
-        (before the final linear layer). If False, pass the features through
-        the final linear layer.
+    layers_returned : str
+        If equals to "classifier", return the prediction of the final layer,
+        If equals to "features, return the features, i.e. the output of the feature extractor
+        (before the final linear layer). If equals to "all",
```
agramfort: run flake8 on your code

braindecode/models/sleep_stager_blanco_2020.py
tgnassou: Even if the code becomes heavy ? (In general case)

agramfort: it's ok to change a public API but it needs some deprecation messages. See for example https://github.com/braindecode/braindecode/pull/413

now before moving forward with this you need to see with others if the API change is ok for the core devs of the project.

besides for me the term "features" is not ideal. For me the word "embedding" is more common to refer to the vector representation before the final layer.

bruAristimunha:
Hello @tgnassou,

I saw that you closed the PR. Do you need some help with this new functionality? I think this new return could be very interesting.

tgnassou:
Hi @bruAristimunha,
Finally I don't need this PR anymore. To deal with my problem, I chose to use hook of pytorch. It allows me to be able to get as many as intermediate layer outputs that I want without change the API of braindecode.

PR #416, opened 2022-10-04
Updating the versions.json.

bruAristimunha:
We will have three versions of the library on the braindecode site. Updating the json.


agramfort (approved):

agramfort:
does it work for you at https://output.circle-artifacts.com/output/job/34744252-3e16-42b6-a502-d355800417e4/artifacts/0/dev/index.html ?

bruAristimunha:
Hi @agramfort,

Small update:

I'm not sure if it's not working because the pointed address doesn't exist, or if the div has some problem. Anyway, I will investigate this during the week to try to make the release possible by the end of this week.

Thank you for your support

bruAristimunha:
I studied how the code in javascript makes the version list, and it looks like it needs the index.html address working. 

I'll go ahead with the release and possibly do a PR to fix any new issues. As it's my first release, I'm possibly going to do it very slowly to ensure nothing is breaking.

cc. @agramfort and @robintibor.

PR #417, opened 2022-10-05
[MRG] Add deepsleepnet to model for sleep staging

tgnassou:
I propose to add the model DeepSleepNet to braindecode. The model of Supratak et. al.  is still cited a lot in the recent studies of sleep staging and I think it can be a good adding to braindecode. 

The model proposed in the paper is for single channels EEG of 30sec with a frequency of 100Hz. I reimplemented the code as proposed in the paper, whithout changing. So all the parameters are hardcoded, and it is not robust to the change of input size.

I check all the size of the layer output, normally everything should be okay. I will think about a method to validate the architecture.

paper: https://arxiv.org/pdf/1703.04046.pdf
code: https://github.com/akaraspt/deepsleepnet

@agramfort


braindecode/models/deepsleepnet.py
```diff
@@ -0,0 +1,239 @@
+import torch
+import torch.nn as nn
+
+
+class small_CNN(nn.Module):  # smaller filter sizes to learn temporal information
```
agramfort, line 5: if this class is for internal is should be prefixed with `_` and class names should be in CamelCase so:
```suggestion
class _SmallCNN(nn.Module):  # smaller filter sizes to learn temporal information
```

braindecode/models/deepsleepnet.py
```diff
@@ -0,0 +1,239 @@
...
+        x = self.conv1(x)
+        x = self.dropout(self.pool1(x))
+        x = self.conv2(x)
+        x = self.conv3(x)
+        x = self.conv4(x)
+        x = self.pool2(x)
+        return x
+
+
+class large_CNN(nn.Module):  # larger filter sizes to learn frequency information
```
agramfort, line 73: idem
```suggestion
class _LargeCNN(nn.Module):  # larger filter sizes to learn frequency information
```

braindecode/models/deepsleepnet.py
```diff
@@ -0,0 +1,239 @@
+import torch
+import torch.nn as nn
+
+
+class small_CNN(nn.Module):  # smaller filter sizes to learn temporal information
+    def __init__(
+        self,
+    ):
+
```
agramfort, line 9:
```suggestion
    def __init__(self):
```

braindecode/models/deepsleepnet.py
```diff
@@ -0,0 +1,239 @@
...
+        x = self.conv4(x)
+        x = self.pool2(x)
+        return x
+
+
+class large_CNN(nn.Module):  # larger filter sizes to learn frequency information
+    def __init__(
+        self,
+    ):
+
```
agramfort, line 77:
```suggestion
    def __init__(self):
```

braindecode/models/deepsleepnet.py
```diff
@@ -0,0 +1,239 @@
...
+        x = self.conv1(x)
+        x = self.dropout(self.pool1(x))
+        x = self.conv2(x)
+        x = self.conv3(x)
+        x = self.conv4(x)
+        x = self.pool2(x)
+        return x
+
+
+class BiLSTM(nn.Module):
```
agramfort, line 142:
```suggestion
class _BiLSTM(nn.Module):
```

test/unit_tests/models/test_models.py
```diff
@@ -344,3 +344,40 @@ def test_eegitnet_shape():
...
+
+    model = DeepSleepNet(n_classes=n_classes, return_feats=True)
+    model.eval()
+
+    rng = np.random.RandomState(42)
+    X = rng.randn(n_examples, n_channels, int(sfreq * input_size_s))
+    X = torch.from_numpy(X.astype(np.float32))
+
+    out = model(X.unsqueeze(2))
+    assert out.shape == (n_examples, model.len_last_layer)
```
agramfort, line 383: there is a lot of copy paste in these 2 tests. Please avoid this.

tgnassou: I try to do the same kind of test of the other sleep model. I changed a little bit these two let me know if I have to modify it or not.

tgnassou:
I wanted to do a test of number parameter of deepsleepnet but the number of parameters is not given in the paper. I found only a paper saying that the model contains 24,7 M of parameters. I have the same number for my model.

Then I found the code of the writer of deepsleepnet in tensorflow. the ouput size of every layer is the same of my model, nevertheless the number of parameter differs. Their model has 22,6M parameters. The difference comes from biLSTM and I did't really get why. 

Is it mandatory to check the number of parameters, knowing that this number is not given in the paper ? Do I have to find an other way to validate the model ?

Link to the paper talking about the number of param (page 8): https://arxiv.org/abs/2108.10600#:~:text=24%20Aug%202021%5D-,DeepSleepNet%2DLite%3A%20A%20Simplified%20Automatic%20Sleep%20Stage,Scoring%20Model%20with%20Uncertainty%20Estimates&text=Deep%20learning%20is%20widely%20used,learn%20feature%20from%20the%20data.

tgnassou (commented):
@agramfort Do you think I can call an Identity layer feature_extractor in order to use the hook of my project's function to get the feature output ? Maybe put a comment to make it clearer ? 
Otherwise the last laye before classifier is a dropout, I don't really know how to deal with it except if I change the name of this only dropout layer.

braindecode/models/deepsleepnet.py
```diff
@@ -197,6 +197,7 @@ def __init__(self, n_classes=5, return_feats=False):
         self.fc = nn.Sequential(nn.Linear(3072, 1024, bias=False),
                                 nn.BatchNorm1d(num_features=1024))
 
+        self.features_extractor = nn.Identity()
```
tgnassou, line 200: @agramfort Do you think I can call an Identity layer feature_extractor in order to use the hook of my project's function to get the feature output ? Maybe put a comment to make it clearer ?
Otherwise the last laye before classifier is a dropout, I don't really know how to deal with it except if I change the name of this only dropout layer.

agramfort: can you add a test using the hook to obtain features? Doing so everyone can see why you need it and if someone remove the identity someday it will break a test.

bruAristimunha:
Hi @agramfort,

I saw this PR and thought, can we integrate this new model? It seems that @tgnassou answered all requested items of your review.

agramfort (commented):
@tgnassou what do you think ?

@bruAristimunha while you are thinking about more uniform API I would say we need common tests like in sklearn. The machinery to extract a specific layer activations should / could be standardized

tgnassou:
It seems good to go for me!

@bruAristimunha I think I waited to finish this PR because we talked with @agramfort to redo the len_last_layer function, but I think it can be done on another PR, with uniform test and standard layer activation extraction.

agramfort:
@tgnassou can you just add a what's new entry ? in the changelog

thx

tgnassou:
Done

PR #418, opened 2022-10-05
[MRG] Add len_last_layer to USleep model as other sleep stager model

tgnassou:
I added the attribut len_last_layer to USleep model like the other sleep stager models. 

@agramfort


agramfort (commented):
can you add a test?

braindecode/models/usleep.py
```diff
@@ -229,34 +233,37 @@ def __init__(self,
...
+                    padding=0,
+                ),                         # output is (B, n_classes, S)
+                nn.ELU(),
+                nn.Conv1d(
+                    in_channels=n_classes,
+                    out_channels=n_classes,
+                    kernel_size=1,
+                    stride=1,
+                    padding=0,
+                ),
```
agramfort, line 263: @tgnassou can you explain somewhere why you don't take as features the activations before the last softmax?

tgnassou: @agramfort I do it because I take thefeatures before what the author calls classifier (3 conv layers). But it is also possible to consider the features the activation before the last softmax. Do you think there is a better choice ?

agramfort: ok then. Can you just add a comment in the code about this?

bruAristimunha:
@tgnassou can I close the abandoned pr?

bruAristimunha:
I think it's unlikely you still need it, and good luck with your defense. I just found out!

PR #428, opened 2022-11-07
New EEG-Inception architecture for motor imagery

cedricrommel:
I realized that the there are two papers proposing models called EEGInception in the recent literature: one for ERP classification (which is already implemented in braindecode) and another for motor imagery. This PR renames the existing EEG-Inception model for ERP to raise the ambiguity and implements the other architecture for motor-imagery tasks.

### DIFF
--- a/braindecode/models/eeginception.py
+++ b/braindecode/models/eeginception.py
@@ -20,7 +20,7 @@ class EEGInception(nn.Sequential):
     """ EEG Inception for ERP-based classification
 
     --> DEPERECATED <--
-    THIS CLASS IS DEPRECATED AND WILL BE REMOVED IN THE NEXT RELEASE OF
+    THIS CLASS IS DEPRECATED AND WILL BE REMOVED IN THE RELEASE 0.9 OF
     BRAINDECODE. PLEASE USE braindecode.models.EEGInceptionERP INSTEAD IN THE
     FUTURE.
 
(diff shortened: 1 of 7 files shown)


braindecode/models/eeginception_erp.py
```diff
@@ -14,10 +14,9 @@ def _transpose_to_b_1_c_0(x):
     return x.permute(0, 3, 1, 2)
 
 
-class EEGInception(nn.Sequential):
```
agramfort, line 17: you are here breaking backward compatibility as this model was shipped in a previous release. Can I ask you to properly deprecate it? thanks

cedricrommel: Sure!

cedricrommel:
Let me know if this is what you had in mind @agramfort

braindecode/models/eeginception.py
```diff
@@ -104,6 +115,13 @@ def __init__(
     ):
         super().__init__()
 
+        warn(
+            "The class EEGInception is deprecated and will be removed in the "
+            "next release of braindecode. Please use "
```
agramfort, line 120: can you put the version number?

braindecode/models/eeginception.py
```diff
@@ -44,28 +50,33 @@ class EEGInception(nn.Sequential):
...
+        Activation function. Defaults to ELU activation as in
+        [Santamaria2020]_.
+    batch_norm_alpha: float, optional
+        Momentum for BatchNorm2d. Defaults to 0.01.
+    depth_multiplier: int, optional
+        Depth multiplier for the depthwise convolution. Defaults to 2 as in
+        [Santamaria2020]_.
+    pooling_sizes: list(int), optional
+        Pooling sizes for the inception blocks. Defaults to 4, 2, 2 and 2, as
+        in [Santamaria2020]_.
```
agramfort, line 79: these changes need to be reverted

docs/api.rst
```diff
@@ -56,7 +56,7 @@ Models
 
     ShallowFBCSPNet
     Deep4Net
-    EEGInception
+    EEGInceptionERP
```
agramfort, line 59: can you add the EEGInceptionMI too here?

cedricrommel: Done

braindecode/models/eeginception.py
cedricrommel: You're right. Done!

braindecode/models/eeginception.py
cedricrommel: Sure! Done

agramfort (approved):

agramfort:
@bruAristimunha @sliwy feel free to merge

braindecode/models/eeginception_mi.py
```diff
@@ -0,0 +1,289 @@
...
+        X2 = self.pooling(X)
+        X2 = self.pooling_conv(X2)
+
+        out = torch.cat(X1 + [X2], 1)
+
+        out = self.bn(out)
+        return self.activation(out)
+
+
+class _ResidualBlockMI(nn.Module):
```
sliwy, line 259: I would rather change the name to `_ResidualModuleMI` to fit the description in the paper. Residual block is also often used to describe whole block with the residual connection inside so it can be misleading.

The forward method looks like a standard `nn.Sequential` forward, so we can remove at least the forward implementation by inheriting from `nn.Sequential` or we can remove this class and create `nn.Sequential` objects directly in `EEGInceptionMI.__init__`.

cedricrommel: Thanks @sliwy ! I just renamed it into `_ResidualModuleMI`, you are right. It is also the way they call it in the paper actually!

Concerning inheriting from `nn.Sequential`, to be honest I am not a big fan and usually prefer to code the `forward` method myself as it gives more flexibility for debugging. Doing it would reduce a bit the number of lines in `_ResidualModuleMI` but without much code clarity improvements IMO. And doing it for `_InceptionModuleMI` and `EEGInceptionMI` could actually degrade clarity I think, as those don't have a straightforward sequential structure. Did I convince you on this point? :)

sliwy: For `_InceptionModuleMI` and `EEGInceptionMI` - no doubts, it should stay as `nn.Module` with custom forward method.

In the case of `_ResidualModuleMI`, I would still use `nn.Sequential` but it's up to you - it depends on taste. 

@cedricrommel great contribution! To me it's ready to merge.

sliwy (approved):
@cedricrommel great stuff! To me it's ready to merge :)

cedricrommel:
Great thanks! I'll merge it then!

bruAristimunha (approved):

PR #429, opened 2022-11-14
Implementation of ATCnet

cedricrommel:


cedricrommel:
ping @tomMoral @agramfort  :)

agramfort (approved):
LGTM although I would make all classes in ATCNet.py with a starting _ to be private except the ATCNet class that is public.

🙏 @cedricrommel

braindecode/models/atcnet.py
```diff
@@ -0,0 +1,735 @@
...
+            **kwargs
+        )
+
+    def forward(self, X):
+        out = self.conv(X)
+        return out[..., :-self.conv.padding[0]]
+
+
+class MaxNormLinear(nn.Module):
+    """Linear layer with MaxNorm constraing on weights.
```
robintibor, line 685:
```suggestion
    """Linear layer with MaxNorm constraining on weights.
```

braindecode/models/atcnet.py
```diff
@@ -0,0 +1,735 @@
...
+            in_features=in_features,
+            out_features=out_features,
+            bias=bias,
+        )
+
+    def forward(self, X):
+        self._max_norm(self.linear.weight)
+        return self.linear(X)
+
+    def _max_norm(self, w):
```
robintibor, line 729: Could you add a small comment here that this will modify w in place? was a bit unsure at first whether it does that

cedricrommel: Sure! you are right, this should be explicitly said! Thanks for the remark!

cedricrommel:
Feel free to merge if it looks good to you :)

tomMoral (approved):
LGTM. 

Just one nitpick but this is on me, I don't like the warpping style when subclassing does the job. feel free to ignore.

tomMoral:
Same can be done with the maxnorm layer.

cedricrommel:
> Same can be done with the maxnorm layer.

Good point! Did the same for `MaxNormLiner`

Issue #286, opened 2021-06-25
Democracy in Action For Naming


robintibor:
So right now in quite some models we have:

https://github.com/braindecode/braindecode/blob/4eb0af49824d764d771792931511b449a092f49f/braindecode/models/deep4.py#L38-L39

Let's vote how to change it!

robintibor:
First vote:

`chans` ![autoformatting](https://github.githubassets.com/images/icons/emoji/unicode/1f389.png "Auto formatting")

vs.

`channels` ![pep8](https://github.githubassets.com/images/icons/emoji/unicode/1f680.png "Auto formatting")

robintibor:
Second vote:

`n_in_chan[nel]s, n_out_chan[nel]s` ![1](https://github.githubassets.com/images/icons/emoji/unicode/1f389.png "1")

`n_chan[nels], n_classes` ![2](https://github.githubassets.com/images/icons/emoji/unicode/1f680.png "2")

`in_channels, out_channels` ![3](https://github.githubassets.com/images/icons/emoji/unicode/2764.png "3")

robintibor:
What a marvelous way to vote! @agramfort notice second vote also up

agramfort:
I am fine to have « in » in the name if this also one with « out ». If there’s no ambiguity I don’t see the point

agramfort:
And if we have a pytorch convention and the semantic matches it should be in_channels

robintibor:
keep in mind in the end n_classes loses meaning for regression tasks obviously

agramfort:
Should we rename it to out_channels then ?

l-omar-chehab:
> And if we have a pytorch convention and the semantic matches it should be in_channels

Agreed, it would be nice to have some consistency with Pytorch naming:
-- Convolutional layers in Pytorch use: nn.Conv1d(in_channels, out_channels)
-- Linear layers in Pytorch use: nn.Linear(in_features, out_features)

We could rename in_chans to in_channels for consistency with Pytorch, in braindecode's convolutional models. 

I think it could be worthwhile to keep n_classes as is:
-- this avoids the naming ambiguity of choosing between out_features and out_channels, depending on if the prediction is outputed from a Linear or Conv layer
-- it is more explicit : we understand via n_classes that we are doing a classification task. This is less obvious if we rename n_classes to out_channels, for eg. 

What do you think?

agramfort:
[...]

PR #439, opened 2023-03-08
Standardizing parameters in sleep models

bruAristimunha:
Hello everyone,

I hope this message finds you well. I wanted to share some observations I made today while studying sleep models. For your reference, I have shared a [Google Spreadsheets](https://docs.google.com/spreadsheets/d/1WZnm5DfgVM3CeRke1SEGHVSkMB1Tt-wkt34fpkfVkDc/edit?usp=sharing) link that includes the table that I made.

I would like to propose some changes that could potentially help disambiguate the parameters of these models, such as renaming "`in_chans`" to "`n_channels`" and "`n_conv_channels`" and "`n_conv_chans`" to simply "`n_conv_channels`".

To this end, I have created a function `check_deprecation_warning` that can help disambiguate other models, particularly in the BCI models, where there is more room for improvement.
 
Regarding the yellow variable in the table, I was initially unsure about its disambiguation. However, upon further analysis of the code's logic, I decided not to change "max_pooling" in the second instance from "max pooling in vector size" (?).

It seems to me that there is a need for improved documentation on how we describe each parameter and better internal standardisation on some internal variables. However, I understand this is not the appropriate PR for this matter. 

I would be grateful if you could review my proposed changes and provide any feedback you may have. Additionally, if you identify any other variables that can benefit from disambiguation, please feel free to share your comments.

Thank you very much in advance for your attention and assistance!


braindecode/models/sleep_stager_blanco_2020.py
```diff
@@ -1,11 +1,16 @@
...
 #
 # License: BSD (3-clause)
 
 import numpy as np
 import torch
 from torch import nn
 
+from .util import check_deprecation_warning
+
+pair_new_names_blanco = {"n_conv_chans": "n_conv_channels"}
```
robintibor, line 12: I woould move this to inside the constructor, even to inside the if... no need to have this so global I think

braindecode/models/sleep_stager_blanco_2020.py
```diff
@@ -48,44 +56,55 @@ class SleepStagerBlanco2020(nn.Module):
...
-                 return_feats=False):
+                 return_feats=False, **kwargs):
         super().__init__()
 
-        input_size = np.ceil(input_size_s * sfreq).astype(int)
-
         self.n_channels = n_channels
+        self.n_conv_channels = n_conv_channels
+
+        if kwargs:
```
robintibor, line 67:
```suggestion
        if kwargs:
            pair_new_names_blanco = {"n_conv_chans": "n_conv_channels"}
```

braindecode/models/util.py
```diff
@@ -1,13 +1,55 @@
...
+        A dictionary containing the old parameter name as key and the new
+        parameter name as value.
+
+    Notes
+    -----
+    This function should be called in the __init__ method of the class that
+    contains the deprecated parameters.
+
+    """
+    if not all([k in old_new_name_pairs for k in kwargs]) and \
```
robintibor, line 34: I feel better make this in a loop and then give a specific error, like which one is not valid... or you could also use set intersection etc instead of loop, but definitely feel better if error warning says which one is not correct

test/unit_tests/models/test_util.py
```diff
@@ -11,6 +11,17 @@
 from braindecode.models.modules import Expression
 from braindecode.models.util import (
     get_output_shape, aggregate_probas, _pad_shift_array)
+from braindecode.models import USleep
+
+
+def test_check_deprecation_warning():
+    in_chans = 2
+
+    with pytest.warns(
```
robintibor, line 20: nice test

bruAristimunha: Thank you @robintibor =)

bruAristimunha:
Hi @robintibor!

I updated the code to use set and moved the variable into the models as you suggested! Thank you so much for the quick and good review :)

I think @agramfort, already make his pre-review with the comment: https://github.com/braindecode/braindecode/issues/286#issuecomment-1461537272

But, I didn't fully understand how to change from kwargs to ",*".

bruAristimunha:
Hi @agramfort

I've taken the time to study [PEP 3102](https://peps.python.org/pep-3102/), and now I have a better understanding of the use of the "," parameter to indicate the end of positional parameters.

However, I'm still a bit unsure if this is a good approach for our particular case. From what I can see, it would only be useful if we were hardcoding all the old and new parameters into the functions. Am I on the right track? 

I think what I'm proposing here is a little simpler and more readable. What do you think?

Thank you very much for your time and consideration!

bruAristimunha:
If I understand correctly, the idea is to change to:

```python 
def __init__(self, in_channels, n_channels, sfreq, n_conv_channels=20, input_size_s=30, n_classes=5, 
n_groups=2, max_pool_size=2, dropout=0.5, apply_batch_norm=False,  return_feats=False, *, in_chans=None):
```

Instead of the code I proposed in the PR with kwargs.

agramfort (commented):
looks reasonable. I am surprised you don't have to any other test. To me an uncaught warning should raise an error in a test and maybe even when building the doc. I am assume their are many other places where you need to do the changes no?

PR #442, opened 2023-03-13
Adding date.txt to TUH Classes

gemeinl:
upon first instantiation now always creating a date txt file. in following instantiations it is read to sort recordings chronologically without the necessity to perform slow operations beforehand


robintibor:
Great, thanks @gemeinl now things will be fast again :)

PR #443, opened 2023-03-21
adding NMT dataset

MohammadJavadD:
As the [NMT](https://dll.seecs.nust.edu.pk/downloads/) dataset is similar to TUH abnormal, with a few changes we can add this dataset to Braindecode! 
[Reference paper ](https://www.frontiersin.org/articles/10.3389/fnins.2021.755817/full)


robintibor:
Thanks for starting this!
Having NMT be a subclass of TUH seems a potentially confusing decision to me.
We also had some code here to load NMT: https://gist.github.com/gemeinl/e932169ed2eca5bf8b25432f0d54b095
Is there any difference in functionality between these two versions as far as you can see?

In order to avoid potential confusions and also that code may break for one dataset if the other is changed, I suggest we either keep them separate as in gist above or we define common functions that both use in a separate file. This might be overkill, definitely I think subclassing TUH  here seems too high a risk of creating unintentional errors later

MohammadJavadD:
> Thanks for starting this! Having NMT be a subclass of TUH seems a potentially confusing decision to me. We also had some code here to load NMT: https://gist.github.com/gemeinl/e932169ed2eca5bf8b25432f0d54b095 Is there any difference in functionality between these two versions as far as you can see?

No, they are the same! I haven’t seen that gist and it worked for me with a few changes! 

> In order to avoid potential confusions and also that code may break for one dataset if the other is changed, I suggest we either keep them separate as in gist above or we define common functions that both use in a separate file. This might be overkill, definitely I think subclassing TUH here seems too high a risk of creating unintentional errors later

Sure! Having a separate file would be a better option.

robintibor:
> > Thanks for starting this! Having NMT be a subclass of TUH seems a potentially confusing decision to me. We also had some code here to load NMT: https://gist.github.com/gemeinl/e932169ed2eca5bf8b25432f0d54b095 Is there any difference in functionality between these two versions as far as you can see?
> 
> No, they are the same! I haven’t seen that gist and it worked for me with a few changes!
> 
> > In order to avoid potential confusions and also that code may break for one dataset if the other is changed, I suggest we either keep them separate as in gist above or we define common functions that both use in a separate file. This might be overkill, definitely I think subclassing TUH here seems too high a risk of creating unintentional errors later
> 
> Sure! Having a separate file would be a better option.

Great would you mind having a try to create this separate file then based on the gist?

bruAristimunha:
Hi @MohammadJavadD,

do you need any help?

bruAristimunha:
Hi @MohammadJavadD,

do you need any help?

MohammadJavadD:
@bruAristimunha I made the discussed change to create a separate class for the NMT dataset, Please take a look.

braindecode/datasets/nmt.py
```diff
@@ -0,0 +1,89 @@
...
+# License: BSD (3-clause)
+
+import os
+import glob
+
+import numpy as np
+import pandas as pd
+import matplotlib.pyplot as plt
+import seaborn as sns
+sns.set(color_codes=True)
```
bruAristimunha, line 16:
```suggestion

```

braindecode/datasets/nmt.py
```diff
@@ -0,0 +1,89 @@
...
+import mne
+mne.set_log_level("ERROR")
```
bruAristimunha, line 18:
```suggestion

```

braindecode/datasets/nmt.py
```diff
@@ -0,0 +1,89 @@
...
+        Parent directory of the dataset.
+    recording_ids: list(int) | int
+        A (list of) int of recording id(s) to be read (order matters and will
+        overwrite default chronological order, e.g. if recording_ids=[1,0],
+        then the first recording returned by this class will be chronologically
+        later then the second recording. Provide recording_ids in ascending
+        order to preserve chronological order.).
+    target_name: str
+        Can be 'gender', or 'age'.
+    preload: bool
```
bruAristimunha, line 44:
```suggestion
    pre_load: bool
```

bruAristimunha:
We need a fetch for the dataset. I am doing something @MohammadJavadD

bruAristimunha:
Can you make the tests in the meantime, @MohammadJavadD?

There are plenty of tests for Temple University on braindecode, and you can follow the same type of mock test.

MohammadJavadD:
> We need a fetch for the dataset. I am doing something @MohammadJavadD

[This](https://github.com/MohammadJavadD/APD_EEG/blob/8f628f37b3485e1a635b92488961aada1b429a7a/s1_data_preparation.ipynb#L89) might help

bruAristimunha:
Thank you for the suggestion @MohammadJavadD, I saw your code, but I will use mne to fetch the data. I am uploading the dataset in a better place than google drive.

bruAristimunha:
If you need any help, please let me know.

MohammadJavadD:
@bruAristimunha I added a mock dataset for the NMT and created some unit tests based on that.

bruAristimunha:
Hey @apmellot, @jpaillard and @dengemann!

FYI.

We are adding this potential brain-age dataset, smaller than the Temple University Dataset, as the community uses the dataset, cc. @MohammadJavadD.  

You may have options about the dataset, but as you work with the TUH dataset, I think you would like to know about it. If you have time to review, that would be nice, but okay otherwise.

MohammadJavadD:
> Hey @apmellot, @jpaillard and @dengemann!
> 
> FYI.
> 
> We are adding this potential brain-age dataset, smaller than the Temple University Dataset, as the community uses the dataset, cc. @MohammadJavadD.
> 
> You may have options about the dataset, but as you work with the TUH dataset, I think you would like to know about it. If you have time to review, that would be nice, but okay otherwise.

We tried sex prediction on this dataset [here](https://www.researchsquare.com/article/rs-3951614/v1), I expect it should work for age as well, but I'm curious to know more about it.

bruAristimunha:
Hey @MohammadJavadD,

Fetch working!

Can you upload your branch and test it to see if the dataset download is working? Too big to be tested in the CI.

bruAristimunha:
I also think that one of your tests is not working.

bruAristimunha (approved):

bruAristimunha:
Thank you so much @MohammadJavadD!

I am super happy with your contribution, and let's continue on the transfer learning tutorial ;)

bruAristimunha:
and btw, thank you @neerajwagh for your google drive!

MohammadJavadD:
Thank you @bruAristimunha for all your help.

neerajwagh:
> and btw, thank you @neerajwagh for your google drive!

No problem! It's worth checking with the NMT authors to see why the link was taken down or not maintained. They could be working on a newer version and can notify you of updates.

bruAristimunha:
Some people have already tried, but no response unfortunately. It appears
that the student has graduated, but not sure.

Anyway, I think it is solved now.

PR #444, opened 2023-04-17
[WIP] Fix cuda problem for DeepSleepNet

tgnassou:
Hi @bruAristimunha, I just saw that I got an issue with my implementation of DeepSleepNet. I forgot to put the hidden layer of the LSTM to the good device leading to a RunTimeError when the training is done on GPU.  

I just fixed it, but I was wondering if there are tests for CPU and CUDA ? It seems that my test was only on CPU, that's why i didn't detect the issue before the merge.

Let me know if I need to write a proper test for GPU.

### DIFF
--- a/braindecode/models/deepsleepnet.py
+++ b/braindecode/models/deepsleepnet.py
@@ -154,8 +154,8 @@ class _BiLSTM(nn.Module):
         # set initial hidden and cell states
         h0 = torch.zeros(
             self.num_layers * 2, x.size(0), self.hidden_size
-        )  # RuntimeError: Input and hidden tensors are not at the same device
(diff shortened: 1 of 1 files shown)


bruAristimunha:
Hi @tgnassou!

Unfortunately there is no way to add a CI with a GPU without cost. 

To the best of my knowledge, there are usually two ways, pay or open a machine on some server and open the port for the CI. It's what they do in Monai/Nvidia, if i remember correctly.

I'm not entirely confident in this information, there may be a way, but I don't know.

Anyway, you can go on with this PR, it is approved for me.

bruAristimunha (approved):

agramfort:
let's make the CI green and merge this one.

PR #446, opened 2023-04-21
update transposition to latest torch convention

matthieutrs:
Updating a transposition to latest torch convention that was raising a deprecation warning message at each call of the transform and clogging training logs.

The warning was:
```
The use of `x.T` on tensors of dimension other than 2 to reverse their shape is deprecated and it will 
throw an error in a future release. Consider `x.mT` to transpose batches of matrices or 
`x.permute(*torch.arange(x.ndim - 1, -1, -1))` to reverse the dimensions of a tensor.
```

This PR implements precisely the second suggestion. I couldn't find any other use of `.T` in the repo but I'd happily update if you point me to others :)

### DIFF
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -501,8 +501,9 @@ def _frequency_shift(X, fs, f_shift):
     analytical = _analytic_transform(padded)
     if isinstance(f_shift, (float, int, np.ndarray, list)):
         f_shift = torch.as_tensor(f_shift).float()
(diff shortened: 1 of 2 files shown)


bruAristimunha:
Nice!

Thank you very much for your PR! Could you register your modification in the what's news file for you to be listed as contributors?

https://raw.githubusercontent.com/braindecode/braindecode/master/docs/whats_new.rst

bruAristimunha:
Thank you @matthieutrs

PR #451, opened 2023-05-22
Fix padding device in EEGResNet

### LINKED ISSUE #450: Wrong padding device in `EEGResNet` if multiple GPUs are available
Hi,
There is an issue with `zeros_for_padding`'s device in `EEGResNet` if `x` is not on the first visible GPU here:

https://github.com/braindecode/braindecode/blob/57e384d4607aa8c04bd574a4b10b2fd0c04d7d08/braindecode/models/eegresnet.py#L235

`zeros_for_padding` is always copied to `cuda:0` while `x` can be on any `cuda:n`. Their concatenation fails.
I will do a fix

PierreGtch:
Fix #450

### DIFF
--- a/braindecode/models/eegresnet.py
+++ b/braindecode/models/eegresnet.py
@@ -229,9 +229,7 @@ class _ResidualBlock(nn.Module):
         stack_2 = self.bn2(self.conv_2(stack_1))  # next nonlin after sum
         if self.n_pad_chans != 0:
             zeros_for_padding = torch.autograd.Variable(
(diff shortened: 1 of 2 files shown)


bruAristimunha:
Hi @PierreGtch!

Nice catch! Can you please update the what's news file? When we get the green in CI, we can merge =)

bruAristimunha (approved):

bruAristimunha:
Thank you @PierreGtch!

PR #454, opened 2023-07-28
Adding Convolutional Transformer model

eeyhsong:


braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+            # pooling acts as slicing to obtain 'patch' along the
+            # time dimension as in ViT
+            nn.Dropout(conv_drop),
+        )
+
+        self.projection = nn.Sequential(
+            nn.Conv2d(
+                kernel, kernel, (1, 1), stride=(1, 1)
+            ),  # transpose, conv could enhance fiting ability slightly
+            Rearrange("b e (h) (w) -> b (h w) e"),
```
martinwimpff, line 152: could be more descriptive
```suggestion
            Rearrange("b d_model 1 seq -> b seq d_model"),
```

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+            fill_value = torch.finfo(torch.float32).min
+            energy.mask_fill(~mask, fill_value)
+
+        scaling = self.emb_size ** (1 / 2)
+        att = F.softmax(energy / scaling, dim=-1)
+        att = self.att_drop(att)
+        out = torch.einsum("bhal, bhlv -> bhav ", att, values)
+        out = rearrange(out, "b h n d -> b n (h d)")
+        out = self.projection(out)
+        return out
```
martinwimpff, line 194: Why do you not use [nn.MultiheadAttention](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html)?

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+            nn.Linear(emb_size, expansion * emb_size),
+            nn.GELU(),
+            nn.Dropout(drop_p),
+            nn.Linear(expansion * emb_size, emb_size),
+        )
+
+
+class GELU(nn.Module):
+    def forward(self, input: Tensor) -> Tensor:
+        return input * 0.5 * (1.0 + torch.erf(input / math.sqrt(2.0)))
```
martinwimpff, line 221: use [nn.GELU](https://pytorch.org/docs/stable/generated/torch.nn.GELU.html) instead

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+    def __init__(self, att_depth, emb_size, att_heads, att_drop):
+        super().__init__(
+            *[
+                TransformerEncoderBlock(emb_size, att_heads, att_drop)
+                for _ in range(att_depth)
+            ]
+        )
+
+
+class ClassificationHead(nn.Sequential):
```
martinwimpff, line 256: Use `nn.Module ` instead if you are writing your model in that way.
```suggestion
class ClassificationHead(nn.Module):
```

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+class TransformerEncoder(nn.Sequential):
+    def __init__(self, att_depth, emb_size, att_heads, att_drop):
+        super().__init__(
+            *[
+                TransformerEncoderBlock(emb_size, att_heads, att_drop)
+                for _ in range(att_depth)
+            ]
+        )
```
martinwimpff, line 253: Use a combination of [nn.TransformerEncoderLayer](https://pytorch.org/docs/stable/generated/torch.nn.TransformerEncoderLayer.html) and [nn.TransformerEncoder](https://pytorch.org/docs/stable/generated/torch.nn.TransformerEncoder.html) instead. Then you can also get rid of `ResidualAdd` and `FeedForwardBlock`. See comment on top.

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+        super().__init__(
+            PatchEmbedding(
+                kernel,
+                kernel_temp_conv,
+                eeg_channel,
+                kernel_avg_pool,
+                stride_avg_pool,
+                conv_drop,
+            ),
+            TransformerEncoder(att_depth, kernel, att_heads, att_drop),
```
martinwimpff, line 111: As mentioned below, use the official torch implementations instead:
```suggestion
            nn.TransformerEncoder(nn.TransformerEncoderLayer(
            d_model=kernel, nhead=att_heads, dim_feedforward=kernel*4, dropout=att_drop, 
            batch_first=True, norm_first=True, activation="gelu"), num_layers=att_depth)
```

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+            eeg_channel,
+            kernel_avg_pool,
+            stride_avg_pool,
+            conv_drop,
+    ):
+        super().__init__()
+
+        self.shallownet = nn.Sequential(
+            nn.Conv2d(1, kernel, (1, kernel_temp_conv), (1, 1)),
+            nn.Conv2d(kernel, kernel, (eeg_channel, 1), (1, 1)),
```
martinwimpff, line 137: Defining the stride is not necessary as 1 is the default.
```suggestion
            nn.Conv2d(1, kernel, (1, kernel_temp_conv)),
            nn.Conv2d(kernel, kernel, (eeg_channel, 1)),
```

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+    Convolutional Transformer for EEG decoding.
+
+    The paper and original code with more details about the methodological
+    choices are availible at the [EEG Conformer]_ and [EEG Conformer Code]_.
+
+    This neural network architecture recieves a different input shape than
+    other models. The input shape should be four-dimensional matrix
+    representing the EEG signals.
+
+         (batch_size, 1, n_channels, n_timesteps)`.
```
martinwimpff, line 26: This is not true, as line 156 unsqueezes the 3D input to 4D.
```suggestion
         (batch_size, n_channels, n_timesteps)`.
```

docs/whats_new.rst
```diff
@@ -41,6 +41,7 @@ Enhancements
 - Adding :class:`braindecode.models.ATCNet` network for motor imagery (:gh:`429` by `Cedric Rommel`_)
 - Adding to :class:`braindecode.datasets.tuh.TUH` compatibility with version 3.0 of TUH dataset (:gh:`431` by `Mohammad Javad D`_, `Bruno Aristimunha`_, `Robin Tibor Schirrmeister`_, `Lukas Gemein`_, `Denis A. Engemann`_ and `Oskar Størmer`_)
 - Adding :class:`braindecode.models.DeepSleepNet` network for sleep staging (:gh:`417` by `Théo Gnassounou`_)
+- Adding :class:`braindecode.models.Conformer` network (:gh:`454` by `Yonghao Song`_)
```
martinwimpff, line 44: I would always (also the file name) call it EEGConformer for consistency.
Also: strictly speaking, the EEGConformer is not a "real" [Conformer ](https://arxiv.org/pdf/2005.08100.pdf).

martinwimpff (changes requested):
@eeyhsong please check if your custom attention modules can be replaced by the torch version. If not, please let us know where your implementation differs and why it is important.
Thank you for contributing to braindecode!

docs/whats_new.rst
eeyhsong: Thanks for your clarification! It's okay to call it EEGConformer. It's just for short before.

braindecode/models/conformer.py
eeyhsong: Okay. Thanks for your help!

braindecode/models/conformer.py
eeyhsong: It's okay. Thank you!

braindecode/models/conformer.py
eeyhsong: Okay, thank you for the modification.

braindecode/models/conformer.py
eeyhsong: I implemented the Transformer part in my previous study to adjust it easier, instead of using the official torch version. Please wait for me a little time to check.

braindecode/models/conformer.py
```diff
@@ -0,0 +1,279 @@
...
+
+    Parameters PatchEmbedding
+    -------------------------
+    - n_kernel: int
+        Number of kernels for the temporal convolution layer (first layer).
+    - n_kernel_temp_conv: int
+        Kernel size of the temporal convolution layer.
+    - eeg_channel: int
+        Number of EEG channels (kernel size of the spatial convolution layer).
+    - n_kernel_avg_pool: int
```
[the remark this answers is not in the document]
bruAristimunha: fixed @martinwimpff

docs/whats_new.rst
bruAristimunha: Fixed @eeyhsong and @martinwimpff

bruAristimunha:
Hello @eeyhsong and @martinwimpff,

To speed up the integration process, I corrected some things that I suggested were wrong and some things that @martinwimpff pointed out.

Also, I realized that there is an error in the library unrelated to this PR that is being handled in the PR. 

@eeyhsong, I think the fundamental thing missing is to see if we can replace attention with the native module without any problem, and the suggestion @martinwimpff made about Pytorch's native module for the transformer part.

After that, tests. Hope we can finish it by Friday :)

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,272 @@
...
+class EEGConformer(nn.Sequential):
+    """EEG Conformer.
+
+    Convolutional Transformer for EEG decoding.
+
+    The paper and original code with more details about the methodological
+    choices are availible at the [EEG Conformer]_ and [EEG Conformer Code]_.
+
+    This neural network architecture recieves a different input shape than
+    other models. The input shape should be four-dimensional matrix
```
sliwy, line 21: You specify `four-dimensional` here but there are only three dimensions below: `(batch_size, n_channels, n_timesteps)`

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,272 @@
...
+
+    .. versionadded:: 0.8
+
+    We aggregate the parameters based on the parts of the models.
+
+    Parameters PatchEmbedding
+    -------------------------
+    - n_kernel: int
+        Number of kernels for the temporal convolution layer (first layer).
+    - n_kernel_temp_conv: int
```
sliwy, line 45: if we speak about kernel size I would prefer notation with `len` in the name. `n_kernel_temp_conv` and other like that may suggest that this corresponds to the number of kernels

sliwy: Similar thing for the `spatial_conv` and pooling layers. I would recommend to keep similar naming to the parameters in `ShallowFBCSPNet` as it is almost the same (also I've found strong link to the `ShallowFBCSPNet` in the article)

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,272 @@
...
+            n_kernel,
+            n_kernel_temp_conv,
+            n_kernel_spatial_conv,
+            n_kernel_avg_pool,
+            stride_avg_pool,
+            drop_prob,
+    ):
+        super().__init__()
+
+        self.shallownet = nn.Sequential(
```
sliwy, line 133: How about using `ShallowFBCSPNet` already implemented in the package? To remove `conv_classifier` and other not needed layers you may use sth like that:
```python
model = ShallowFBCSPNet(in_chans=22, 40, conv_nonlin=nn.ELU(), first_conv_bias=True)
new_model = nn.Sequential()
for name, module_ in model.named_children():
    if "conv_classifier" in name:
        break
    new_model.add_module(name, module_)
model = new_model
```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,272 @@
...
+            TransformerEncoder(att_depth, n_kernel, att_heads, att_drop_prob),
+            ClassificationHead(n_kernel, fc_dim, n_classes),
+        )
+
+
+class PatchEmbedding(nn.Module):
+    """Patch Embedding.
+
+    The authors used a convolution moduleto capture local features,
+    instead of postion embedding.
```
sliwy, line 118:
```suggestion
    instead of position embedding.
```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,272 @@
...
+        )
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = x.unsqueeze(dim=1)  # add one extra dimension
+        x = self.shallownet(x)
+        x = self.projection(x)
+        return x
+
+
+class MultiHeadAttention(nn.Module):
```
sliwy, line 160: If this module and the rest below are used only internally in this file they should be private classes with `_` as first letter in the name, e.g., `_MultiHeadAttention`

sliwy:
Hi @eeyhsong, congrats for the PR! Nice that we can have one more model :)

I added some comments. @eeyhsong do you think you may use `ShallowFBCSPNet` model already implemented in braindecode with some specific parameters to match your article? Then we may reduce redundant code and also it may be easier to make link between your model and `ShallowFBCSPNet` by Robin.

braindecode/models/conformer.py
eeyhsong: Hello， @martinwimpff and @bruAristimunha
Thanks for your help!
I've compared torch implementation and manual implementation (running two programs on two same GPUs). The memory (~7G->9G) and time consumption (12min -> 13 min per subject) were raised with the torch version. The accuracy decreased by 4.xx% on the BCI competition IV 2a dataset.
So I suggest using the original implementation, which could also help users to figure out the network design. 

If you still prefer the torch version, I may need more time to compare the difference in detail.

martinwimpff: Can you please check the number of parameters? Just to see if both methods yield the same architecture. 

One difference that I know of is the weight initialization of `nn.MultiheadAttention` vs. the weight initialization of your `nn.Linear` layers (xavier_uniform vs. kaiming_uniform). So what you could also check is whether the acc decrease is just coming from this weight init.

Strange to see that the torch version needs more memory as it usually should be better optimized and therefore more efficient.

braindecode/models/eegconformer.py
bruAristimunha: What do you think @eeyhsong? Make sense for you?

bruAristimunha: I was thinking more about this, and for me, it is best if we keep separate in another PR for two reason:
- 1) Given the tokenization process, the number of channels is always "equals" 1.  So a different network input as the traditional input in the ShallowNet. When compared with the original shallow, it potentially can lead to a different network, as already pointed out in the case of the attention layer and/or the addition of the transformer layer.
- 2) ShallowNet will be refactored soon in code sprint by @robintibor to remove the 4d entry used in the cropped option.

All these points can be optimized later with a better standardizing of the variables and modules.

braindecode/models/eegconformer.py
bruAristimunha: I think I solved this name variable issue. Can you please check @sliwy?

braindecode/models/conformer.py
bruAristimunha: Hey @martinwimpff and @eeyhsong,

To speed it up a bit, can we address this issue in another issue/pull request? I think this question can be extended to other models in the library.

bruAristimunha:
Hi @eeyhsong!

So, I tried my best to integrate the model into braindecode, and now we need to fix the tested and solved the questions from @sliwy and @martinwimpff.

I was testing the model with different shapes, and it looks like there is a slight problem with the transformer layer. Would you take a look and fix it?

Possibly debugging, you can find what's wrong. I think it's some parameter I missed or the tokenization. My debug code is below:

```python
import torch

from braindecode.models import EEGConformer

batch_size = 16
n_channels = 32
n_timesteps = 1000

model =  EEGConformer(n_classes=2, n_channels=n_channels)

X =  torch.rand(batch_size, n_channels, n_timesteps)

model(X)
```


I'm trying to speed up the integration for this week, as I don't know if I'll be available for the PR next week. Can you help me integrate your model this week?

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,313 @@
...
+    Convolutional Transformer for EEG decoding.
+
+    The paper and original code with more details about the methodological
+    choices are available at the [Song2022]_ and [ConformerCode]_.
+
+    This neural network architecture recieves a traditional braindecode input.
+    The input shape should be three-dimensional matrix representing the EEG
+    signals.
+
+         (batch_size, n_channels, n_timesteps)`.
```
sliwy, line 24:
```suggestion
         `(batch_size, n_channels, n_timesteps)`.
```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+
+    def __init__(
+            self,
+            n_classes,
+            n_channels,
+            n_filters_conv=40,
+            n_filters_time=25,
+            filter_time_length=25,
+            n_kernel_avg_pool=75,
```
sliwy, line 87: `n_kernel_avg_pool` is missing in the docstring - I would suggest sorting the params in the docstring according to the order in the code

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,313 @@
...
+        x = self.patch_embedding(x)
+        x = self.transformer(x)
+        x = self.classification_head(x)
+        return x
+
+
+class PatchEmbedding(nn.Module):
+    """Patch Embedding.
+
+    The authors used a convolution moduleto capture local features,
```
sliwy, line 124:
```suggestion
    The authors used a convolution module to capture local features,
```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,313 @@
...
+            n_classes=n_classes)
+
+    def forward(self, x: Tensor) -> Tensor:
+        x = self.patch_embedding(x)
+        x = self.transformer(x)
+        x = self.classification_head(x)
+        return x
+
+
+class PatchEmbedding(nn.Module):
```
sliwy, line 121: Shouldn't this also be private class? If it should be public we may add docs for arguments.

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,313 @@
...
+            Number of classes for classification.
+        drop_prob_1 : float
+            Dropout probability for the first dropout layer.
+        drop_prob_2 : float
+            Dropout probability for the second dropout layer.
+        out_channels : int
+            Number of output channels for the first linear layer.
+        hidden_channels : int
+            Number of output channels for the second linear layer.
+        """
```
sliwy, line 286: Missing docstring for `return_features` - do we use that anywhere?
```suggestion
        return_features : bool
            Whether to return input features.
        """
```

sliwy:
Hey @bruAristimunha, I found out what's the problem with the code. Actually, `emb_size` for `self.transformer` and `self.classification_head` should be equal to `n_filters_time` and for now we are providing there another param - `n_filters_conv`. I will fix that by having just one parameter (it will be the same as in @eeyhsong original implementation).

braindecode/models/eegconformer.py
bruAristimunha: I added, but I converted to private. Make more sense to used as private.

bruAristimunha:
Thank you so much @sliwy! I owe you a beer or wine ;p

I fixed my draft tests. If it turns green, I think I have nothing else to add.

braindecode/models/conformer.py
bruAristimunha: It will handle in #463.

eeyhsong:
Hi @bruAristimunha, 
Very sorry for not helping with this problem. Is there any other thing I can help?
Greatly appreciate for compensating a lot of details to clarify the parameters. 🙏🙏

bruAristimunha:
Hey @eeyhsong,

Don't worry. I have two simple questions for you, and if the reviewers are happy, I think we can integrate it into braindecode.

1) what is my method's limitation on the number of channels?

2) What is the model's limitation on the window size?

I want to do the tests with these two parameters... try to indicate to the user what parameter needs to be reduced to work in any case.

However, I don't think I understand very well how to handle the edge cases 1 channels or 128/256 channels, and small time windows like 100 points or very large windows 10000 points. Can you help me?
Could be something like, the model does not work with 1 channel. To work with small windows you need reduce x1, x2 parameters. 

Best,

bruAristimunha:
This practical intuition is perfect!

Please add in the model documentation.

 I recommend also putting it directly as a warning if the input is something different than expected (too many channels).

Example: https://github.com/braindecode/braindecode/blob/master/braindecode/models/sleep_stager_eldele_2021.py#L79

With that I'm more than happy to merge :)

sliwy:
@bruAristimunha apart from the format changes in one file everything looks good to me!

test/unit_tests/models/test_models.py
```diff
@@ -12,9 +12,11 @@
 
 from braindecode.models import (
     Deep4Net, EEGNetv4, EEGNetv1, HybridNet, ShallowFBCSPNet, EEGResNet, TCN,
-    SleepStagerChambon2018, SleepStagerBlanco2020, SleepStagerEldele2021, USleep,
-    DeepSleepNet, EEGITNet, EEGInception, EEGInceptionERP, EEGInceptionMI, TIDNet, ATCNet)
-
+    SleepStagerChambon2018, SleepStagerBlanco2020, SleepStagerEldele2021,
```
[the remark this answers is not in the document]
bruAristimunha: Yes, Pycharm integrated autofix. I reverted.

eeyhsong:
Hi @bruAristimunha,
I've added commits to the channel warning, and a suggestion on the data scale. Please check again.
🤝🤝

bruAristimunha (approved):
LGTM :)

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,357 @@
...
+
+    The EEG Conformer architecture is composed of three modules:
+        - PatchEmbedding
+        - TransformerEncoder
+        - ClassificationHead
+
+    Notes
+    -----
+    The authors recommend using augment data before using Conformer, e.g. S&R,
+    at the end of the code.
```
agramfort, line 35:
```suggestion
    The authors recommend using data augmentation before using Conformer, e.g. S&R,
    at the end of the code.
```
what is S&R? can you spell it out?

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,357 @@
...
+    n_classes: int
+        Number of classes to predict (number of output filters of last layer).
+    n_channels: int
+        Number of channels to be used as number of spatial filters.
+    n_filters_time: int
+        Number of temporal filters, defines also embedding size.
+    filter_time_length: int
+        Length of the temporal filter.
+    pool_time_length: int
+        Length of temporal poling filter.
```
agramfort, line 58:
```suggestion
        Length of temporal pooling filter.
```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,357 @@
...
+        Number of self-attention layers.
+    att_heads: int
+        Number of attention heads.
+    att_drop_prob: float
+        Dropout rate of the self-attention layer.
+    final_fc_length: int
+        The dimension of the fully connected layer.
+    return_features: bool
+        If True, the forward method returns the features before the
+        last classification layer. Defaults to False.
```
agramfort, line 73:
```suggestion
        last classification layer. Defaults to False.

```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,357 @@
...
+        x = self.transformer(x)
+        x = self.classification_head(x)
+        return x
+
+
+class _PatchEmbedding(nn.Module):
+    """Patch Embedding.
+
+    The authors used a convolution module to capture local features,
+    instead of postion embedding.
```
agramfort, line 136:
```suggestion
    instead of position embedding.
```

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,350 @@
...
+            # time dimension as in ViT
+            nn.Dropout(p=drop_prob),
+        )
+
+        self.projection = nn.Sequential(
+            nn.Conv2d(
+                n_filters_time, n_filters_time, (1, 1), stride=(1, 1)
+            ),  # transpose, conv could enhance fiting ability slightly
+            Rearrange("b d_model 1 seq -> b seq d_model"),
+        )
```
martinwimpff, line 192: Possible extension: additional parameter `embedding_size`. Otherwise the projection layer is not necessary.

martinwimpff (commented):
Nice work @eeyhsong ! 👍

bruAristimunha:
Hi @eeyhsong!

I corrected the tests I had created. I think you are very close to iterating your model to braindecode.

I was chatting with @martinwimpff, and the only point is that we are looking to switch to PyTorch for any possible part of the model. To minimize what we need to maintain and ensure that PyTorch guarantees sustainability for as many versions as possible. This will be better for your more general model usage. 

In this context, can you please provide a mini-code version with the code you tested to native layers? 🙏🏽 It could be a comment on this GitHub PR, just for us to try it out and get on the same page about using personality layers.

I appreciate your effort on this thread, and we're almost done after these review rounds.

eeyhsong:
Hi @bruAristimunha!
I've upload an implementation Line 306-332. You may test with the torch version. I'm not sure if it could be more simple with `nn.`.
If you insist for the torch version, I think its okay, for more general usage.

Sincerey,

bruAristimunha:
Thank you @eeyhsong! 

@martinwimpff and I will take a look :)

braindecode/models/eegconformer.py
```diff
@@ -0,0 +1,378 @@
...
+    att_heads : int
+        Number of attention heads.
+    att_drop : float
+        Dropout probability for the attention layers.
+
+    """
+    def __init__(self, att_depth, emb_size, att_heads, att_drop, forward_expansion=4):
+        super().__init__()
+        self.enc_lay = nn.TransformerEncoderLayer(d_model=emb_size, nhead=att_heads, dim_feedforward=emb_size*forward_expansion, dropout=att_drop)
+        self.enc = nn.TransformerEncoder(self.enc_lay, num_layers=att_depth)
```
martinwimpff, line 328:
```suggestion
        enc_lay = nn.TransformerEncoderLayer(d_model=emb_size, nhead=att_heads, dim_feedforward=emb_size*forward_expansion, dropout=att_drop, batch_first=True, norm_first=True, activation="gelu)
        self.enc = nn.TransformerEncoder(enc_lay, num_layers=att_depth)
```

martinwimpff: This should be equivalent (apart from the weight initialization).

eeyhsong:
Thanks for your time. 🤝🤝

bruAristimunha:
@martinwimpff, If you approve your review, the PR will be merge =)

eeyhsong:
Hi @bruAristimunha, you've done a detailed and clear comparison. 
I also learned usages for braindecode with your code. I think I will strongly recommend my colleagues to use it.

Thanks again for your effort!

martinwimpff (approved):

bruAristimunha:
Online @eeyhsong ;) https://braindecode.org/dev/generated/braindecode.models.EEGConformer.html

eeyhsong:
Got it, thanks! 🤝🤝

PR #468, opened 2023-08-18
Replacing Pytorch layers to Rearrange from einops

bruAristimunha:


braindecode/models/eegnet.py
```diff
@@ -178,20 +180,13 @@ def __init__(
         self.add_module("softmax", nn.LogSoftmax(dim=1))
         # Transpose back to the the logic of braindecode,
         # so time in third dimension (axis=2)
-        self.add_module("permute_back", Expression(_transpose_1_0))
+        self.add_module("permute_back",
+                        Rearrange("batch 1 ch t -> batch ch t 1"))
```
dcwil, line 184: This should be `Rearrange("batch 1 ch t -> batch 1 t ch") right?

braindecode/models/eegnet.py
```diff
@@ -328,7 +323,7 @@ def __init__(
         # Transpose back to the the logic of braindecode,
         # so time in third dimension (axis=2)
         self.add_module(
-            "permute_2", Expression(lambda x: x.permute(0, 1, 3, 2))
+            "permute_2",  Rearrange("batch 1 ch t -> batch ch t 1")
```
dcwil, line 326: This should be ".. -> batch 1 t ch" ?

braindecode/models/tidnet.py
```diff
@@ -131,22 +132,9 @@ def __init__(self, s_growth, t_filters, in_chans, input_window_samples, drop_pro
...
-            to dim:
-            batch, 1, chans, time
-
-            """
-            return x.permute([0, 3, 1, 2])
-
         self.temporal = nn.Sequential(
             Ensure4d(),
-            Expression(_permute),
+            Rearrange("batch C T 1 -> batch 1 T C"),
```
dcwil, line 137: ".. -> batch 1 C T"?

dcwil (changes requested):
I think there's 3 or so where the einop output isn't quite right

braindecode/models/eegnet.py
```diff
@@ -79,8 +80,8 @@ def __init__(
         self.add_module("ensuredims", Ensure4d())
         # b c 0 1
         # now to b 1 0 c
```
robintibor, line 82: this comment is and was wrong then? maybe just remove it.

bruAristimunha: Removed @robintibor

requirements.txt
```diff
@@ -3,7 +3,7 @@ pandas
 scipy
 matplotlib
 h5py
-skorch<0.14
+skorch<0.13
```
agramfort, line 6: do we still need this? why? can you add comment here?

braindecode/models/eegnet.py
```diff
@@ -328,7 +329,7 @@ def __init__(
         # Transpose back to the the logic of braindecode,
         # so time in third dimension (axis=2)
         self.add_module(
-            "permute_2", Expression(lambda x: x.permute(0, 1, 3, 2))
+            "permute_2",   Rearrange("batch x y z -> batch x z y")
```
agramfort, line 332:
```suggestion
            "permute_2", Rearrange("batch x y z -> batch x z y")
```

requirements.txt
bruAristimunha: I reversed @agramfort, but let me explain better why I started restricting the libraries. 

Not really sure why, but there was an API break between 0.13 and 0.14 on skorch. An argument has become mandatory in NeuralNetwork, and the fit method now requires y or something like that.

We noticed with new users, during the code sprint in Basel, that when they tried to install and run the braindecode tests, a lot of errors came up related to this skorch issue.

Another problem is that conda takes longer (+2 hours in some computers) to create a braindecode env. I solved this in CI by switching to Mamba, but installing Mamba on new users' computers takes a lot of work and time. There was one participant who couldn't even install braindecode. This is why I'm tending to restrict the skorch version.

agramfort:
ok let's move on here but I would suggest to open an issue so support of
newer version is in the roadmap.
If you think skorch devs are creating more problems than they should you
should contact them eg by pinging
them in your issue.

bruAristimunha:
> ok let's move on here but I would suggest to open an issue so support of newer version is in the roadmap. If you think skorch devs are creating more problems than they should you should contact them eg by pinging them in your issue.
> […](#)
> Message ID: ***@***.*** com>

Okay @agramfort, I will ping next time. 

Can we merge @dcwil, @robintibor and @agramfort?

agramfort:
thx @bruAristimunha

bruAristimunha:
Hi Dan (@dcwil)!

I need your review or approval to be able to merge the PR :)

PR #469, opened 2023-08-18
Adding poetry and changing the CI

bruAristimunha:


agramfort:
@bruAristimunha pinning versions with == is a bad idea. It makes dev environnements impossible to update and most of us don't use a dedicated env for braindecode. I would suggest to use >= for versions.

robintibor:
tend to agree with @agramfort

bruAristimunha:
Closed @agramfort and @robintibor.

agramfort:
I don't say we should close. You can still use poetry and not a setup
file but don't pin version numbers with == or <= as much as possible.
And I don't think lock files are necessary for us here.

bruAristimunha:
I will start again!

PR #491, opened 2023-09-04
DOC Plot regression

Sara04:
Update plot regression example.


bruAristimunha:
Thank you @Sara04 =)

Sara04:
Thanks, @bruAristimunha ! :)

PR #501, opened 2023-09-05
WIP Unify the EEGClassifier and EEGRegressor

bruAristimunha:


braindecode/eegneuralnet.py
```diff
@@ -0,0 +1,153 @@
...
+            )
+        named_by_user = True
+        train_valid_callbacks = [
+            (train_name, train_scoring, named_by_user),
+            (valid_name, valid_scoring, named_by_user)
+        ]
+        return train_valid_callbacks
+
+    # pylint: disable=arguments-differ
+    def get_loss(self, y_pred, y_true, *args, **kwargs):
```
PierreGtch, line 62: Why do we have this method? Could we remove it?

braindecode/regressor.py
```diff
@@ -70,155 +63,13 @@ def __init__(self, *args, cropped=False, callbacks=None,
...
-                BatchScoring(
-                    valid_loss_score, name="valid_loss", target_extractor=noop,
-                ),
-            ),
-            ("print_log", PrintLog()),
-        ]
-
     # Method added to fix `predict_proba` behavior to return proper values
     # in cropped mode even if regressor does not return probabilities.
     # We implement `predict_proba` because it exists in skorch.NeuralNetRegressor.
```
PierreGtch, line 75: What does this comment refer to?

braindecode/eegneuralnet.py
sliwy: It was introduced to counter the skorch implementation of `NeuralNetClassifier.get_loss`. In this method https://github.com/skorch-dev/skorch/blob/35cbaa73892c42f1d5131cdd406246b618f0aecd/skorch/classifier.py#L144-L150 they apply log to `y_pred` when using `NLLoss` which in our case would break the model because we would have LogSoftmax in models and then an additional Log applied in skorch.

This is definitely not needed in the case of `EEGREgressor`.

braindecode/regressor.py
bruAristimunha: doesn't make sense

braindecode/eegneuralnet.py
bruAristimunha: Removed in EEGRegressor

braindecode/regressor.py
sliwy: this comment refers to the existence of `predict_proba` for Regressor -> it doesn't make much sense to have `predict_proba` method for regressors as there is no probability returned. We actually implemented the `predict_proba` because it was implemented in skorch also for regressor and in the case we don't implement it here, then someone could use `predict_proba` not adjusted to cropped decoding.

braindecode/classifier.py
```diff
@@ -70,155 +63,13 @@ def __init__(self, *args, cropped=False, callbacks=None,
...
-            (
-                "valid_loss",
-                BatchScoring(
-                    valid_loss_score, name="valid_loss", target_extractor=noop,
-                ),
-            ),
-            ("print_log", PrintLog()),
-        ]
-
     def predict_proba(self, X):
```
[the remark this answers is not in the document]
bruAristimunha: It didn't work @PierreGtch. I tried to fix it, but it seemed like I would need to change the callback parser more. I preferred to leave it for later, given that this is not really the main objective of PR.

PR #502, opened 2023-09-05
Adding pre commit

bruAristimunha:


sliwy:
@bruAristimunha I dream about having pre-commit in braindecode -> how about applying it after merging all code sprint PRs?

second thing, maybe we can agree on some sort of formatter like black? I remember we discussed that with @robintibor and @agramfort but the consensus was that we don't do that, anything changed from that moment?

agramfort:
black has won the battle for me :)

.pre-commit-config.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+
+  -   repo: https://github.com/asottile/blacken-docs
+      rev: 1.15.0
+      hooks:
+      -   id: blacken-docs
+          additional_dependencies: [black==23.3.0]
+          exclude: ^.github/
+
+  - repo: https://github.com/PyCQA/isort
```
sylvchev, line 43: I think it could be better to use only ruff and remove isort and flake8 precommit, since ruff could do all the job. I could help you with that.

sylvchev (commented):
Nice changes for the import, need to validate the "black for the docs" part with Alex.

.pre-commit-config.yaml
bruAristimunha: Okay.. Changing everything in favor of ruff.

Issue #477, opened 2023-09-02
Augmentations should work on single examples


robintibor:
This is promised in our doc and would be useful but does not seem to be the case, probably because of the example-wise mask created in the base augmentation class.

bruAristimunha:
If you join this week @cedricrommel, a nice issue for you =)

cedricrommel:
Closing this

PR #503, opened 2023-09-06
Get data augmentation to work on single examples

cedricrommel:
This PR intends to fix issue https://github.com/braindecode/braindecode/issues/477.
It allows to augment single inputs which do not have a batch dimension, just channles and time dimensions.

### DIFF
--- a/braindecode/augmentation/base.py
+++ b/braindecode/augmentation/base.py
@@ -72,9 +72,16 @@ class Transform(torch.nn.Module):
             Transformed labels. Only returned when y is not None.
         """
         X = torch.as_tensor(X).float()
+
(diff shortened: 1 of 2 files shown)


cedricrommel:
So, the slight converage drop comes from the following change: we now import create_from_mne_epochs from braindecode.datasets instead of braindecode.datautil in the augmentation/test_base.py

This should be a good thing, since we should not import from datautil anymore as I understand ^^

What should I do? Keep with the old bad import which does not impact the coverage or accept this slight coverage drop? I prefer the latter.

PierreGtch (approved):

bruAristimunha:
Thank you @cedricrommel

PR #506, opened 2023-09-06
Replace temporal and spatial convolutions with single merged convolution

dcwil:


bruAristimunha:
Old:
loss_curve_old

New:
loss_curve_new

Same confusion matrix! Kudo @dcwil!
confusion_new

bruAristimunha:
Btw, @Sara04 and @dcwil, I think I should use this to do something (build a paper maybe) or for more methods...

PR #507, opened 2023-09-06
ENH accelerate load_concat_dataset by pickling unloaded raw object

tomMoral:


braindecode/datautil/serialization.py
```diff
@@ -114,12 +116,33 @@ def _load_signals_and_description(path, preload, is_raw, ids_to_load=None):
 
 
 def _load_signals(fif_file, preload, is_raw):
```
sliwy, line 118: maybe we could add type hints when modifying this function? reading the code could be much easier then

braindecode/datautil/serialization.py
```diff
@@ -114,12 +116,33 @@ def _load_signals_and_description(path, preload, is_raw, ids_to_load=None):
...
+
+    # Reading the raw file from pickle if it has been save before.
+    # The pickle file only contain the raw object without the data.
+    pkl_file = fif_file.with_suffix(".pkl")
+    if pkl_file.exists():
```
sliwy, line 123: maybe we can add some logging that we'are loading from pickle? I imagine situation in which you modify raw files and you're stuck with old pickle files.

braindecode/datautil/serialization.py
```diff
@@ -146,6 +169,8 @@ def load_concat_dataset(path, preload, ids_to_load=None, target_name=None,
     -------
     concat_dataset: BaseConcatDataset of BaseDatasets or WindowsDatasets
     """
+    path = Path(path)
```
sliwy, line 172: I would allow in the parameters to pass `pathlib.Path` also as path parameter.

tomMoral: Did this! :)

braindecode/datautil/serialization.py
tomMoral: Maybe for another PR.

tomMoral:
Thanks for the review @sliwy ! Appreciated.

Issue #511, opened 2023-09-07
Remove LogSoftmax layer at the end


PierreGtch:
We will remove the `LogSoftmax` layer at the end of every model.

In order not to break the API, we will add a parameter `add_logsoftmax` to the models, true by default. If the argument is true, we raise a deprecated warning.

PierreGtch:
closed by #513

PR #513, opened 2023-09-07
Remove logsoftmax

Sara04:
Addressing the issue  #511 by setting up the flag add_logsoftmax with a warning that the LogSoftmax as a final layer will be removed in the future. Proposed by @PierreGtch .


braindecode/models/tcn.py
```diff
@@ -50,7 +48,6 @@ def __init__(
             n_filters=None,
             kernel_size=None,
             drop_prob=None,
-            add_log_softmax=None,
```
sliwy, line 53: this parameter should stay here but with default value set to True (to maintain the behavior implemented now in this PR). It should be also back in the docstring with small adjustement. Value of `add_log_softmax` should be used in `super().__init__`

braindecode/models/tcn.py
```diff
@@ -29,8 +29,6 @@ class TCN(EEGModuleMixin, nn.Module):
         kernel size of the convolutions
     drop_prob: float
         dropout probability
-    add_log_softmax: bool
```
sliwy, line 32: it should be restored in the docstring as well

braindecode/models/usleep.py
```diff
@@ -178,7 +175,6 @@ def __init__(
             input_window_seconds=30,
             time_conv_size_s=9 / 128,
             ensure_odd_conv_size=False,
-            apply_softmax=False,
```
sliwy, line 181: we should add `add_log_softmax` param here, in the docstring and use it in `super().__init__`

braindecode/models/deep4.py
```diff
@@ -138,6 +138,7 @@ def __init__(
             n_times=n_times,
             input_window_seconds=input_window_seconds,
             sfreq=sfreq,
+            add_log_softmax=True,
```
sliwy, line 141: it should be added to params of the model in the docstring, and used  here

sliwy (dismissed):
Thx @Sara04 Great job! I like the idea of convention of having `out_fun` and the approach of LogSoftmax/Identity. Some details I want to discuss:
1. I have doubts whether property called `add_log_softmax` is not misleading, suggesting that there is an action of adding log_softmax being performed. I have commented with some propositions (@bruAristimunha, @PierreGtch  please tell what you think as well).
2. If I understand well this PR we need to add `add_log_softmax` param in all the models as a parameter and then use it in the `super.__init__()`. In most of the models it is always set to True in `super.__init__()`, I left comments for some of the cases here. 

Edit: I didn't notice that it is a draft, so maybe reviewed too early :joy:

Sara04:
Thank you very much for your comments, @sliwy ! :smile: 
I will reply to them one by one.

braindecode/models/base.py
```diff
@@ -158,6 +165,17 @@ def sfreq(self):
             )
         return self._sfreq
 
+    @property
+    def add_log_softmax(self):
+        if self._add_log_softmax is None:
+            raise ValueError("add_log_softmax flag not specified.")
+
+        if self._add_log_softmax:
```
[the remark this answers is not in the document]
Sara04: So far, in a great number of models, LogSoftmax non-linearity was hardcoded in the model. We would like to remove it, but preserve compatibility with the old definition of the models in the following 1-2 years. The add_log_softmax is a temporary flag that will be removed. The users who are used to old API do not need to change anything, but will get the warning that in the future LogSoftmax as output function will be removed.

braindecode/models/eegconformer.py
```diff
@@ -377,7 +382,10 @@ def __init__(self, final_fc_length, n_classes,
             nn.Linear(hidden_channels, n_classes),
         )
         self.return_features = return_features
-        self.classification = nn.LogSoftmax(dim=1)
+        if add_log_softmax:
+            self.classification = nn.LogSoftmax(dim=1)
```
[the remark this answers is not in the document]
Sara04: Yes, why not

PierreGtch:
Hi @sliwy, thanks for your review! We discuss your comments with Sara and I will answer them individually

braindecode/models/base.py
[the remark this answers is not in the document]
PierreGtch: Agree with @sliwy , `add_log_softmax` should be either true of false. We can ignore the case if the user passes `None`

braindecode/models/base.py
```diff
@@ -158,6 +165,17 @@ def sfreq(self):
             )
         return self._sfreq
 
+    @property
+    def add_log_softmax(self):
```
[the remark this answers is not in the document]
PierreGtch: I think it’s best to keep the property name like that to stick to the model parameter name

braindecode/models/eegconformer.py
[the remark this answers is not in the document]
PierreGtch: It is going to be deprecated so I don’t think we should bother.
Plus, we don’t want it to become a feature, we want the users to see the warning, update their scripts and set `add_log_softmax=False`

braindecode/models/base.py
```diff
@@ -158,6 +165,17 @@ def sfreq(self):
...
+        if self._add_log_softmax is None:
```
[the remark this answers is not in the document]
Sara04: Removed! Thanks!

braindecode/models/hybrid.py
```diff
@@ -37,19 +41,37 @@ def __init__(self, in_chans, n_classes, input_window_samples):
...
+            n_chans,
+            n_outputs,
+            n_times=n_times,
             n_filters_time=30,
             n_filters_spat=40,
             filter_time_length=28,
             final_conv_length=29,
         )
 
+        n_chans, n_outputs, n_times = deprecated_args(
```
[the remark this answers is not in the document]
Sara04: Fixed! Thanks!

braindecode/models/deep4.py
Sara04: Fixed, thanks!

braindecode/models/usleep.py
Sara04: Fixed, thanks!

PierreGtch:
> @Sara04 @PierreGtch based on this [#513 (comment)](https://github.com/braindecode/braindecode/pull/513#discussion_r1319780333) I think that I may not understand well what the flow of logsoftmax is going to look like. For now, all models (or almost all) are created with softmax in the end. In the new flow if a user wants to have classification model would need to add softmax on their own?

Yes exactly, the user will have to add the softmax himself as this is standard practice. But in practice, you will not need to because if you use the `nn.CrossEntropyLoss`, it’s already integrated in the loss.

braindecode/models/base.py
```diff
@@ -158,6 +169,15 @@ def sfreq(self):
             )
         return self._sfreq
 
+    @property
+    def add_log_softmax(self):
+        if self._add_log_softmax:
+            warnings.warn("LogSoftmax final layer will be removed! " +
```
sliwy, line 175: In this warning, can we specify that this will be changed behavior of the package and the logsoftmax won't be removed from this particular object when accessing the property?

examples/advanced_training/plot_bcic_iv_4_ecog_cropped.py
```diff
@@ -181,12 +181,11 @@
     n_chans,
     n_classes,
     final_conv_length=2,
+    add_log_softmax=False,
 )
-# We are removing the softmax layer to make it a regression model
+
 new_model = torch.nn.Sequential()
```
sliwy, line 187: we can remove this whole snippet -> then the model will be of proper type and we can use all the `EEGModuleMixin` features, including `to_dense_prediction_model` and `get_output_shape`

braindecode/models/tcn.py
bruAristimunha: With metaclass=NumpyDocstringInheritanceInitMeta from EEGMixer, There is not need

examples/advanced_training/plot_bcic_iv_4_ecog_cropped.py
bruAristimunha: okay!

bruAristimunha:
Can we merge this @Sara04, @PierreGtch and @sliwy? Looks okay for me.

braindecode/models/usleep.py
```diff
@@ -178,12 +175,12 @@ def __init__(
...
             chs_info=None,
             n_times=None,
             in_chans=None,
             n_classes=None,
             input_size_s=None,
+            add_log_softmax=True,
```
PierreGtch, line 183:
```suggestion
            add_log_softmax=False,
```

braindecode/models/tcn.py
```diff
@@ -50,12 +48,12 @@ def __init__(
             n_filters=None,
             kernel_size=None,
             drop_prob=None,
-            add_log_softmax=None,
             chs_info=None,
             n_times=None,
             input_window_seconds=None,
             sfreq=None,
             n_in_chans=None,
+            add_log_softmax=True,
```
PierreGtch, line 56:
```suggestion
            add_log_softmax=False,
```
As it was default to false before, we should leave the same default here

PierreGtch (changes requested):
Just two small changes in case the `add_log_softmax` was already there before and defaulted to false: we should keep the same default.

Otherwise all good to merge! Thanks a lot @Sara04 for all this work!!!

bruAristimunha:
Thank you @Sara04, @PierreGtch, @sliwy!

Sara04:
Thanks @PierreGtch , @sliwy , @bruAristimunha !

Issue #432, opened 2023-02-07
Document or check that number of samples is large enough


Ardillen66:
Hello,

I was experimenting with Deep4Net and came upon an undocumented issue. I tried using an input window of 64 channels by 375 samples and while this worked with ShallowFBCSPNet, I got errors from PyTorch about a mismatch in kernel size compared to the input (9x1 input for 10x1 kernel in this case). I tracked this down as a problem with the input_window_samples dimension that became too small in the 4th Conv layer of the model.

The solution to this problem was to either decrease the kernel size for the 4th Conv layer or to increase the number of input samples.

I suggest adding a note in the documentation to avoid users spending time trying to track down this specific issue. Maybe there could be a recommended minimal input size for each model. Ideally, there would also be a check that can give a sensible error to quickly resolve this issue. Another solution would be to introduce padding where necessary, but I am not familiar enough with the models to do this myself.

If I have the time, I will try to add this myself. But I cannot give any guarantees.

Thank you for the great work on this library!

bruAristimunha:
Hi Arnau Dillen (@Ardillen66)!

I like your suggestions; I've been through something similar. Just a quick recap to see if I understood correctly:

1) Add some warning about sample sizes, try to check the minimum size, and establish some relationship for all 17 braindecode models and add this for the doc and the code.

2) improve error handling to try to warn the user that this kernel size issue  is usually related to the input size.

3) try to introduce padding mechanisms to the models (idk if this is a good idea),

Could you write some warning text? This would help the library and make it easier to implement the features. It is essential to have these multiple contributions to the braindecode to have a text that most people can understand.

Ardillen66:
Hi,

Sounds good!

I will try to write some documentation with the warning when I have the time and make a pull request for the updated docs.

bruAristimunha:
Thank you! let me know if you need any help.

bruAristimunha:
Thank you! let me know if you need any help.

bruAristimunha:
Hi @Ardillen66,

Do you need some assistence?

Ardillen66:
Hi,

Sorry for the long silence. I haven't had the opportunity to take care of this yet. If this can wait I can take care of the documentation  this month (I hope) and we can start looking into padding where appropriate after.

sliwy:
@bruAristimunha @PierreGtch as we will have `EEGModuleMixin` really soon I started to solve this issue ->
[...]

PR #514, opened 2023-09-08
EegModuleMixin extension

sliwy:
Adding `input_shape`, `get_output_shape`, `to_dense_prediction_model`

Deprecating `get_output_shape` and `to_dense_prediction_model`, will be removed in version 1.0.

When we have a proper way of removing logsoftmax we may replace functions with methods everywhere.

Solve #432 with raising error suggesting that input size may be too small.

### DIFF
--- a/braindecode/models/base.py
+++ b/braindecode/models/base.py
@@ -165,8 +165,7 @@ class EEGModuleMixin(metaclass=NumpyDocstringInheritanceInitMeta):
         """Input data shape."""
         return (1, self.n_chans, self.n_times)
 
-    @property
-    def output_shape(self) -> Tuple[int]:
+    def get_output_shape(self) -> Tuple[int]:
         """Returns shape of neural network output for batch size equal 1.
 
         Returns
--- a/braindecode/models/base.py
+++ b/braindecode/models/base.py
@@ -1,4 +1,5 @@
 # Authors: Pierre Guetschel
+#          Maciej Sliwowski
 #
 # License: BSD-3
 
--- a/braindecode/models/base.py
+++ b/braindecode/models/base.py
@@ -5,6 +5,8 @@
 import warnings
 from typing import Iterable, List, Optional, Dict
 
+import numpy as np
+import torch
 from docstring_inheritance import NumpyDocstringInheritanceInitMeta
 from torchinfo import ModelStatistics, summary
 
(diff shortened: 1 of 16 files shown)


sliwy:
Responding to @tgnassou comment:

> function get_output_shape only works for class which heritates from nn.Sequential. I would like to use it to replace the function _len_last_layer of SleepStagerChambon (https://github.com/braindecode/braindecode/blob/57280a6f870aa8cf1a1740bcbaa87f9253b39c24/braindecode/models/sleep_stager_chambon_2018.py#L120). 


Actually the problem is in trying to use `get_output_shape` in nn.Module in a moment when part of a model that is used in `forward` does not exist. I don't see an easy fix. Maybe creating a method that performs forward until last fc layer can be possible and have a kind of convention how this method should be named.

@tgnassou what kind of error do you receive? maybe it would be easier to think of solution

sliwy:
and yes, for partially initialized models it will work only for `nn.Sequential` because `nn.Sequential` `forward` can be called even when all the modules are not yet added to the model. 

In case of `nn.Module` it will be problematic to do that because parts of model do not exist and are used directly in the forward. We may think of try-exception clause in case some modules are not defined yet? but I don't think it;s worth implementing

PierreGtch:
> Responding to @tgnassou comment:
> 
> > function get_output_shape only works for class which heritates from nn.Sequential. I would like to use it to replace the function _len_last_layer of SleepStagerChambon (https://github.com/braindecode/braindecode/blob/57280a6f870aa8cf1a1740bcbaa87f9253b39c24/braindecode/models/sleep_stager_chambon_2018.py#L120
> > ).
> 
> Actually the problem is in trying to use `get_output_shape` in nn.Module in a moment when part of a model that is used in `forward` does not exist. I don't see an easy fix. Maybe creating a method that performs forward until last fc layer can be possible and have a kind of convention how this method should be named.
> 
> @tgnassou what kind of error do you receive? maybe it would be easier to think of solution

@sliwy, @tgnassou, once @brunaafl will be finished with #520, the next step will be to make methods `swap_final_layer` and `drop_final_layer` that we allow to easily replace the final layer with a placeholder identity function. This way we will be able to also compute the output shape for the `nn.Module`!

braindecode/models/base.py
```diff
@@ -158,6 +161,85 @@ def sfreq(self):
...
+                        "During model prediction RuntimeError was thrown showing that at some "
+                        f"layer `{str(exc).split('.')[-1]}` (see above in the stacktrace). This "
+                        "could be caused by providing too small `n_times`/`input_window_seconds`. "
+                        "Model may require longer chunks of signal in the input than "
+                        f"{self.input_shape}."
+                    )
+                    raise ValueError(msg) from exc
+                raise exc
+
+    def to_dense_prediction_model(self, axis: Tuple[int] = (2, 3)) -> None:
```
PierreGtch, line 200: Not familiar with this function, so I trust you with this @sliwy

sliwy:
@PierreGtch It's the convention I thought of, if we have a specific name of the last layer then it's possible

braindecode/models/util.py
```diff
@@ -3,11 +3,16 @@
...
+import torch
 from scipy.special import log_softmax
+from sklearn.utils import deprecated
 
 
+@deprecated(
+    "will be removed in version 1.0. Use EEGModuleMixin.to_dense_prediction_model method directly "
+    "on the model object."
+)
 def to_dense_prediction_model(model, axis=(2, 3)):
```
PierreGtch, line 16: Should we directly call `model.to_dense_prediction_model()` in this function ?

braindecode/models/util.py
```diff
@@ -54,6 +59,10 @@ def to_dense_prediction_model(model, axis=(2, 3)):
             module.stride = tuple(new_stride)
 
 
+@deprecated(
+    "will be removed in version 1.0. Use EEGModuleMixin.get_output_shape method directly on the "
+    "model object."
+)
 def get_output_shape(model, in_chans, input_window_samples):
```
PierreGtch, line 66: same comment with `model.get_output_shape()`

braindecode/models/util.py
sliwy: I would not do that because it works for any `nn.Sequential` model -> for example if we remove logsoftmax with the current flow it won't work. We have to wait for the logsoftmax removal within EEGModuleMixin. After that I'm all in

braindecode/models/util.py
sliwy: the same as for the previous comment

braindecode/models/util.py
PierreGtch: ok no problem 👍

PierreGtch (approved):

PR #515, opened 2023-09-08
Not using mne epochs for windowing

robintibor:
Not calling `mne.Epochs` construction anymore during windowing,using directly the computed crop indices on the mne raw.
For now, in same `WindowsDataset` class as before, potentially splitting the class into two later.


braindecode/datasets/base.py
```diff
@@ -194,7 +209,21 @@ def __getitem__(self, index):
...
+
+        i_window_in_trial, i_start, i_stop = crop_inds
+        if hasattr(self.windows, 'metadata'):  # mne epochs case:
+            X = self.windows.get_data(item=index)[0].astype('float32')
+        else:
+            X = self.windows._getitem((slice(None), slice(i_start, i_stop)), return_times=False)
+        X = X.astype('float32')
+        # ensure we don't give the user the option
+        # to accidentally modify the underlying array
+        X = X.copy()
```
sliwy, line 225: how much can this slow down `__getitem__`? This will be called all the time during training, curious whether it could be a potential speed up?

braindecode/datasets/base.py
```diff
@@ -194,7 +209,21 @@ def __getitem__(self, index):
...
+        # necessary to cast as list to get list of three tensors from batch,
+        # otherwise get single 2d-tensor...
+        crop_inds = self.crop_inds[index].tolist()
+
+        i_window_in_trial, i_start, i_stop = crop_inds
+        if hasattr(self.windows, 'metadata'):  # mne epochs case:
+            X = self.windows.get_data(item=index)[0].astype('float32')
+        else:
+            X = self.windows._getitem((slice(None), slice(i_start, i_stop)), return_times=False)
+        X = X.astype('float32')
```
sliwy, line 222: can this casting be done once for the whole raw file instead of casting at every `__getitem__`?

sliwy: Maybe we should be able to select precision? I imagine that someone may want to save memory and train model with fp16?

sliwy: and BTW, we are casting two times in the case of `hasattr(self.windows, 'metadata')==True`

braindecode/datasets/base.py
```diff
@@ -150,32 +150,47 @@ class WindowsDataset(BaseDataset):
...
         self.transform = transform
         self.last_target_only = last_target_only
         if targets_from not in ('metadata', 'channels'):
             raise ValueError('Wrong value for parameter `targets_from`.')
         self.targets_from = targets_from
 
-        self.crop_inds = self.windows.metadata.loc[
-            :, ['i_window_in_trial', 'i_start_in_trial',
-                'i_stop_in_trial']].to_numpy()
+        self.crop_inds = metadata.loc[
```
sliwy, line 189: how about keeping it in the form that is needed from the beginning -> I don't know how long will be `.to_list()` call but if it's done at every `__getitem__` this could have some influence

braindecode/datasets/base.py
```diff
@@ -150,32 +150,47 @@ class WindowsDataset(BaseDataset):
...
         Holds additional info about the windows.
     transform : callable | None
         On-the-fly transform applied to a window before it is returned.
     targets_from : str
-        Defines whether targets will be extracted from mne.Epochs metadata or mne.Epochs `misc`
+        Defines whether targets will be extracted from  metadata or from `misc`
         channels (time series targets). It can be `metadata` (default) or `channels`.
+    last_target_only : bool
+        If targets are obtained from misc channels whether all targets if the entire
+    (compute) window will be returned or only the last target in the window.
```
tomMoral, line 165:
```suggestion
        (compute) window will be returned or only the last target in the window.
```

braindecode/datasets/base.py
```diff
@@ -150,32 +150,47 @@ class WindowsDataset(BaseDataset):
...
+    metadata : pandas.DataFrame
+        Dataframe with crop indices, so `i_window_in_trial`, `i_start_in_trial`, `i_stop_in_trial`
+        as well as `targets`.
     """
+
     def __init__(self, windows, description=None, transform=None, targets_from='metadata',
-                 last_target_only=True):
+                 last_target_only=True, metadata=None, ):
```
tomMoral, line 172:
```suggestion
                 last_target_only=True, metadata=None):
```

braindecode/datasets/base.py
tomMoral: This is actually what is done here as this is not done in `__getitem__` no?

braindecode/datasets/base.py
```diff
@@ -194,7 +209,21 @@ def __getitem__(self, index):
...
         """
-        X = self.windows.get_data(item=index)[0].astype('float32')
+
+        # necessary to cast as list to get list of three tensors from batch,
+        # otherwise get single 2d-tensor...
+        crop_inds = self.crop_inds[index].tolist()
+
+        i_window_in_trial, i_start, i_stop = crop_inds
+        if hasattr(self.windows, 'metadata'):  # mne epochs case:
+            X = self.windows.get_data(item=index)[0].astype('float32')
```
tomMoral, line 219:
```suggestion
            X = self.windows.get_data(item=index)[0]
```

braindecode/datasets/base.py
tomMoral: I would say this is not a big issue as this is done asynchronously by the data loader.
A potential solution would be to use `X.flags.writeable = False` to make the array not writable without copy but this might break a lot of code so I would keep it that way.

braindecode/preprocessing/windowers.py
```diff
@@ -287,6 +292,111 @@ def _create_windows_from_events(
...
+    if any(np.diff(events[:, 0]) <= 0):
+        raise NotImplementedError('Trial overlap not implemented.')
+
+    description = events[:, -1]
+
+    metadata = pd.DataFrame({
+        'i_window_in_trial': i_window_in_trials,
+        'i_start_in_trial': starts,
+        'i_stop_in_trial': stops,
+        'target': description})
```
tomMoral, line 315:
```suggestion
        'target': description
    })
```

braindecode/preprocessing/windowers.py
```diff
@@ -287,6 +292,111 @@ def _create_windows_from_events(
...
+        # We could also just say we just assume window size=trial size
+        # in case not given, without this condition...
+        # but then would have to change functions overall
+        # to deal with varying window sizes hmmhmh
+        assert np.all(this_trial_sizes == window_size_samples), (
+            'All trial sizes should be the same if you do not supply a window '
+            'size.')
+
+    description = events[:, -1]
+
```
tomMoral, line 399: I woudl avoid code duplication here as the code is long. You could either make the first part of the function a sub function, or simply add a parameter to switch from epochs to raws and use `functools.partial(_create_windows_from_events, from_epochs=True))` as the `create_window_fn`.

braindecode/preprocessing/windowers.py
```diff
@@ -387,22 +497,17 @@ def _create_fixed_length_windows(
...
     )
-    windows_ds = WindowsDataset(mne_epochs, ds.description, targets_from=targets_from,
-                                last_target_only=last_target_only)
+    windows_ds = WindowsDataset(
+        ds.raw,
+        ds.description,
+        targets_from=targets_from,
+        last_target_only=last_target_only,
+        metadata=metadata,
+        )
```
tomMoral, line 510:
```suggestion
    )
```

tomMoral (commented):
Looks nice! A few comments.

braindecode/datasets/base.py
sliwy: I meant we're calling `.tolist()` in `__getitem__` in this line https://github.com/braindecode/braindecode/pull/515/files#diff-0e3c9da7298e0ba7355437c41fcfd2d4ef72d0dddb5eb06e857a25499c5d3e64R215
maybe it can be done in the `__init__` and then just accessed

tomMoral: It is a list of three items so probably okay I would say but maybe this can be benchmarked.

robintibor: yeah should be fine

braindecode/datasets/base.py
robintibor: yeah probably copying single window should be negligible overall...

braindecode/datasets/base.py
robintibor: Solved, no longer exists

braindecode/preprocessing/windowers.py
robintibor: Good idea, done

tomMoral:
yes, brackets on new lines makes more happy 😁

braindecode/datasets/base.py
robintibor: done

braindecode/datasets/base.py
robintibor: > can this casting be done once for the whole raw file instead of casting at every __getitem__?
I think not possible with mne unfortunately, at least that was my understanding. Otherwise of course would be better to be in float32 all the time 

> Maybe we should be able to select precision? I imagine that someone may want to save memory and train model with fp16?
Maybe in the future, although if purely for training speedup/GPU memory, could be done outside dataset I guess.

> and BTW, we are casting two times in the case of hasattr(self.windows, 'metadata')==True
Thanks, fixed

bruAristimunha:
Hey @robintibor!

From the CI it looks like it is finished, is there something missing in the PR?

robintibor:
So still need to reenable the outcommented test or remove it, and possibly rename dataset classes, but might do renaming of dataset classes in separate PR

robintibor:
Trying to get at it today

robintibor:
So this would be ready from my side now I think @bruAristimunha feel free to merge. See also new issues I created #543  and #544

bruAristimunha:
Amazing @robintibor 🚀!!! Thank you so much for the code and review, @robintibor, @tomMoral and @sliwy =)

PR #488, opened 2023-09-04
Normalise models BCI

PierreGtch:
cf #457

Issue #457, opened 2023-07-29
[(Brainde)Code Sprint] Normalise models implementations


PierreGtch:
**So far**, all the Braindecode models have been implemented independently, either inheriting from `torch.nn.Module` or from `torch.nn.Sequential`. Yet, they all receive inputs of the same shape (batch, channels, time), and they share many parameters and processing steps:
- *Shared by all models*:
    - number of classes (parameter)
    - number of channels (parameter)
    - fully-connected + softmax at the end (processing step)
- *Shared by some models*:
    - temporal length of the input (parameter)
    - sampling frequency (parameter)
    - dropout probability (parameter)
    - temporal and spatial filter at the beginning (processing step)

---

**The goal** of this meta-issue is to exploit these similarities to build a more unified framework. Specifically, this means:
- Create base [abstract](https://docs.python.org/3/library/abc.html) models from which all the Braindecode models should inherit;
- Define a common naming scheme for all models parameters;
- Define as many common parameters as possible in the base models;
- Implement as many processing steps as possible in the base models;
- For every inheriting model, explain the link between the new "unified" arguments and the arguments from the original implementations/articles.

---

**In a second step**, such a unified framework would allow to easily:
-  Get the embedding (last hidden layer) of any model;
- Indifferently pass a number of samples or a number of seconds for the time-related parameters;
- Improve the skorch EEG classifier and regressor:
    - they could automatically fill some parameters during `fit` according to the training data, such as the number of classes, the temporal length of the input, the sampling frequency (requires `mne.Epochs`) or the number of channels;
    - because they would require a model inheriting from the base Braindecode models, they could simply be initialized with a model name, allowing to programmatically load any Braindecode model.

robintibor:
Interesting, 2 points:
1) Looking a bit how [timm](https://github.com/huggingface/pytorch-image-models) does things may be helpful
2) `fully-connected + softmax at the end (processing step)` Not sure this is actually good though, like often one could absorb softmax into the loss during training as well... also fully connected depends on how you define it, quite  a few models do only have a convolutional layer that is extended to be "fully-connected" depending on the passed arguments.

PierreGtch:
Yes, I'm not a big fan of integrating the softmax in the network either but it is implemented this way right now and it is the expected default for skorch classifiers. Maybe it would be worth changing that.

Indeed, a fully connected layer might not be the best option for all models but it would be nice if at least all the classification layers could share the same name to be able to remove it automatically. 

[...]

PR #517, opened 2023-09-08
Automatically populate the signal-related parameters 

PierreGtch:
cf #488, #457


PierreGtch:
@sliwy I merged the `test_eeg_classifier.py` and `test_eeg_regressor.py` tests and updated them to use a “realistic" fake model with input-related parameters.

Could you help me update the test you had in them?

sliwy:
@PierreGtch what do you need from me? some specific test you think about?

PierreGtch:
@sliwy I broke the tests:
- `test_trialwise_predict_and_predict_proba`
- `test_cropped_predict_and_predict_proba`
- `test_cropped_predict_and_predict_proba_not_aggregate_predictions`
- and `test_predict_trials`
but I’m not familiar with the `cropped` parameter and the `CroppedLoss`. So not sure how to update them. Did you write them?

sliwy:
taking a look!

sliwy:
@PierreGtch let's move back here:
`test_eegneuralnet.py` is fixed in the PR: https://github.com/PierreGtch/braindecode/pull/3

When it comes to test_scoring.py and test_post_epoch_train_scoring it fails because we do not provide y in this line:
clf.fit(train_set, y=None, epochs=4)
and we use torchDataset but not BaseConcatDataset and not WindowsDataset so signal_kwargs['n_outputs'] is not set anywhere. We would need another elif for just plain torchDataset in _set_signal_args? what do you think?

braindecode/eegneuralnet.py
```diff
@@ -114,3 +135,78 @@ def _default_callbacks(self):
...
+        else:
+            raise ValueError(
+                f"X must be a numpy array or a Dataset, "
+                f"got {type(X)!r}."
+            )
+
+        # kick out missing kwargs:
+        module_kwargs = dict()
+        all_module_kwargs = inspect.signature(self.module.__init__).parameters.keys()
+        for k, v in signal_kwargs.items():
```
sliwy, line 182: why do we use only `signal_kwargs` here?

sliwy: if we start using kwargs for skorch we should use it for everything, otherwise we end up with reinitialization of model during fit to defualt parameters of the module

bruAristimunha:
Hey @PierreGtch and @sliwy,

Please ping me if you need any help to solve the design conflict. To be honest, I looked the discussion at the https://github.com/PierreGtch/braindecode/pull/3, and I still don't get the big picture. I am available to chat or meet if necessary.

PierreGtch:
As discussed in https://github.com/PierreGtch/braindecode/pull/3, we deprecate passing an initialized module to `_EEGNeuralNet` and skip setting signal-related arguments in this case.

PierreGtch:
@sliwy are you ok to merge this version?

PierreGtch:
> Hey @PierreGtch and @sliwy,
> 
> Please ping me if you need any help to solve the design conflict. To be honest, I looked the discussion at the [PierreGtch#3](https://github.com/PierreGtch/braindecode/pull/3), and I still don't get the big picture. I am available to chat or meet if necessary.

Sorry @bruAristimunha, missed your comment. The goal is to automatically populate all the signal-related parameters when we can and not break things when not possible.

braindecode/regressor.py
```diff
@@ -160,8 +161,16 @@ def predict_trials(self, X, return_targets=True):
             num_workers=self.get_iterator(X, training=False).loader.num_workers,
         )
 
-    def fit(self, X, y, **kwargs):
+    def fit(self, X, y=None, **kwargs):
```
sliwy, line 164: need a docstring for public API

test/unit_tests/test_eegneuralnet.py
```diff
@@ -0,0 +1,353 @@
...
+        train_split=None,
+        max_epochs=1,
+    )
+    net.fit(X, y=y)
+    assert net.module_.n_times == 10
+    assert net.module_.n_chans == 3
+    assert net.module_.n_outputs == (1 if isinstance(net, EEGRegressor) else 4)
+
+
+def test_set_signal_params_epochs(eegneuralnet_cls, preds):
```
sliwy, line 264: just a pass?

sliwy (changes requested):
I would restore a test with initialized `nn.Module` provided to the `EEGClassifer` - the one that was failing before (for example `test_post_epoch_train_scoring`).
Let's check that the warning is shown and that the model is not reinitialized with default params.

Apart from that and some details everything looks good to me! :)

test/unit_tests/test_eegneuralnet.py
PierreGtch: sorry it was not meant to be committed. This test will be for #529

braindecode/classifier.py
```diff
@@ -2,11 +2,13 @@
 #          Robin Schirrmeister <robintibor@gmail.com>
 #          Lukas Gemein <l.gemein@gmail.com>
 #          Bruno Aristimunha <b.aristimunha@gmail.com>
+#          Pierre Guetschel
```
robintibor, line 5: Email?

robintibor:
> > Hey @PierreGtch and @sliwy,
> > Please ping me if you need any help to solve the design conflict. To be honest, I looked the discussion at the [PierreGtch#3](https://github.com/PierreGtch/braindecode/pull/3), and I still don't get the big picture. I am available to chat or meet if necessary.
> 
> Sorry @bruAristimunha, missed your comment. The goal is to automatically populate all the signal-related parameters when we can and not break things when not possible.

I missed this, but would prefer we don't do that (deprecating passing initialized module). I think there is a tradeoff between automating more stuff (like setting parameter according to dataset) and being simpler/more transparent. So in my view should be fine to also pass initialized module, and in that case as is now, do not set parameters according to dataset... no reason to break this workflow I think...

PierreGtch:
@robintibor I can just turn the warning into an info and remove the deprecated part

robintibor:
sounds sensible

PierreGtch:
Then I'm ready to merge

bruAristimunha:
Hi @sliwy!

I enabled auto merge, if good for me you, just approve and will be merged.

robintibor:
Would be good to have some docs example for usage I think? To see how it all works?

PierreGtch:
Yes it's planned but I will wait for #529 and #528

sliwy (approved):

PR #529, opened 2023-09-10
Add support for mne.Epochs in _EEGNeuralNet

PierreGtch:
Passing mne.Epochs will allow to automatically populate more parameters of the model (cf #517 )


robintibor:
Be aware we are getting rid of mne epochs for efficiency reasons in https://github.com/braindecode/braindecode/pull/515

PierreGtch:
Yes when datasets are too large to load in memory. 
This PR is to simplify the case when all the data can be loaded in memory and you typically just pass a numpy array. Now you will be able to pass an mne.Epochs

robintibor:
Hm no idea was to move away from mne epochs entirely within the braindecode dataset classes, both for code simplicity as well as due to long times instantiating mne epochs in the create windows functions. But here it looks like anyways there are no braindecode dataset classes involved? might be good to have both some examples of this for docs as well as tests so it can be better understood :)

PierreGtch:
Exactly, here there is no braindecode dataset involved. It would just allow to do this:
```python
epo = mne.Epochs(...)
y = ...
net = EEGClassifier(...)
net.fit(epochs, y)
```
It's only to facilitate the usage for small datasets.

bruAristimunha:
I loved =)

bruAristimunha:
Elaborating more... I think it's an interesting counterproposal to what Robin is removing... We removed this internal support to make it more external, this way the neuro user can use it more easily. Do you agree @robintibor?

robintibor:
In general yes, I think right now for me hard to keep overview over everything, once we have things implemented with examples we should see from user perspective whether things are clear. Like also if it's clear when to use which option etc. Also from developer perspective how much overhead we have from multiple ways of specifying the data... Here this part seems minimal so that seems good :)

bruAristimunha:
Thank you @PierreGtch and @robintibor  =)

PR #530, opened 2023-09-10
Explain how to add a new model

### LINKED ISSUE #525: Adding a page on how to add a new model for braindecode.
We know how to do this, but it is unclear for new users, especially after the addition of the new EEGMixin. 

In my mind, this is a subpart of the contribution page (only in GitHub at the moment). Maybe we can do it like it was done on the installation page (divide into sub-parts).

PierreGtch:
Closes #525, cf #457

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,6 +116,71 @@ git checkout -b <branch-name>
 Now you can start writing the code to address the issue.
 Make sure to properly document and test your code.
 
+#### Add a model
+Braindecode models are all written in pytorch. They should all inherit from `braindecode.models.base.EEGModuleMixin`.
+You can use the following template to write a new model:
+```python
+class MyNewModel(EEGModuleMixin, nn.Module):
+    '''
+    My new model.
+
+    Model described in [1]_.
+
+    Parameters
+    ----------
+    my_parameter : int
+        DESCRIBE YOUR PARAMETERS HERE, AND WHY YOU SET THESE
+        DEFAULT VALUES
+
+    References
+    ----------
+    .. [1] First Author, Second Author (YEAR).
+       Title of the Article that Introduce this Architecture.
+       Journal or Conference name, DATE.
+       Online: URL/DOI OF THE PUBLICATION
+    '''
+
+    def __init__(
+            self,
+            n_outputs=None,
+            n_chans=None,
+            chs_info=None,
+            n_times=None,
+            input_window_seconds=None,
+            sfreq=None,
+            my_parameter=10,  # ADD YOUR PARAMETERS HERE
+    ):
+        super().__init__(
+            n_outputs=n_outputs,
+            n_chans=n_chans,
+            chs_info=chs_info,
+            n_times=n_times,
+            input_window_seconds=input_window_seconds,
+            sfreq=sfreq,
+        )
+        self.my_parameter = my_parameter
+
+        # DEFINE YOUR MODEL HERE
+        emb_dim = None  # CHANGE THIS
+        
+        self.final_layer = nn.Linear(emb_dim, self.n_outputs)
+
+    def forward(self, X):
+        # IMPLEMENT THE FORWARD PASS
+        return self.final_layer(X)
+```
+A few important notes:
+- The new model should accept all the parameters: `n_outputs`, `n_chans`, `chs_info`, `n_times`, `input_window_seconds`, and `sfreq`.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -116,6 +116,71 @@ git checkout -b <branch-name>
 Now you can start writing the code to address the issue.
 Make sure to properly document and test your code.
 
+#### Add a model
+Braindecode models are all written in pytorch. They should all inherit from `braindecode.models.base.EEGModuleMixin`.
+You can use the following template to write a new model:
```
bruAristimunha, line 121:
```suggestion
The models present in braindecode are PyTorch models. To standardize inputs, variables, and documentation, all models inherit from the EEGModuleMixin class. With this class, we take care to standardize variables and ensure that your model can run within our EEGClassifier or EEGRegressor.

Let's demonstrate how to implement a model for braindecode. Suppose we want to add a model called MyNewModel.
```

bruAristimunha:
It's ok for me @PierreGtc. But as this is core, related to what you have recently implemented, and involves an important part of the contributor part, I'm also asking Alex and Robin to review it as well.

bruAristimunha:
And @sliwy and @martinwimpff :)

CONTRIBUTING.md
```diff
@@ -59,70 +68,258 @@ pip install -e .[moabb,tests,docs]
...
 ```
 git checkout -b <branch-name>
 ```
 
 ### Write Code
+
 Now you can start writing the code to address the issue.
 Make sure to properly document and test your code.
 
+### Adding Your model into the Braindecode
```
robintibor, line 146:
```suggestion
### Adding Your model into Braindecode
```

Don't think there should be a "the" here

robintibor:
Thanks already for the explanations. In general, I think we need to try to make braindecode usage as simple and transparent and possible. Some concrete questions:
* is each and every of these mandatory parameters necessary?
* is it necessary that each of these is specified via ... =None?
* there are use cases where one wants a model that outputs `batch_size,n_times,n_targets`, e.g. `n_targets` maybe number of classes to predict or variables to regress. This explanation seems in conflict with that? What is `n_outputs` exactly meant to be?

PierreGtch:
> is each and every of these mandatory parameters necessary?
It won’t fail if the class doesn’t have all the parameters but I think it’s simpler for the users if all the models allow for all these parameters.
> is it necessary that each of these is specified via ... =None?
They can have a not-None default value.
> there are use cases where one wants a model that outputs batch_size,n_times,n_targets, e.g. n_targets maybe number of classes to predict or variables to regress. This explanation seems in conflict with that? What is n_outputs exactly meant to be?.
Hmmm I did not know we had this case with 3D outputs. 
Is this the cropped decoding? 
Which models can have a 3D output?

n_outputs replaced  n_classes. I changed it to also make sense if the model is used for regression.

robintibor:
In general one may want to have a temporal dimension in the output for segmentation-like tasks and we do that as well e.g., here https://braindecode.org/dev/auto_examples/advanced_training/plot_bcic_iv_4_ecog_cropped.html#sphx-glr-auto-examples-advanced-training-plot-bcic-iv-4-ecog-cropped-py.
To me this is not only in case one does cropped decoding way we do it, in general, you can have scenarios where you want a time dimension in your output, like if you want to temporally segment the signal.

PierreGtch:
Do you have ideas on how to explain this `n_outputs` parameters @robintibor?

agramfort (approved):

robintibor:
Thanks for the work and great explanations. I feel might be more sensible to put this at the end of the `CONTRIBUTING.md` file rather than somewhere in the middle... if you want to make it more visible you could add a pointer at the point it is currently written to say this part (how to add a model) comes at the end... Otherwise to me in current form it breaks a bit the structure as it is quite long and specific to one type of contribution. Also felt it might be nice to add a bit more info about the mandatory parameters meaning, maybe copy their docstring there and make clear that some of them may not be used/needed by the model and that that is fine. What do you think?

PierreGtch:
Yes I agree, it's much longer than the other sections. I put it at the end and added the definitions of the parameters.

bruAristimunha:
For me, it looks good, and we already have two approvals. As this document is in GitHub only and still needs to be incorporated into the documentation (in another PR). I will apply the merge for now and work on adjusting it further in another PR that I will do. I hope you don't mind @robintibor.

bruAristimunha:
Thank you @PierreGtch!

Issue #453, opened 2023-07-13
Implementation of SPD methods (example: TMSNET)


bruAristimunha:
Hey there!
I recently chatted with @rkobler through a colleague who's the author of TMSNET. We discussed the idea of bringing his network into braindecode (https://github.com/rkobler/TSMNet).
This network has a different approach compared to what we currently have in braindecode, but it seems like the community is leaning towards something like SPDNet and EEG.
Considering incorporating the first network layer as a covariance method won't disrupt the core principles of deep learning on raw data.
I wanted to keep the discussion in the issue open on whether someone should port this network or something similar to braindecode.

rkobler:
Hi! Thank you for creating this issue. I am happy to support the integration efforts.

bruAristimunha:
Related with this: https://github.com/GeometricBCI/Tensor-CSPNet-and-Graph-CSPNet

PR #534, opened 2023-09-12
[MRG] Add SPDNet

tgnassou:
Hi, I started implementing SPDNet in Braindecode, as discussed in the issue [#453](https://github.com/braindecode/braindecode/issues/453).  
It is just a draft, and the model is not working. I added an example if someone wants to try it. I still need the pooling and batch normalization layers to have the same SPDNet as [Huang et al.](https://arxiv.org/abs/1608.04233). I will also add the Riemannian batch norm of [Kobler et al.](https://arxiv.org/abs/2206.01323) and [Brooks et al.](https://arxiv.org/abs/1909.02414). 

I am new in this area of SPDNet. If anyone has any insights or suggestions, please let me know.


agramfort:
Ping @rkobler

bruAristimunha:
Heyyy @tgnassou!!

I added the requirements to your example compile. I added Pyriemann, because I planned to push a small tutorial on how to mix braindecode and pyriemann related to my research with @sylvchev. I hope you don't mind.

setup.py
```diff
@@ -40,6 +40,7 @@
         'docs': ['sphinx_gallery', 'sphinx_rtd_theme', 'pydata_sphinx_theme', 'numpydoc',
                  'memory_profiler', 'pillow', 'ipython', 'sphinx_design', 'docstring_inheritance',
                  'lightning'],
+        'riemann': ['pyriemann', 'geoopt'],
```
agramfort, line 43: if it's just for an example I would put this in docs key no? it becomes required for the doc public.

also how do you test now that braindecode can work without geoopt?

bruAristimunha: hmm, true @agramfort!

bruAristimunha: solved @agramfort!

jpaillard:
Hi !
Interesting implementation, thx for sharing.
I was wondering if there is a reason why you adopted Riemannian gradients with geoopt instead of a parameterized optimization approach such as [GeoTorch](https://github.com/lezcano/geotorch). 
According to [this paper (theorem 4.3)](https://proceedings.neurips.cc/paper_files/paper/2019/file/1b33d16fc562464579b7199ca3114982-Paper.pdf)  both approaches theoretically lead to the same outcome but maybe there is a numerical difference ? 
If it makes sense I can provide an implementation using GeoTorch for comparison :)

rkobler:
Hi @tgnassou!

Great that you started to implement SPD type models in braindecode. I am happy to support and help in the implementation process!

@jpaillard it would be definitely interesting to test both approaches. So far, I only have experience with geoopt.

bruAristimunha:
The review is yours @agramfort :)

braindecode/models/functions.py
```diff
@@ -45,3 +45,104 @@ def transpose_time_to_spat(x):
...
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        s, U, s_modified = ctx.saved_tensors
+        def derivative(s): return s.reciprocal()
+        return modeig_backward(grad_output, s, U, s_modified, derivative)
+
+
+class regm(torch.autograd.Function):
+    """Regularized matrix logarithm of a symmetric matrix.
```
agramfort, line 116: it's not regularized in the classic sense where you add a constant on the diag. It puts all values less then threshold to 0 so it outputs rank deficient matrices unless I miss something.

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,238 @@
...
+    def forward(self, X):
+        return self.W.transpose(-1, -2) @ X @ self.W
+
+
+class ReEig(nn.Module):
+    """ReEig layer from [1]_.
+
+    This class add non-linearity to the network by
+    applying a rectified linear unit to the eigenvalues
+    of a symmetric matrix.
```
agramfort, line 91: Say that by doing this it makes the matrix non-negative and positive if threshold > 0.

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def forward(self, X):
+        return regm.apply(X, self.threshold_)
+
+
+class LogEig(nn.Module):
+    """LogEig layer from [1]_.
+
+    This class perform Riemannian projection into a flat space
+    by applying the logarithm to the eigenvalues of a symmetric matrix.
```
agramfort, line 121: say also that is flattens the output as the logm returns a matrix not a vector in maths/scipy.

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,238 @@
...
+
+class SPDNet(EEGModuleMixin, nn.Module):
+    """SPDNet from [1]_.
+
+    This class is a SPDNet implementation from [1]_.
+
+    Parameters
+    ----------
+    input_type : str
+        Input type
```
agramfort, line 171:
```suggestion
```

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,238 @@
...
+           AAAI
+    """
+    def __init__(
+        self,
+        input_type,
+        n_chans,
+        subspacedim,
+        threshold=1e-4,
+        n_outputs=1,
+        chs_info=None,
```
agramfort, line 200: check that all parameters are documented in the docstring

test/unit_tests/models/test_models.py
```diff
@@ -657,6 +658,59 @@ def hook(model, input, output):
...
+        input_type="cov",
+        n_chans=n_channels,
+        n_outputs=n_classes,
+        subspacedim=subspacedim,
+        n_times=int(sfreq * input_size_s)
+    )
+    model.eval()
+
+    y_pred2 = model(cov)  # 3D inputs
+    assert y_pred2.shape == (n_examples, n_classes)
```
agramfort, line 711: here you test that it runs. I would argue to test that the loss goes down as otherwise you don't check the correctness of the code.

braindecode/models/functions.py
```diff
@@ -45,3 +45,104 @@ def transpose_time_to_spat(x):
...
+        output, s, U, s_modified = modeig_forward(X, function_applied)
+        ctx.save_for_backward(s, U, s_modified)
+        ctx.threshold_ = threshold
+        return output
+
+    @staticmethod
+    def backward(ctx, grad_output):
+        s, U, s_modified = ctx.saved_tensors
+        def derivative(s): return s > ctx.threshold_
+        return modeig_backward(grad_output, s, U, s_modified, derivative), None
```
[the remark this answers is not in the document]
tgnassou: For SPDNet, it makes sense, but for TSMNet, for example, you will need a logm independently of regm. And it's clearer to have LogEig and ReEig layers to fit the paper.

braindecode/models/spdnet.py
tgnassou: Only some of the parameters are documented, but it is the same in several models in braindecode. I didn't know if I needed to add them or not. If this is a mistake in other models, a fix doc PR is needed. WDYT @bruAristimunha ?

agramfort: does it mean that here chs_info cannot be set (/ does not make sense) for SPDNet?

bruAristimunha: Pinging @PierreGtch

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Parameters
+    ----------
+    input_type : str
+        If "raw", the input is a batch of raw EEG signals and
+        a covariance matrix is computed
+        If "cov", the input is a batch of covariance matrices and
+        the covariance matrix is used as input
+    n_chans : int
+        Number of channels
```
PierreGtch, line 166:
```suggestion
```
no need, it will automatically be added by `EEGModuleMixin`

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+    subspacedim : int
+        Subspace dimension
+    threshold : float
+        Threshold for the rectified linear unit
+    n_outputs : int
+        Output shape
```
PierreGtch, line 172:
```suggestion
```
same

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+           A Riemannian Network for SPD Matrix Learning
+           AAAI
+    """
+    def __init__(
+        self,
+        input_type,
+        n_chans,
+        subspacedim,
+        threshold=1e-4,
+        n_outputs=1,
```
PierreGtch, line 188: Why do you use 1 as default value for `n_outputs` and not `None`?

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+        self.bimap = BiMap(n_chans, subspacedim)
+        self.reeig = ReEig(threshold)
+        self.logeig = LogEig(subspacedim, tril=tril)
+        self.len_last_layer = subspacedim * (subspacedim + 1) // 2 if tril else subspacedim**2
+        self.classifier = torch.nn.Linear(self.len_last_layer, n_outputs)
+
+        if add_log_softmax:
+            self.logsoftmax = nn.LogSoftmax(dim=1)
+        else:
+            self.logsoftmax = nn.Identity()
```
PierreGtch, line 218:
```suggestion
```
The goal in the long term is to get rid of the log softmax layer in every model (cf #511). The parameter `add_log_softmax` was introduced to throw a depreciation for the models that had one. But new models should not have a log softmax layer at all.

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+        tril=True,
+    ):
+        super().__init__(
+            n_outputs=n_outputs,
+            n_chans=n_chans,
+            chs_info=chs_info,
+            n_times=n_times,
+            input_window_seconds=input_window_seconds,
+            sfreq=sfreq,
+            add_log_softmax=add_log_softmax,
```
PierreGtch, line 203:
```suggestion
```

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+        input_type,
+        n_chans,
+        subspacedim,
+        threshold=1e-4,
+        n_outputs=1,
+        chs_info=None,
+        n_times=None,
+        input_window_seconds=None,
+        sfreq=None,
+        add_log_softmax=False,
```
PierreGtch, line 193:
```suggestion
```

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+            self.logsoftmax = nn.Identity()
+
+    def forward(self, X):
+        X = self.cov(X)
+        X = self.bimap(X)
+        X = self.reeig(X)
+        X = self.logeig(X)
+        X = self.classifier(X)
+        output = self.logsoftmax(X)
+        return output
```
PierreGtch, line 227:
```suggestion
        return X
```

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+            n_times=n_times,
+            input_window_seconds=input_window_seconds,
+            sfreq=sfreq,
+            add_log_softmax=add_log_softmax,
+        )
+        if input_type == "raw":
+            self.cov = CovLayer()
+        elif input_type == "cov":
+            self.cov = nn.Identity()
+        self.bimap = BiMap(n_chans, subspacedim)
```
PierreGtch, line 209:
```suggestion
        self.bimap = BiMap(self.n_chans, subspacedim)
```
So that `n_chans` can be inferred in case only `chs_info` is passed

braindecode/models/spdnet.py
```diff
@@ -0,0 +1,227 @@
...
+        self.reeig = ReEig(threshold)
+        self.logeig = LogEig(subspacedim, tril=tril)
+        self.len_last_layer = subspacedim * (subspacedim + 1) // 2 if tril else subspacedim**2
+        self.classifier = torch.nn.Linear(self.len_last_layer, n_outputs)
```
PierreGtch, line 213:
```suggestion
        self.classifier = torch.nn.Linear(self.len_last_layer, self.n_outputs)
```

braindecode/models/spdnet.py
tgnassou: Thanks, I wasn't sure if I needed to put it or not.

bruAristimunha:
Hey @tgnassou and @agramfort, any update about this PR?

bruAristimunha (commented):
solving some comments

bruAristimunha:
@tgnassou, can I assume this PR? I think will be super convenient with my paper that I'm finishing this month about SPDNet.

tgnassou:
Yeah sure!

bruAristimunha:
thank you @tgnassou

bruAristimunha:
Things exploded when I updated the branch. I will re-start in a fresh branch/PR @tgnassou

PR #535, opened 2023-09-12
changed default value of final_conv_length to 'auto' in Deep4Net

sliwy:
After last changes, default value of `final_conv_length` in `Deep4Net` is `None` and code fails. Before it was mandatory to provide this paramter. I'm setting default to `"auto" -> it should not break earlier code as before it had to be specified.


bruAristimunha (approved):
Lgtm

robintibor:
good catch! and what about shallow? should be same Problem there no?

sliwy:
for shallow `filter_time_length` is set to 25 by default, it was the same before recent changes and it still works as before :)

bruAristimunha:
Thank you @sliwy =)

Issue #537, opened 2023-09-13
USleep TypeError when printing modules


brunaafl:
`
m = USleep()
for name, module in m.named_modules():
    print(name)
    print(module)
`
is returning the following error:
`TypeError: kernel_size has an unexpected type: <class 'numpy.int64'>`

More specifically:
`
Traceback (most recent call last):
  File "/home/brunalopes/PycharmProjects/braindecode/eeg/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3526, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-c1a9f6a7a3e7>", line 5, in <module>
    print(module)
  File "/home/brunalopes/PycharmProjects/braindecode/braindecode/models/base.py", line 291, in __str__
    return str(self.get_torchinfo_statistics())
  File "/home/brunalopes/PycharmProjects/braindecode/braindecode/models/base.py", line 282, in get_torchinfo_statistics
    return summary(
  File "/home/brunalopes/PycharmProjects/braindecode/eeg/lib/python3.10/site-packages/torchinfo/torchinfo.py", line 223, in summary
    summary_list = forward_pass(
  File "/home/brunalopes/PycharmProjects/braindecode/eeg/lib/python3.10/site-packages/torchinfo/torchinfo.py", line 275, in forward_pass
    summary_list, _, hooks = apply_hooks(model_name, model, x, batch_dim)
  File "/home/brunalopes/PycharmProjects/braindecode/eeg/lib/python3.10/site-packages/torchinfo/torchinfo.py", line 634, in apply_hooks
    global_layer_info[module_id] = LayerInfo(
  File "/home/brunalopes/PycharmProjects/braindecode/eeg/lib/python3.10/site-packages/torchinfo/layer_info.py", line 59, in __init__
    self.kernel_size = self.get_kernel_size(module)
  File "/home/brunalopes/PycharmProjects/braindecode/eeg/lib/python3.10/site-packages/torchinfo/layer_info.py", line 168, in get_kernel_size
    raise TypeError(f"kernel_size has an unexpected type: {type(k)}")
TypeError: kernel_size has an unexpected type: <class 'numpy.int64'>
`

sliwy:
Looks like something for me, I'm gonna see why it's not working

PR #538, opened 2023-09-13
fix kernel_size np.int64 dtype in torchinfo

sliwy:
It solves #537 where torchinfo was throwing wrong `kernel_size` type.

### DIFF
--- a/braindecode/models/usleep.py
+++ b/braindecode/models/usleep.py
@@ -201,7 +201,7 @@ class USleep(EEGModuleMixin, nn.Module):
         del in_chans, n_classes, input_size_s
 
         max_pool_size = 2  # Hardcoded to avoid dimensional errors
-        time_conv_size = np.round(time_conv_size_s * self.sfreq).astype(int)
+        time_conv_size = int(np.round(time_conv_size_s * self.sfreq))
         if time_conv_size % 2 == 0:
             if ensure_odd_conv_size:
                 time_conv_size += 1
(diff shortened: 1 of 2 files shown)


bruAristimunha:
Hey @sliwy,

I searched and look's like the Chambon is impact by this issue https://github.com/search?q=repo%3Abraindecode%2Fbraindecode%20.astype(int)&type=code

Can you make a small test to test all the models? You can use the models_dict proposed by @PierreGtch.

PierreGtch:
On the same topic, it would be nice if we had, for every model, a set parameters that can initialise it. This would be useful for testing if the implementations are conform and all the methods of EEGModuleMixin working on every model.

sliwy:
Hey @bruAristimunha, I checked and for `SleepStagerChambon2018` it works :) no clue why :joy: 

For the test, definitely, it could be something like this in the beginning but  we need to know parameters to use or set good default values:
```
from braindecode.models.util import models_dict
@pytest.mark.parametrize("model_cls", models_dict.values())
def test_model_torchinfo(model_cls):
    print(model_cls(n_times=5000, n_chans=1, n_outputs=2))
```

bruAristimunha:
Let's use two shapes:

For the BCI models, we will use BCI 2014 (1, 22, 1001),
For sleep, we will use Sleep Dataset (1, 2, 3000)./

bruAristimunha:
As we have already opened other issues to address the requested extra, I will consider this issue closed and we will continue with another PR. Thank you very much @sliwy!!

PR #551, opened 2023-10-16
[augmentation] fix type hints for Output of Transform.forward

tgy:
- [x] fix `Output` type annotation, now reflecting that it can return `y` as a
      tuple of `torch.Tensor`
- [x] define new `Operation` type hint, to annotate the `Transform.operation`
      attribute, to better check operation return types
- [x] define `Batch`, `Output` and `Operation` as type aliases

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,7 +36,7 @@ jobs:
       run: python --version
     # Update pip
     - name: Update pip
-      run: pip install --upgrade pip
+      run: python -m pip install --upgrade pip
     - name: Install Braindecode from Current Checkout
       run: pip install -e .[moabb,tests]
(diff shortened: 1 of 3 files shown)


tgy:
seems that `TypeAlias` was not supported before Python 3.10, removed it

fixed the trailing whitespaces as well (flake8 should now pass, although we should switch to ruff - yes, saw the other PR)

tgy:
@bruAristimunha any idea why the Windows test are failing? seems it's failing why before running tests, during a pip install that fails

bruAristimunha:
Okay. Thank you @tgy for fixing the CI for the window. All good for me =)

bruAristimunha:
Thank you for fix type @tgy

PR #553, opened 2023-10-23
Updating to MOABB 1.0.0

sylvchev:
MOABB v1.0.0 is out and changes datasets class names, and session/run naming scheme. This PR update examples and code to this new version.

Also, it replaces the requirements/setup.py by the pyproject.toml file, that is compatible with pip and allows simpler maintenance. It is [supported by setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) as required by [PEP 621](https://peps.python.org/pep-0621/).


bruAristimunha:
Thank you @sylvchev

PR #558, opened 2023-11-07
Replacing the real dataset

bruAristimunha:


bruAristimunha:
Not super good, same average speed @robintibor, but we can work more in the future. My opinion, merge this PR and leave the issue open.

robintibor:
Hm, would argue for removing this from what's new? like some internal test stuff not really useful to know for any user?

bruAristimunha:
indeed @robintibor,  I ended up realizing that there was a small error in the whats new file. I fixed the file.

robintibor:
great, merged

Issue #561, opened 2023-12-07
Annotations are now always set in MOABB


dcwil:
A relatively recent update to moabb means that annotations will always be set, so for at least BNCI2014001 the cut trials come out wrong. I'm not sure what the intended approach at moabb is, and whether we will need to modify our annotating or not, but with their current workflow, this if statement will never be triggered. 

https://github.com/braindecode/braindecode/blob/22da938bec9445aaa7a7a58a6abcfb14bb2d1d71/braindecode/datasets/moabb.py#L39

PR #563, opened 2023-12-08
Fix annotations

dcwil:
See #561 

Now that annotations are always set in MOABB, this if statement was never triggered. This meant the datasets were loading without the interval offset applied to their annotations.


bruAristimunha:
thank you @dcwil

Issue #552, opened 2023-10-18
[docs] Model overview


skjerns:
It would be really great to have an overview over the models that are implemented in braindecode. 

Currently it's a bit confusing to hunt down the details in the publications, maybe a table with properties, and an accompanying visualization would help? Probably some parts of this could be automated and augmented with details such as what input window or sampling frequency is recommended.

bruAristimunha:
Hey @skjerns!

You are right! I will work on this!

PR #571, opened 2024-02-14
[MRG] Creating tests to ensure that all the models work with default parameters, mapping the mandatory parameters for the models

### LINKED ISSUE #542: Create test for a default parameter for the models
@sliwy not sure all the models can be initialized with their default parameters.

Plus I think we should test if the models accept inputs with the desired shape and return outputs with the correct shape:
```python
from braindecode.models.util import models_dict
@pytest.mark.parametrize("model_cls", models_dict.values())
def test_model_torchinfo(model_cls):
    model = model_cls(n_times=5000, n_chans=1, n_outputs=2)
    assert model(torch.ones(3, 1, 5000)).shape == (3, 2)
```

_Originally posted by @PierreGtch in https://github.com/braindecode/braindecode/issues/538#issuecomment-1725468288_

bruAristimunha:
After our last sprint, all models can be built without specifying any parameters.

Everything should be inferred at runtime. Everything is working more or less, but some models need adjustments.

closes #542


bruAristimunha:
At the moment, it is not working with:
- EEGResNet;
- TCN;
- HybridNet;
- SleepStagerEldele

PierreGtch:
@bruAristimunha during the sprint, we discussed that we did not want to change (for now) the behavior when using the default parameters of the classes to avoid breaking pre-existing codes. This lead to a bit of heterogeneity in the class init parameters. Could this be related?

PierreGtch:
@bruAristimunha see my PR here https://github.com/bruAristimunha/braindecode/pull/5
The test I implemented there should be faster than by using `EEGClassifier` and more exhaustive. But it does not test the backward pass

bruAristimunha:
Thaaaaaaank you so much @PierreGtch, I need to pay for a beer or a French sweet for that.

PierreGtch:
Hahaha yes! See you soon in Paris, I hope

But before merging this, I think we should discuss more about the change of default parameters of the models. Especially `input_window_seconds` becoming `None` by default in a few models, it might break some stuff

test/unit_tests/models/test_integration.py
```diff
@@ -0,0 +1,261 @@
...
+    ("USleep", ["n_chans", "n_outputs", "n_times", "sfreq"], dict(sfreq=128)),
+]
+
+
+def test_completeness__models_test_cases():
+    models_tested = set(x[0] for x in models_test_cases)
+    all_models = set(models_dict.keys())
+    assert (
+        all_models == models_tested
+    ), f"Models missing from models_test_cases: {all_models - models_tested}"
```
bruAristimunha, line 153: I liked this test @PierreGtch!

PierreGtch:
@bruAristimunha in https://github.com/bruAristimunha/braindecode/pull/6 I modified your test to pass epochs with the correct shape for each model to `EEGClassifier` instead of datasets. It works with every model except those that return 2d outputs (i.e., `HybridNet` and `TCN`).  How should we handle those?

bruAristimunha:
I think we can ignore it for these two models, it would be cool to fix it for the TCN, but I'm not sure how. 

I have the feeling that the parameter list should be inside the library, not just in tests. It can be very useful to build the model table in the documentation, as requested in #552

@agramfort, Do you want to give any comments, perhaps suggest some other type of test that you think is important? Any input from you is welcome :)

PierreGtch:
Indeed, good idea! 
In this case, we should use the signal parameters used in the original publication of each model, not just some parameters that work. I can work on this but only after march 7.

PR #574, opened 2024-02-21
Allow inferring signal parameters from torch tensors

PierreGtch:
Currently, we can infer the shape of the input signal when numpy arrays are passed to `EEGClassifier` but not when torch tensors are.  This small change allows for both


bruAristimunha (approved):

bruAristimunha:
Nice doc adjust @PierreGtch, thank you for your PR.

PR #590, opened 2024-04-12
ENH Add type hints to datasets

### LINKED ISSUE #588: Add type hints
It would be nice to have more type hints in the library.

This would facilitate the use of packages like [jsonargparse](https://jsonargparse.readthedocs.io/en/v4.27.7/index.html) which requires type hints for the mandatory arguments at least.

But the easy type hints for generic structures (`list`, `dict`...) are only supported in python 3.9 and the unions with `|` only arrive in python 3.10, so maybe let’s wait until we roll out 3.8 at least.

PierreGtch:
Add type hints 

#588

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -10,6 +10,8 @@ Dataset classes.
 #
 # License: BSD (3-clause)
 
+from __future__ import annotations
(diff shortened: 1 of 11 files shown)


bruAristimunha:
Thank you @PierreGtch

PR #600, opened 2024-05-07
Add type hints to preprocessing functions

PierreGtch:


agramfort (approved):
diff looks good but I feel we here need to add mypy support to check that it's actually correct.

+1 for merge and add mypy CI job in another PR

bruAristimunha:
Thank you @PierreGtch

PR #607, opened 2024-05-09
Preprocessing `EEGWindowsDataset`

### LINKED ISSUE #605: Preprocessing on `EEGWindowsDataset` vs `WindowsDataset`
Hey,
When applying preprocessing steps using `braindecode.preprocessing.preprocess` on a collection of `EEGWindowsDataset`s, the steps are applied on the `mne.io.Raw` objects, whereas if using `WindowsDataset`s they are applied on the `mne.Epochs`.

`create_windows_from_events` can return both types of objects depending on its parameters, so this can lead to unexpected behaviours for the user.

1. I think we should not deprecate `use_mne_epochs` https://github.com/braindecode/braindecode/blob/92a864ae97aeed9e8b2b13fa268fef2d235b571f/braindecode/preprocessing/windowers.py#L151-L156 and actually leave the option to the user to intentionally set `use_mne_epochs=True`
2. we should warn when preprocessing a `EEGWindowsDataset` or change the current behaviour

PierreGtch:
This PR:
1. Exposes the `use_mne_epochs` parameter of `create_windows_from_events`
2. warns when preprocessing an `EEGWindowsDataset` because the preprocessing steps will be applied on the raw data, not on epochs

Adding a `use_mne_epochs`  parameter to `create_fixed_length_windows` and `create_windows_from_target_channels` would be desirable but more complicated. 

Closes #605


bruAristimunha:
For me it's almost good @PierreGtch, just a few more tests to increase coverage and maybe a test for the function's behavior.

PierreGtch:
@bruAristimunha Done :)

bruAristimunha:
Lgtm @PierreGtch, thank you for your contribution.

PR #573, opened 2024-02-14
[MRG] Adding BIOT to braindecode

bruAristimunha:

PR #608, opened 2024-05-09
Add SegmentationReconstruction augmentation 

gustavohenriquesr:
Hi, I created a new data augmentation class named SegmentationReconstruction to braindecode. This process was originally proposed in [this](https://ieeexplore.ieee.org/abstract/document/7109822) paper, which I implemented and used in my own work.

This technique works on the time domain, and the ideia is to segment each trial into several segments and then generate new artificial trials, as a concatenation of segments coming from different and randomly selected training trials from the same class.


test/unit_tests/augmentation/test_transforms.py
```diff
@@ -631,6 +633,21 @@ def test_mixup_transform(rng_seed, random_batch, alpha, beta_per_sample):
 ]
 
 
+@pytest.mark.parametrize("n_segments", [5, 10, None])
```
bruAristimunha, line 636: Please test more, and add more variability for the tests (more functions and parameters).


Include edge cases, n segment 0, 1, ..., number of dots, you need to test with prob 0 and 1, make work with composition with another augmentation and/or pre processing. 

More test also using augmented mode as present in the examples.

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,92 @@ def mixup(X, y, lam, idx_perm):
         y_b[idx] = y[idx_perm[idx]]
 
     return X_mix, (y_a, y_b, lam)
+
+
+def segmentation_reconstruction(X, y, n_segments=None, random_state=None):
```
bruAristimunha, line 975: What happens here if we try the limit cases? No X, no y, 0 segments, and if we don't inform the label? Make this function work in all scenarios. Try to think of more cases too.

Try to capture these scenarios and try to inform the user what is wrong, e.g. number of segments greater than the number of points.

Similar to what is presented at the EEG conformer, but more robust.

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,92 @@ def mixup(X, y, lam, idx_perm):
...
+    -------
+    torch.Tensor
+        Transformed inputs.
+    torch.Tensor
+        Transformed labels.
+    References
+    ----------
+    .. [1] Lotte, F. (2015). Signal processing approaches to minimize or
+    suppress calibration time in oscillatory activity-based brain–computer
+    interfaces. Proceedings of the IEEE, 103(6), 871-890.
```
bruAristimunha, line 1001: I think are you missing some reference here, right?

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,92 @@ def mixup(X, y, lam, idx_perm):
...
+    # Find the factors of the number of samples to segment
+    if n_segments is None:
+        n_segments = int(X.shape[2])
+        n_segments_list = []
+        for i in range(1, int(n_segments**0.5) + 1):
+            if n_segments % i == 0:
+                n_segments_list.append(i)
+    assert isinstance(n_segments, (int, float))
+    # parse the float to int
+    n_segments = int(n_segments)
```
bruAristimunha, line 1016: Please invert the order, and change the assert

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,92 @@ def mixup(X, y, lam, idx_perm):
...
+    for X_class, class_index in data_classes:
+        # Determine class-specific dimensions
+        n_samples, n_channels, window_size = X_class.shape
+        # Segment Size
+        segment_size = window_size // n_segments
+        # Initialize an empty tensor for augmented data
+        X_aug = torch.zeros_like(X_class)
+        # Use PyTorch's random generator for consistency
+        for idx_sample in range(n_samples):
+            for idx_segment in range(n_segments):
```
bruAristimunha, line 1034: Question/challenge: can you replace those two for loop to one for loop with a numpy operation or one numpy operation only?

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,92 @@ def mixup(X, y, lam, idx_perm):
...
+    # Concatenate the augmented data and labels
+    aug_data = torch.cat(aug_data, dim=0)
+    aug_label = torch.cat(aug_label, dim=0)
+
+    idx_shuffle = rng.permutation(len(aug_data))
+
+    aug_data = aug_data[idx_shuffle]
+    aug_label = aug_label[idx_shuffle]
+
+    return aug_data, aug_label
```
bruAristimunha, line 1061: I think this return will not provide compatibility with augmentation module

bruAristimunha (changes requested):

bruAristimunha:
Hey @gustavohenriquesr,

Thank you for your first contribution!

I am super happy about this ;)

I made some initial comments and I will try to ask for some extra revision, as I am already familiar with the code.

bruAristimunha:
Extra comment, please edit the source doc to include in the documentation page.

braindecode/augmentation/functional.py
gustavohenriquesr: I handled the limit cases suitable for this function.

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+        for i in range(1, int(n_segments**0.5) + 1):
+            if n_segments % i == 0:
+                n_segments_list.append(i)
+        n_segments = n_segments_list[-1]
+
+    elif not (isinstance(n_segments, (int, float)) and 1 <= n_segments <= X.shape[2]):
+        raise ValueError(
+            f"Number of segments must be a positive integer less than "
+            f"(or equal) the window size. Got {n_segments}"
+        )
```
bruAristimunha, line 1043: I change here para raise. It is not nice to assert inside function; better only on tests.

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+
+    # Initialize lists to store augmented data and corresponding labels
+    aug_data = []
+    aug_label = []
+
+    # Getting the random_state
+    rng = check_random_state(random_state)
+
+    if False:
```
bruAristimunha, line 1052: I put it as false because it is not working with the augmentation module. Here, we have two options:

1) solve the issue in the base of the augmentation module (better)
2) add a variable to control this, and set this increase as false by default. (raising an error if you use it in the augmentation module).

Any option you implement is acceptable to me, @gustavohenriquesr.

bruAristimunha, line 1052:
```suggestion
    if y is not None:
```

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+        aug_data.append(X_aug)
+
+    # Concatenate the augmented data and labels
+    aug_data = torch.cat(aug_data, dim=0)
+
+    idx_shuffle = rng.permutation(len(aug_data))
+
+    aug_data = aug_data[idx_shuffle]
+
+    if False:
```
bruAristimunha, line 1077:
```suggestion
    if y is not None:
```

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+        if X.shape[0] != y.shape[0]:
+            raise ValueError("Number of samples in X and y must be the same.")
+
+    if n_segments is None:
+        n_segments = int(X.shape[2])
+        n_segments_list = []
+        for i in range(1, int(n_segments**0.5) + 1):
+            if n_segments % i == 0:
+                n_segments_list.append(i)
+        n_segments = n_segments_list[-1]
```
cedricrommel, line 1037: I find the construction of n_segments curious here when the kwarg is None. Could you explain or provide a reference which explains this please, just for my own culture ?

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+    else:
+        X_aug, n_trials = _segmentation_reconstruction(X, n_segments, rng)
+        aug_data.append(X_aug)
+
+    # Concatenate the augmented data and labels
+    aug_data = torch.cat(aug_data, dim=0)
+
+    idx_shuffle = rng.permutation(len(aug_data))
+
+    aug_data = aug_data[idx_shuffle]
```
cedricrommel, line 1075: Same here for the random opertation: could it be moved to the class level please?

braindecode/augmentation/transforms.py
```diff
@@ -1070,3 +1072,61 @@ def get_augmentation_params(self, *batch):
...
+
+    Parameters
+    ----------
+    probability : float
+        Float setting the probability of applying the operation.
+    random_state: int | numpy.random.Generator, optional
+        Seed to be used to instantiate numpy random number generator instance.
+        Used to decide whether to transform given the probability
+        argument. Defaults to None.
+    n_segments : int
```
cedricrommel, line 1089: "optional" missing

braindecode/augmentation/transforms.py
```diff
@@ -1070,3 +1072,61 @@ def get_augmentation_params(self, *batch):
             "lam": lam,
             "idx_perm": idx_perm,
         }
+
+
+class SegmentationReconstruction(Transform):
+    """Applies a segmentation-reconstruction transform to the input data, as
+       proposed in _[1].
```
cedricrommel, line 1079: I guess we could say slightly more about what the transform does: creates synthetic EEG by randomly mixing segments of recordings of the same class from the batch.

braindecode/augmentation/transforms.py
```diff
@@ -1070,3 +1072,61 @@ def get_augmentation_params(self, *batch):
...
+
+    Parameters
+    ----------
+    probability : float
+        Float setting the probability of applying the operation.
+    random_state: int | numpy.random.Generator, optional
+        Seed to be used to instantiate numpy random number generator instance.
+        Used to decide whether to transform given the probability
+        argument. Defaults to None.
```
cedricrommel, line 1088: Also used to sample the segments mixing.

test/unit_tests/augmentation/test_transforms.py
```diff
@@ -631,6 +636,60 @@ def test_mixup_transform(rng_seed, random_batch, alpha, beta_per_sample):
...
+        iterator_train=AugmentedDataLoader,
+        iterator_train__transforms=transform,
+        batch_size=5,
+        max_epochs=1,
+        classes=[0, 1],
+        train_split=predefined_split(dataset),
+        verbose=0,
+    )
+
+    _ = clf.fit(X=dataset)
```
cedricrommel, line 663: I wonder whether this test is worth it. The idea of testing that the training pipeline works using augmentations is appealing, but costs a lot in CI time, which we would like to keep low. If we wanted to do this, we could not write this kind of test for every augmentation, so we would probably want to do something generic (not specific to SegmentationReconsruction augmentation). As for now, I think the test at line 710 is enough to check pipeline compliance and would remove this one.

What do you think @bruAristimunha ?

test/unit_tests/models/test_integration.py
cedricrommel, line 1: Any reason why we are changing this in this PR? It seems unrelated to the PR topic IMO. I would do it in another PR  or at list enrich the commit message on why this is being done here.

cedricrommel (changes requested):
Thanks a lot for the contribution @gustavohenriquesr ! It's been a while I wanted an augmentation like this to be added to the library and I'm glad you are doing it!

I think the PR is great and have just a few comments. Most of them are small things, except maybe for the location where randomness should be introduced: we made the design choice of only sampling:doing random operations outside `functionals.py`, to stay close to torchvision design.

test/unit_tests/augmentation/test_transforms.py
bruAristimunha: Hi @cedricrommel! =)

I created this extra test to catch an error not being seen with the test on line 710. The mask integration was not working correctly. @gustavohenriquesr solved here: https://github.com/braindecode/braindecode/pull/608/commits/def88c59891b63c936aefc18503b327b93f3cb94

Regarding creating this style of test integration, I understand that it increases CI time a little, but since [573](https://github.com/braindecode/braindecode/pull/573) and [571](https://github.com/braindecode/braindecode/pull/571), we have been adopting this approach, as some models were not working in our braindecode contract. @agramfort made the same point you made in the past, but as the benefits are necessary, it's okay to have this integration test. Alex comments [here](https://github.com/braindecode/braindecode/pull/573#discussion_r1491712651).

Regarding this specific test, it is outside our top 10 tests that take the most time; I suggest keeping it and modifying it to be generic for all augmentations if you want.

I spent time optimizing testing and documentation during the last sprint. If necessary, the secret will be to optimize the FakeData loader more.

test/unit_tests/models/test_integration.py
bruAristimunha: Indeed, unrelated, I think I'll improve the commit message. It was just to improve access to this type of dataloader

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+
+
+def _segmentation_reconstruction(X, n_segments, rng):
+    n_trials, n_channels, window_size = X.shape
+    # Segment Size
+    segment_size = window_size // n_segments
+    # Initialize an empty tensor for augmented data
+    X_aug = torch.zeros_like(X)
+    # Generate random indices within the class-specific dataset
+    rand_idx = rng.randint(0, n_trials, (n_trials, n_segments))
```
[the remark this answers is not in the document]
gustavohenriquesr: Hi @cedricrommel,

I was studying this random samplings and I think the only way to move it to transforms.py would copy practically the whole functional in transforms.py, since the samplings envolves iterations over the batch (analogous problem in GaussianNoise, in which we would have to move all function transforms.py also). If you have a suggestion that I'm not seeing, please, send me a message so I can fix it!

braindecode/augmentation/functional.py
```diff
@@ -969,3 +970,115 @@ def mixup(X, y, lam, idx_perm):
...
+
+    See [1]_ for details.
+
+    Parameters
+    ----------
+    X : torch.Tensor
+        EEG input example or batch.
+    y : torch.Tensor
+        EEG labels for the example or batch.
+    n_segments : int
```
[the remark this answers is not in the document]
gustavohenriquesr: Done!

braindecode/augmentation/functional.py
gustavohenriquesr: Hi @cedricrommel,

The choice for the default value came from some tests I did in a recent work (yet to be published) in which I use this type of augmentation and also some results I obtained while testing the integration of the technique in braindecode. Firstly, the factors of the number of samples seem to perform better than random values (since the segmentation forms a trial with the right size, without the need for padding). Secondly, in the datasets I tested (mainly BNCI2014_001), a “reasonable” number of segments (close to the square root of the number of samples) seems to have a better result than too many segments (probably due to the loss of coherence with so many different segments coming together to form a single trial). I therefore chose to use the maximum number of segments within this “reasonable” range for a general context. We can do more tests in the future to find a better default value, if possible.

braindecode/augmentation/transforms.py
gustavohenriquesr: Done!

gustavohenriquesr:
@bruAristimunha 

I have removed some functional tests that no longer apply with the new rearrangement of functions. The cases handled in the functions are now at the class level (transforms.py) and are no longer useful for functional testing. We can add more tests to the transform class, if you think it's necessary.

braindecode/augmentation/functional.py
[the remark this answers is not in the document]
cedricrommel: Perfect ! That’s precisely what I had in mind :)

cedricrommel (approved):
LGTM!!

Thanks for the changes and explanations!

bruAristimunha:
Thank you so much, @gustavohenriquesr! I hope this makes our script simpler with this augmentation now integrated, and I am hoping for the paper to be accepted! 🤞🏽 

Thank you so much, too, for your revision, @cedricrommel; I would love to have more reviews from you here if you have time for open source 🙏🏽

gustavohenriquesr:
Glad to contribute! I also thank you @bruAristimunha @cedricrommel for helping me improve the code. I hope to be able to make further contributions in the future!

PR #611, opened 2024-05-15
Adding SPARCNet and ContraWR

bruAristimunha:
In Biot's article, in addition to the BIOT model, the author centralizes codes from other models.

The list is:

1. SPaRCNet,
2. ContraWR,
3. CNN-Transformer,
4. FFCL and
5. ST-Transformer (predecessor of EEG Conformer, which we already have implemented here).

In this PR, I am proposing to implement 1) SPaRCNet and 2) ContraWR. 

3) CNN-Transformer seemed very similar to ContraWR, I don't think it's worth implementing.
4) The FFCL does not appear to be correctly implemented.

### DIFF
--- a/braindecode/models/sparcnet.py
+++ b/braindecode/models/sparcnet.py
@@ -255,7 +255,7 @@ class SPARCNet(EEGModuleMixin, nn.Module):
             input_window_seconds=input_window_seconds,
             sfreq=sfreq,
         )
-        del n_outputs, n_chans, chs_info, n_times, sfreq
+        del n_outputs, n_chans, chs_info, n_times, sfreq, input_window_seconds
 
         # add initial convolutional layer
         # the number of output channels is the smallest power of 2
--- a/braindecode/models/sparcnet.py
+++ b/braindecode/models/sparcnet.py
@@ -1,0 +1,353 @@
+from __future__ import annotations
+
+import math
+from collections import OrderedDict
+
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+from .base import EEGModuleMixin
+
+
+class DenseLayer(nn.Sequential):
+    """
+    A densely connected layer with batch normalization and dropout.
+
+    Parameters
+    ----------
+    input_channels : int
+        Number of input channels.
+    growth_rate : int
+        Rate of growth of channels in this layer.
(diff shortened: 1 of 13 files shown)


bruAristimunha:
fyi @mccorsi

braindecode/models/sparcnet.py
```diff
@@ -0,0 +1,353 @@
...
+    --------
+    >>> x = torch.randn(128, 5, 1000)
+    >>> model = TransitionLayer(5, 18)
+    >>> y = model(x)
+    >>> y.shape
+    torch.Size([128, 18, 500])
+    """
+
+    def __init__(
+        self, input_channels, output_channels, conv_bias=True, batch_norm=True
```
agramfort, line 171: you should aim for consistency with pytorch naming conventions as much as possible so there use in_channels, out_channels

braindecode/models/sparcnet.py
```diff
@@ -0,0 +1,353 @@
...
+    This implementation is not guaranteed to be correct, has not been checked
+    by original authors.
+
+    Parameters
+    ----------
+    block_layers : int, optional
+        Number of layers per dense block. Default is 4.
+    growth_rate : int, optional
+        Growth rate of the DenseNet. Default is 16.
+    bn_size : int, optional
```
agramfort, line 211: bn sounds like batch_norm on first read. maybe use bottleneck_size to be explicit

braindecode/models/sparcnet.py
```diff
@@ -0,0 +1,353 @@
...
+            n_times=n_times,
+            input_window_seconds=input_window_seconds,
+            sfreq=sfreq,
+        )
+        del n_outputs, n_chans, chs_info, n_times, sfreq, input_window_seconds
+
+        # add initial convolutional layer
+        # the number of output channels is the smallest power of 2
+        # that is greater than the number of input channels
+        out_channels = 2 ** (math.floor(np.log2(self.n_chans)) + 1)
```
agramfort, line 263: why not use math.log2 here as you use math.floor

agramfort (commented):
at some point it would be good to do a full evaluation of all these architectures on a some standard datasets

bruAristimunha:
The green button is yours @agramfort

PR #612, opened 2024-05-17
Fix `tmax` in `SleepPhysionet`

OverLordGoldDragon:
## Summary

Fix the right crop boundary in `SleepPhysionet` so wake-margin cropping keeps the complete final sleep annotation and uses an explicit half-open interval.

## What changed

- compute the exclusive stop from the final sleep annotation's onset **plus its duration**, then add the requested wake margin;
- call `Raw.crop(..., include_tmax=False)` instead of subtracting one sample manually;
- clamp a right-edge crop to the valid one-past-the-last-sample boundary, preserving the recording's final sample;
- update the live integration assertion to check the semantic 30-minute margins around the complete annotated sleep interval;
- add deterministic, network-free 100 Hz regressions for ordinary, left-clamped, and right-clamped intervals.

The change does not alter channel handling, annotation parsing, serialization, or public constructor arguments.

## Validation

- focused offline tests: 3 passed, 5 network tests skipped;
- live Sleep-EDF subject-0/night-1 suite: 8 passed;
- lazy and preloaded live reads agree at 2,523,000 samples and retain exactly 1,800 seconds of wake margin on each side;
- full dataset suite: 214 passed, 5 skipped;
- all applicable pre-commit hooks and diff checks pass;
- fresh independent review approved exact commit `d0e9e38e5b9d7b320432cee5b773e9ee0a802114` with no findings.

Dataset source and checksums were verified against PhysioNet Sleep-EDF Expanded. No data or third-party code is redistributed by this change.


agramfort:
that gives you now samples of 3000 values for 30s at 100Hz and before it was 3001?

OverLordGoldDragon:
I don't follow, can you clarify?

OverLordGoldDragon:
I don't understand `assert duration_h < 7 and duration_h > 6` in `test_sleep_physionet.py`, where did these values come from? I checked and everything seems sensible, with `duration_h` being `7.008330555555555`.

agramfort:
if you are breaking a test it suggests there is a reason things are done
this way? sorry I don't have the full background here.
Why do you think it's a bug?

OverLordGoldDragon:
> `annots[idx]["onset"]` is inclusive on time *starts*, so to include all of `idx`, add duration and decrement a sample

Put differently, the intent is to include the entire segment `annots[idx]`, but that's not what happens, only the very first sample is being included. I.e. currently, if `x = data[start:end]`, it's

```
start = onset
end = onset + 1
```

whereas it should be

```
start = onset
end = onset + duration
```

OverLordGoldDragon:
Also, the existing code fails the new assertion I added (it's off by one sample, as expected), mine doesn't. The assertion's motivated by https://sci-hub.st/10.1109/TNSRE.2018.2813138 ,

> a human sleep expert looks at the different time series recorded over the night and assigns to each 30 s time segment
a sleep stage following a reference nomenclature such as ...

braindecode/datasets/sleep_physionet.py
```diff
@@ -102,8 +102,11 @@ def _load_raw(
...
 
             # Crop raw
-            tmin = annots[int(sleep_event_inds[0])]["onset"] - crop_wake_mins * 60
-            tmax = annots[int(sleep_event_inds[-1])]["onset"] + crop_wake_mins * 60
+            dT = 1 / raw.info["sfreq"]
+            a_tmin = annots[sleep_event_inds[0]]
+            a_tmax = annots[sleep_event_inds[-1]]
+            tmin = a_tmin["onset"] - crop_wake_mins * 60
+            tmax = a_tmax["onset"] + a_tmax["duration"] - dT + crop_wake_mins * 60
             raw.crop(tmin=max(tmin, raw.times[0]), tmax=min(tmax, raw.times[-1]))
```
agramfort, line 110: maybe you just want to keep old version but set include_tmax=False in raw.crop?

OverLordGoldDragon: That only resolves `- dT`.

agramfort: can you already fix accordingly? also can you see why CIs fail?

bruAristimunha: Addressed on exact head d0e9e38e. The crop now uses the half-open MNE contract explicitly (`include_tmax=False`), includes the complete final annotation duration, and clamps a physical recording boundary to the one-past-end sample. The offline regression spies the `Raw.crop` call and covers ordinary, left-clamped, and right-clamped intervals; an independent live Sleep-EDF check confirmed 2,523,000 samples, exact 1,800 s margins around the complete final annotation, lazy/preloaded equality, and preservation of the final sample. Fresh exact-SHA CI is now fully green, including the retried documentation build.

PR #613, opened 2024-05-17
Remove redundant `eval()` & `train()` in `SleepStagerChambon2018`

OverLordGoldDragon:
They don't affect shapes.

This doesn't seem to belong anywhere in `whats_new`, nor is it important, so I omitted, but feel free to edit.

### DIFF
--- a/braindecode/models/sleep_stager_chambon_2018.py
+++ b/braindecode/models/sleep_stager_chambon_2018.py
@@ -133,10 +133,8 @@ class SleepStagerChambon2018(EEGModuleMixin, nn.Module):
             )
 
     def _len_last_layer(self, n_channels, input_size):
-        self.feature_extractor.eval()
         with torch.no_grad():
             out = self.feature_extractor(torch.Tensor(1, 1, n_channels, input_size))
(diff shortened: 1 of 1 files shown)


braindecode/models/sleep_stager_chambon_2018.py
```diff
@@ -133,10 +133,8 @@ def __init__(
             )
 
     def _len_last_layer(self, n_channels, input_size):
-        self.feature_extractor.eval()
```
agramfort, line 136: why do you say it's redundant? layers like drop out need it and it's not the same as no_grad

OverLordGoldDragon: The method operates on output shapes, which are unaffected.

agramfort:
I am not convinced we should merge this one.

OverLordGoldDragon:
> but is it a bad practice to move the model to eval mode when you are not
training?

No, it's just confusing as it implies intent, i.e. that it'd be different if we removed the code. (I found it confusing and had to look into it).

Feel free to proceed as seen fit.

PR #617, opened 2024-05-17
[MNT] Fix matplotlib colormaps deprecation

bruAristimunha:
Matplotlib, since version 3.9, released yesterday, May 17th, does not support `get_cmap` function. They recommend using `colormap[name]` instead.

`colormap` was introduced in matplotlib 3.5 on May 16/11/2021, and since then, they have warned users to change this get_cmap.

The matplotlib warning:

```bash
 MatplotlibDeprecationWarning: The get_cmap function was deprecated in Matplotlib 3.7 and will be removed two minor releases later. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
  colors = cm.get_cmap("viridis", 5)(range(5))
```
As colormap has been present in matplotlib since 2021, it doesn't affect users or force users to update matplotlib.


agramfort:
will this work with older versions too or are you forcing people to upgrade?

bruAristimunha:
Hi @agramfort,

I updated the description to include more details in this PR and why it does not affect the users, in my opinion.

PR #629, opened 2024-06-03
MNT Removing MNE version pin 

bruAristimunha:


bruAristimunha:
The issue was solve on moabb side

Issue #628, opened 2024-05-31
Include more data augmentation


bruAristimunha:
The idea of the issue is to list and include more augmentation for braindecode. 

We would like to integrate something related to source estimation, interpolation, or masking augmentation that has already been published. 

Can you work on this @gustavohenriquesr? 

A little bird told me that you need to make more PR for braindecode ;)
 
Nice start point:
https://github.com/Navidfoumani/EEG2Rep

gustavohenriquesr:
Hi @bruAristimunha 

Sure, I will work on it! Thank you for the reference.

bruAristimunha:
Hey @gustavohenriquesr,

Some deadlines for you:

- Create the draft PR (07/06/2024);
- Create the tests (07/06/2024, maximum 11/06/2024)
- Integrate the PR (14/06/2024)

PR #631, opened 2024-06-07
Add MaskEncoding augmentation

gustavohenriquesr:
Hi, I've implemented another data augmentation class called MaskEncoding. 

The idea is similar to SmoothTimeMask but there is no smoothing of the masking. In addition, the masking can be divided into parts and spread across the signal (these parts can overlap, and I chose to leave it this way). This implementation was initially based on [this article](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=10439237), and I modified it as I saw appropriate.

Adresses #628

### DIFF
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -1043,7 +1043,7 @@ def segmentation_reconstruction(
     return aug_data, y
 
 
-def mask_encoding(X, y, time_start, segment_length, splits):
+def mask_encoding(X, y, time_start, segment_length, n_segments):
     """Replaces a contiguous part (or parts) of all channels by zeros
     (if more than one segment, it may overlap).
 
@@ -1057,11 +1057,11 @@ def mask_encoding(X, y, time_start, segment_length, splits):
          EEG labels for the example or batch.
      time_start : torch.Tensor
          Tensor of integers containing the position (in last dimension) where to
-         start masking the signal. Should have "splits" times the size of the first
-         dimension of X (i.e. "splits" start positions per example in the batch).
+         start masking the signal. Should have "n_segments" times the size of the first
+         dimension of X (i.e. "n_segments" start positions per example in the batch).
      segment_length: int
          Length of each segment to zero out.
-     splits : int
+     n_segments : int
          Number of segments to zero out in each example.
 
      Returns
@@ -1079,7 +1079,7 @@ def mask_encoding(X, y, time_start, segment_length, splits):
      32 (2024): 875-886.
     """
 
-    batch_indices = torch.arange(X.shape[0]).repeat_interleave(splits)
+    batch_indices = torch.arange(X.shape[0]).repeat_interleave(n_segments)
     start_indices = time_start.flatten()
     mask_indices = start_indices[:, None] + torch.arange(segment_length)
 
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -1041,3 +1041,54 @@ def segmentation_reconstruction(
         return aug_data, aug_label
 
     return aug_data, y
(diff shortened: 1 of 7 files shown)


braindecode/augmentation/functional.py
```diff
@@ -1041,3 +1041,54 @@ def segmentation_reconstruction(
...
+         EEG input example or batch.
+     y : torch.Tensor
+         EEG labels for the example or batch.
+     time_start : torch.Tensor
+         Tensor of integers containing the position (in last dimension) where to
+         start masking the signal. Should have "splits" times the size of the first
+         dimension of X (i.e. "splits" start positions per example in the batch).
+     segment_length: int
+         Length of each segment to zero out.
+     splits : int
```
agramfort, line 1064:
```suggestion
     n_segments : int
```
as you use the word segment elsewhere.

test/unit_tests/augmentation/test_transforms.py
```diff
@@ -690,6 +691,61 @@ def test_segmentation_rec_with_large_n_segments(time_aranged_batch):
...
+            # Check max_mask_ratio cannot be outside interval [0,1] and
+            # splits must be a positive integer
+            transform = MaskEncoding(
+                1.0, max_mask_ratio=max_mask_ratio,
+                splits=splits,
+                random_state=rng_seed,
+            )
+            # Check splits cannot be higher than (max_mask_ratio * window_size)
+            ones_batch = ones_and_zeros_batch()
+            common_tranform_assertions(
```
agramfort, line 722:
```suggestion
            common_transform_assertions(
```

bruAristimunha:
Hey @gustavohenriquesr,

The code is already good, so I will ask something else: can you measure the impact of augmentation using the ShallowNet and BCNI datasets? To integrate something, we need to have some interesting results! 

The idea would be to do a cross-subject approach with 5-fold, very similar to [Cedric](https://braindecode.org/stable/auto_examples/advanced_training/plot_data_augmentation_search.html)'s tutorial, but with a splitter for cross-validation with the augmentation. You can decide whether to validate the augmentation parameters or the probability. Use the same parameters from https://github.com/eeg-augmentation-benchmark/eeg-augmentation-benchmark-2022/blob/1df778f7359581b1297d18dbdabf2b2fd4fcfa2f/eeg_augmentation_benchmark/models.py#L64C1-L78C57 

Small inspiration to build the cross-validation object:
- the BraindecodeKFold from @dengemann, @hubertjb and @gemeinl: https://github.com/meeg-ml-benchmarks/brain-age-benchmark-paper/blob/c57eba38c8b90dac9354f0f4f8148dbed7e56029/deep_learning_utils.py#L25

- the benchopt splitters from @tomMoral: https://github.com/benchopt/benchmark_bci/blob/main/benchmark_utils/splitter.py

If you need an extension for the deadline, ping me on Discord ;)

gustavohenriquesr:
Hello, @bruAristimunha 

Thank you for the patience.

I did the tests you requested using a simplified strategy of what @cedricrommel [proposed](https://arxiv.org/pdf/2206.14483) earlier for some other augmentations. The results show that using more than one segment benefits the model, while very high masking ratios when `n_segments = 1` worsen performance. When we increase the number of segments, as in this case I tested with 5 segments, the model responds well up to a certain masking ratio. 

Leave it to you to decide whether it's worth adding this method to Braindecode. 

Later, with more time, I can do more tests with a wider range of values if you like.

splits

mask_ratio

mask_ratio_5

bruAristimunha (approved):

bruAristimunha:
Amazing work @gustavohenriquesr ;)

@agramfort, the green button is your!

agramfort (approved):

PR #640, opened 2024-09-24
EHN: including EEGTCNet

bruAristimunha:
With this neural network, we can fully remove deep learning support from moabb. Related with benchopt BCI, cc. @tomMoral.


bruAristimunha:
The global comments about this code... It is already good, but it does not use the braindecode dependencies (EEGNetV4 and TCN). The main limitation is the dimensionality issue that I am thinking more about.

PR #643, opened 2024-09-25
EHN: Including Syncnet

bruAristimunha:
Describe in https://papers.nips.cc/paper_files/paper/2017/hash/7993e11204b215b27694b6f139e34ce8-Abstract.html


bruAristimunha:
The windows issue is not related: https://github.com/h5py/h5py/issues/2505

agramfort (commented):
is this model tested? did you run it on some data to see if it gives sensible results?

bruAristimunha:
Yes @agramfort!

I tested with moabb, the competition dataset cross-session and results were compatible~same range with eegnetv4.

I trained for 50 epochs and didn't explore any parameters



Since this model is a bit different, I think the learning and other parameters will be a bit different

A lot of oscillation... I can include the benchmark figure on the ongoing models PRs

PR #645, opened 2024-09-25
[EHN] Including EEGChannelNet

bruAristimunha:


bruAristimunha:
Nevermind, this model is not good

PR #647, opened 2024-09-26
[EHN] Including LGGNET

bruAristimunha:
This architecture is a little more complicated. Together with MEET, EEGProgress, EEGSym, and Tsinception (#641), the model needs the channel names with indexes for different reasons. 

I need help deciding if I should include channel_list into the base braindecode or infer or even leave this to be informed when defining the model. (This breaks the convention we established in the last sprint)....

Ideas about the design @agramfort, @tomMoral, @PierreGtch, @robintibor?

### DIFF
--- a/braindecode/models/__init__.py
+++ b/braindecode/models/__init__.py
@@ -32,6 +32,7 @@ from .contrawr import ContraWR
 from .eegnex import EEGNeX
 from .eegtcnet import EEGTCNet
 from .syncnet import SyncNet
+from .lggnet import LGGNet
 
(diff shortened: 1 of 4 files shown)


PierreGtch:
@bruAristimunha We already have the list of channel names through the 'chs_info' parameter :) 

'chs_info[i]["name"]'

bruAristimunha:
oh yet! Suuuuper true @PierreGtch  I was just remembering that it was a dictionary, not that it was a list. All solved =) I will integrate the rest of the models too ;)

PR #656, opened 2024-10-01
[EHN] Filterbank as module

bruAristimunha:
In this PR, we've implemented a FilterBank inside a Pytorch Module, we have a very strong dependency on MNE and we haven't implemented a way to apply the IIR filter.

Basically we use mne to create the filter and use fftconvolve to apply the filter. The implementation of fftconvolve comes from pytorch audio. 

What I need help with here, as not an expert in filtering:

- Check that I haven't made any fundamental mistakes, such as additional dimensions or related details.
- Consider whether this way of applying the bandpass, using fftconv, is correct.
- Suggest ways of optimising the code.
- Request tests and requirements for merging.


bruAristimunha:
Small code to test

```python
    from torch import randn

    x = randn(16, 10, 1000)

    layer = FilterBank(sfreq=256, nchans=10)
    with torch.no_grad():
        out = layer(x)

    print(out.shape)
```

bruAristimunha:
How is make in mne
```python

import numpy as np

from mne.filter import _check_zero_phase_length, next_fast_len
from mne.utils import logger


def _smart_pad_torch(x, n_pad, pad="reflect_limited"):
    """Pad vector x."""
    n_pad = np.asarray(n_pad)
    assert n_pad.shape == (2,)
    if (n_pad == 0).all():
        return x
    elif (n_pad < 0).any():
        raise RuntimeError("n_pad must be non-negative")
    if pad == "reflect_limited":
        # need to pad with zeros if len(x) <= npad
        left_zero_pad = torch.zeros(max(n_pad[0] - len(x) + 1, 0))
        right_zero_pad = torch.zeros(max(n_pad[1] - len(x) + 1, 0))
        reflection_left = 2 * x[0]
        reflection_right = 2 * x[-1]
        padded = torch.cat(
            [
                left_zero_pad,
                reflection_left - torch.flip(x[1 : n_pad[0] + 1], dims=[0]),
                x,
                reflection_right - torch.flip(x[-n_pad[1] - 2 + 1 : -1], dims=[0]),
                right_zero_pad,
            ]
        )

        return padded

    else:
        left_pad, right_pad = n_pad.tolist()
        return torch.nn.functional.pad(x, (left_pad, right_pad), mode=pad)


def _1d_overlap_filter_torch(x, n_h, n_edge, phase, h_fft, pad, n_fft):
    """Do one-dimensional overlap-add FFT FIR filtering using PyTorch."""

    # Pad to reduce ringing
    x_ext = _smart_pad_torch(x, (n_edge, n_edge), pad)
    n_x = x_ext.shape[0]
    x_filtered = torch.zeros_like(x_ext)

    n_seg = n_fft - n_h + 1
    n_segments = (n_x + n_seg - 1) // n_seg  # Equivalent to ceil division
    shift = ((n_h - 1) // 2 if phase.startswith("zero") else 0) + n_edge

    # Actual filtering step
    for seg_idx in range(n_segments):
        start = seg_idx * n_seg
        stop = (seg_idx + 1) * n_seg
        seg = x_ext[start:stop]

        # Pad segment to length n_fft
        seg = torch.cat(
            [seg, torch.zeros(n_fft - seg.shape[0], dtype=seg.dtype, device=seg.device)]
        )

        # FFT of the segment
        x_fft = torch.fft.rfft(seg, n=n_fft)
        x_fft *= h_fft
        prod = torch.fft.irfft(x_fft, n=n_fft)

        # Overlap-add operation
        start_filt = max(0, start - shift)
        stop_filt = min(start - shift + n_fft, n_x)
        start_prod = max(0, shift - start)
        stop_prod = start_prod + stop_filt - start_filt
        x_filtered[start_filt:stop_filt] += prod[start_prod:stop_prod]

    # Remove mirrored edges and cast to original dtype
    x_filtered = x_filtered[: n_x - 2 * n_edge].type_as(x)
    return x_filtered


def _overlap_add_filter_torch(
    x,
    h,
    n_fft=None,
    phase="zero",
    pad="reflect_limited",
):
    """Filter the signal x using h with overlap-add FFTs."""
    # set up array for filtering, reshape to 2D, operate on last axis
    orig_shape = x.shape
    nchans = orig_shape[1]
    # reshaping data to 2D
    x = x.view(-1, x.shape[-1])
    # Extend the signal by mirroring the edges to reduce transient filter
    # response
    _check_zero_phase_length(len(h), phase)
    if len(h) == 1:
        return x * h**2 if phase == "zero-double" else x * h
    n_edge = max(min(len(h), x.shape[1]) - 1, 0)
    logger.debug(f"Smart-padding with:  {n_edge} samples on each edge")
    n_x = x.shape[1] + 2 * n_edge

    # Determine FFT length to use
    min_fft = 2 * len(h) - 1
    if n_fft is None:
        max_fft = n_x
        if max_fft >= min_fft:
            # cost function based on number of multiplications
            N = 2 ** np.arange(
                np.ceil(np.log2(min_fft)), np.ceil(np.log2(max_fft)) + 1, dtype=int
            )
            cost = (
                np.ceil(n_x / (N - len(h) + 1).astype(np.float64))
                * N
                * (np.log2(N) + 1)
            )

            # add a heuristic term to prevent too-long FFT's which are slow
            # (not predicted by mult. cost alone, 4e-5 exp. determined)
            cost += 4e-5 * N * n_x

            n_fft = N[np.argmin(cost)]
        else:
            # Use only a single block
            n_fft = next_fast_len(min_fft)
    logger.debug(f"FFT block length:   {n_fft}")
    if n_fft < min_fft:
        raise ValueError(
            f"n_fft is too short, has to be at least 2 * len(h) - 1 ({min_fft}), got "
            f"{n_fft}"
        )

    # Figure out if we should use CUDA
    h_fft_torch = torch.fft.rfft(h, n=n_fft)

    # Process each row separately
    saving = [None] * nchans
    for chan in range(nchans):
        saving[chan] = _1d_overlap_filter_torch(
            x[chan], len(h), n_edge, phase, h_fft=h_fft_torch, pad=pad, n_fft=n_fft
        )

    return torch.stack(saving)


```

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
-            return waveform
-        return _apply_sinc_resample_kernel(
-            waveform, self.orig_freq, self.new_freq, self.gcd, self.kernel, self.width
-        )
+        Apply the filter bank to the input signal.
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            Input tensor of shape (batch_size, time_points).
```
tomMoral, line 744: missing `n_chans`?

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+                verbose=verbose,
             )
-            self.register_buffer("kernel", kernel)
+            if not self.method_iir:
+                filts[f"{idx}_band"] = from_numpy(filt).float()
+            else:
+                b = from_numpy(filt["b"]).float()
+                a = from_numpy(filt["a"]).float()
+
+                filts[f"{idx}_band"] = (b, a)
```
tomMoral, line 733: Maybe more readable?
```suggestion
                filts[f"band_{idx}"] = (b, a)
```

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+            )
+            low_freq = 4
+            high_freq = 40
+
+            band_filters = self._create_band_filters(low_freq, high_freq, band_filters)
+        if not isinstance(band_filters, list):
+            ValueError(
+                "`band_filters` should be a list of tuples if you want to "
+                "use them this way."
+            )
```
tomMoral, line 687: Add check that the `band_filters[0]` is splitable in 2 values.

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+                b = from_numpy(filt["b"]).float()
+                a = from_numpy(filt["a"]).float()
+
+                filts[f"{idx}_band"] = (b, a)
 
-    def forward(self, waveform: Tensor) -> Tensor:
-        r"""
-        Args:
-            waveform (Tensor): Tensor of audio of dimension (..., time).
+            self.filts = filts
```
tomMoral, line 735: If you want to be able to move to GPU, you need to store this as
```suggestion
            self.filts = nn.ModuleDict(filts)
```

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+
+        for band_idx in range(n_bands):
+            if not self.method_iir:
+                filtered = self._apply_fir(
+                    x, self.filts[f"{band_idx}_band"], self.n_chans
+                )
+            else:
+                filtered = self._apply_irr(
+                    x, self.filts[f"{band_idx}_band"], self.n_bands
+                )
```
tomMoral, line 763: I would rather have the section of this method in the `__init__`, stored as `self._apply_filter_func` to avoid having the `if` in the forward.

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            Input tensor of shape (batch_size, time_points).
+
+        Returns
+        -------
+        torch.Tensor
+            Filtered output tensor of shape (batch_size, n_bands, filtered_time_points).
```
tomMoral, line 749: missing `n_chans`

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+            if i == num_bands - 1:
+                current_end = end
+            else:
+                current_end = current_start + base_interval_size - 1
+            intervals.append((current_start, current_end))
+            current_start = current_end + 1
+
+        # Convert intervals to a list of tuples (low_freq, high_freq)
+        band_filters = [(low, high) for low, high in intervals]
+        return band_filters
```
tomMoral, line 792: seems weird. Why not:
```
        intervals = torch.linspace(start, end, (end - start + 1) // num_bands)
        band_filters = [(low, high) for low, high in zip(intervals, intervals[1:])]
```

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+        # Add band dimension
+        # Shape: (batch_size, 1, nchans, time_points)
+        filtered = filtered.unsqueeze(1)
+        # returning the filtered
+        return filtered
+
+    @staticmethod
+    def _apply_irr(x: Tensor, filter, n_bands: int) -> Tensor:
+        x = x.unsqueeze(2).repeat(1, 1, n_bands, 1)
+        x = filtfilt(x, filter["a"], filter["a"], clamp=False)
```
tomMoral, line 815: All this should be tested to check that the filtering is done right.
```suggestion
        x = filtfilt(x, filter["a"], filter["b"], clamp=False)
```

braindecode/models/modules.py
```diff
@@ -528,102 +529,289 @@ def extra_repr(self) -> str:
...
+
+        # Add band dimension
+        # Shape: (batch_size, 1, nchans, time_points)
+        filtered = filtered.unsqueeze(1)
+        # returning the filtered
+        return filtered
+
+    @staticmethod
+    def _apply_irr(x: Tensor, filter, n_bands: int) -> Tensor:
+        x = x.unsqueeze(2).repeat(1, 1, n_bands, 1)
```
tomMoral, line 814: I don't think you need the repeat: https://pytorch.org/audio/main/generated/torchaudio.functional.filtfilt.html

tomMoral (commented):
I did a first pass.
I would be much more confident with tests that check that you obtain the same results with `mne` filtering modules.

braindecode/models/eegsimpleconv.py
```diff
@@ -13,8 +13,8 @@
 
 import torch
 from torch import nn
+from torchaudio.transforms import Resample
 
-from braindecode.models.modules import Resample
```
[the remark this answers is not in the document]
bruAristimunha: Basically, I include torch audio as a dependency for braindecode. Until now, for some functions, we had to copy the code from torch audio to avoid dependency.

With the filterbank module, we would have to add a good part of the filter module, so I decided to directly include torch audio and get rid of all the copied code.

I'll improve the PR description by including this detail.

braindecode/models/modules.py
bruAristimunha: okay, removed

braindecode/models/modules.py
bruAristimunha: Logic changed

braindecode/models/modules.py
bruAristimunha: Now store at the GPU

braindecode/models/modules.py
bruAristimunha: checker made

bruAristimunha:
Hey @qbarthelemy,

Here, I have a filtering implementation using pytorch, which should reproduce the mne results. Currently, I cannot reproduce the results of mne using pytorch, and I need help.

From what I have noticed in the past few days, there are several reasons for this.

1) Implementation differences. The obvious difference is related to a padding issue; in SciPy we perform a padding that we are not performing here. Other people have already noticed this here: https://github.com/pytorch/audio/issues/2063

Even disabled all the options related to padding, I am still noticing this difference: small code to reproduce this issue. I am not sure if it is something relevant or not. For the IIR filter, almost everything looks the same, except in the beginning :X

```python
import numpy as np
import matplotlib.pyplot as plt
from torch import randn, from_numpy
from mne.filter import create_filter
from torchaudio.functional import filtfilt as filtfilt_torch
from scipy.signal import filtfilt as filtfilt_scipy

n_chans = 10
x = randn(1, n_chans, 1000)
x = x.unsqueeze(2)  # Shape: (batch_size, n_chans, 1, n_times)

filt = create_filter(data=None, sfreq=256, l_freq=4, h_freq=8, method="iir", iir_params=dict(output="ba")) # creating iir filter
a_torch = from_numpy(filt["a"]).float()
b_torch = from_numpy(filt["b"]).float()

# Apply filtering using torchaudio's filtfilt
filtered_torch = filtfilt_torch(waveform=x, a_coeffs=a_torch, b_coeffs=b_torch, clamp=False)

filtered_scipy = filtfilt_scipy(b=filt["b"], a=filt["a"], padtype=None, x=x.numpy())

difference = np.abs(filtered_torch.numpy()-filtered_scipy)

plt.plot(difference.flatten())
plt.show()
```
image


Reproduce difference in FIR filters:
```python
import numpy as np
import matplotlib.pyplot as plt
from torch import randn, from_numpy
from mne.filter import create_filter
from torchaudio.functional import fftconvolve as fftconvolve_torch
from scipy.signal import fftconvolve as fftconvolve_scipy


batch_size = 1
n_chans = 22

x = randn(batch_size, n_chans, 1000)

filt = create_filter(data=None, sfreq=256, l_freq=4, h_freq=8) # creating fir filter

filt_expanded = from_numpy(filt).unsqueeze(0).repeat(n_chans, 1).unsqueeze(0)

filtered_torch = fftconvolve_torch(
    x, filt_expanded, mode="same"
)

# Apply convolution using scipy's fftconvolve (per channel)
x_numpy = x.numpy()  # Shape: (1, n_chans, n_times)
filt_numpy = filt_expanded.numpy()  # Shape: (1, n_chans, filter_length)

# Initialize array to hold scipy-filtered data
filtered_scipy = np.zeros_like(x_numpy)

#  avoid cross-channel convolution
for batch in range(batch_size):
    for chan in range(n_chans):
        signal = x_numpy[batch, chan, :]  # Shape: (n_times,)
        filter_chan = filt_numpy[batch, chan, :]  # Shape: (filter_length,)

        convolved = fftconvolve_scipy(signal, filter_chan, mode='same')

        # Store the convolved signal
        filtered_scipy[batch, chan, :] = convolved

difference = np.abs(filtered_torch.numpy()-filtered_scipy)

#plt.plot(x.numpy().flatten(), label="original")
plt.plot(filtered_torch.numpy().flatten(), label="torch")
plt.plot(filtered_scipy.flatten(), label="scipy")
plt.plot(difference.flatten(), label="absolute diff")
plt.legend()
plt.show()

```
image


2) I am very unsure if applying the filters (IIR and FIR) this way is correct or the best way.
MNE apply filters in a very specific way.

> If it is FIR, I am applying:
>      fftconvolute(x, coeffi_b)
> If it is IIR, I am applying:
>     filt_ta = filtfilt(x, filter["a"], filter["b"], clamp=False)


3) It seems that IIR works well for frequencies above 12Hz, and for frequencies below 12Hz, FIR works better when we have a very narrow bandwidth. I made a test that captures this effect, but I am very uncertain about the test.

I think the general message and request for your review is that, functionally, everything is okay, but I am very uncertain if everything is correct.

The default filters created using the correct implementation:

image

qbarthelemy (commented):
My two cents for IIR filtering.

`filtfilt` of `torchaudio` simply applies `lfilter` to the signal, then to the returned signal,
see https://github.com/pytorch/audio/blob/main/src/torchaudio/functional/filtering.py#L701

`filtfilt` of `scipy`, before applying `lfilter` to the signal and to its returned version,
starts by adjusting the initial inner state `z_i`
https://github.com/scipy/scipy/blob/v1.14.1/scipy/signal/_signaltools.py#L4202
`z_i` is then used by `lfilter`, allowing to avoid the transient at the beginning of the filtered signal (so, at the beginning of each channel).

This initialization of inner state of filter can explain the differences you observe in numerical values. You can confirm adding a comparison only on `lfilter`.

For FIR filtering, I will look later.

bruAristimunha:
All good, @qbarthelemy =) Everything matches now, including IIR and FIR. Thank you so much for the revision! 


and @sylvchev, indeed, filtering is a kind of dark magic, millions of details.

qbarthelemy (approved):
Good job @bruAristimunha !

sylvchev (approved):

braindecode/models/modules.py
```diff
@@ -523,3 +529,329 @@ def forward(self, x):
...
+        ----------
+        x : torch.Tensor
+            Input tensor of shape (batch_size, time_points).
+        n_chans: int
+            Number of chans passed internally inside the object
+
+        Returns
+        -------
+        torch.Tensor
+            Filtered output tensor of shape (batch_size, n_bands, filtered_time_points).
```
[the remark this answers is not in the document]
bruAristimunha: You are correct!

braindecode/models/modules.py
```diff
@@ -523,3 +529,331 @@ def forward(self, x):
...
+        Width of the transition band at the high cut-off frequency in Hz
+        (low pass or cutoff 2 in bandpass). Can be "auto"
+        (default in 0.14) to use a multiple of ``h_freq``::
+
+            min(max(h_freq * 0.25, 2.), info['sfreq'] / 2. - h_freq)
+
+        Only used for ``method='fir'``.
+    method : str, default='fir'
+        ``'fir'`` will use FIR filtering, ``'iir'`` will use IIR
+        forward-backward filtering (via :func:`~scipy.signal.filtfilt`).
```
tomMoral, line 591: Move this one before the method specific arguments to make this easier to read (so before `filter_length`).

tomMoral: Maybe also add a link toward why to chose vs the other if you have one?

braindecode/models/modules.py
```diff
@@ -523,3 +529,331 @@ def forward(self, x):
...
+                if "output" in iir_params:
+                    if iir_params["output"] == "sos":
+                        warn(
+                            "It is not possible to use second-order section filtering with Torch. Changing to filter ba",
+                            UserWarning,
+                        )
+                        iir_params["output"] = "ba"
+
+        self._apply_filter_func = None
+
```
tomMoral, line 730: Not really useful no?
```suggestion
```

braindecode/models/modules.py
tomMoral: Aren't you missing `n_chans` in the shape?
```suggestion
            Input tensor of shape (batch_size, n_chans, time_points).
```

braindecode/models/modules.py
```diff
@@ -523,3 +529,331 @@ def forward(self, x):
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """
+        Apply the filter bank to the input signal.
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            Input tensor of shape (batch_size, time_points).
+        n_chans: int
+            Number of chans passed internally inside the object
```
tomMoral, line 775:
```suggestion
```

braindecode/models/modules.py
```diff
@@ -523,3 +529,331 @@ def forward(self, x):
...
+        n_chans: int
+            Number of channels
+
+        Returns
+        -------
+        Tensor
+            Filtered tensor of shape (batch_size, 1, n_chans, n_times).
+        """
+        # Shape: (nchans, filter_length)
+        # Expand to (nchans, filter_length)
```
tomMoral, line 815: Seems that the comments are wrong no?

braindecode/models/modules.py
```diff
@@ -523,3 +529,331 @@ def forward(self, x):
...
+                method=method,
+                iir_params=iir_params,
+                phase=phase,
+                fir_window=fir_window,
+                fir_design=fir_design,
+                verbose=verbose,
+            )
+            if not self.method_iir:
+                b = from_numpy(filt).float()
+                filts[f"band_{idx}"] = {"b": b}
```
tomMoral, line 750: Why not call `b` filt to avoid confusion with `a, b`?

braindecode/models/modules.py
```diff
@@ -523,3 +529,331 @@ def forward(self, x):
...
+        n_bands = self.n_bands
+        output = []
+
+        for band_idx in range(n_bands):
+            filtered = self._apply_filter_func(x, self.filts[f"band_{band_idx}"])  # type: ignore
+            output.append(filtered)
+
+        # Shape: (batch_size, n_bands, nchans, time_points)
+        output = torch.cat(output, dim=1)
+        return output
```
tomMoral, line 792: Isn't it faster to do:
```suggestion
        return torch.cat(
            self._apply_filter_func(x, p_filt) for p_filt in self.filts.values(),
            dim=1
         )
```

docs/whats_new.rst
```diff
@@ -51,6 +51,7 @@ Enhancements
 - Ensure consistency in the last layer using tests (:gh:`642` by `Bruno Aristimunha`_ )
 - Ensuring consistency on the expose of the activation function (:gh:`637` by `Bruno Aristimunha`_ )
 - Adding :class:`braindecode.models.SyncNet` (:gh:`643` by `Bruno Aristimunha`_ )
+- Creating the FilterBank module for new models (:gh:`656` by `Bruno Aristimunha`_ )
```
tomMoral, line 54:
```suggestion
- Creating the `FilterBanklayer` module for new models (:gh:`656` by `Bruno Aristimunha`_ )
```

test/unit_tests/models/test_modules.py
```diff
@@ -305,3 +329,413 @@ def test_safelog_extra_repr(eps, expected_repr):
...
+        n_chans=n_chans,
+        sfreq=sfreq,
+        band_filters=None,
+        method="fir"
+    )
+
+    output = layer(sample_input)
+    assert output.shape == (
+    sample_input.shape[0], layer.n_bands, n_chans, sample_input.shape[2]), \
+        "Output shape is incorrect for FIR filtering."
```
tomMoral, line 741: Usually, it is a good convention to start with the simple tests at the top of the file so I would move these tests to the top.

Also, you can merge them:
```suggestion
@pytest.mark.parametrize('method', ['iir', 'fir'])
def test_forward_pass_filter_bank(method, sample_input):
    """Test the forward pass of the FilterBankLayer with IIR filtering."""
    n_chans = 8
    sfreq = 100
    layer = FilterBankLayer(
        n_chans=n_chans,
        sfreq=sfreq,
        band_filters=None,
        method=method,
    )

    output = layer(sample_input)
    assert output.shape == (
        sample_input.shape[0], layer.n_bands, n_chans, sample_input.shape[2]
    ), f"Output shape is incorrect for {method} filtering."
```

test/unit_tests/models/test_modules.py
```diff
@@ -305,3 +329,413 @@ def test_safelog_extra_repr(eps, expected_repr):
...
+            sfreq=sfreq,
+            band_filters=None,
+            method="iir",
+            iir_params=iir_params
+        )
+
+    assert layer.filts is not None, "Filters should be initialized."
+    assert layer.filts["band_0"][
+               "a"].dtype == torch.float64, "Filter coefficients should be float64."
+
```
tomMoral, line 706: Move these before complex test with scipy.

test/unit_tests/models/test_modules.py
```diff
@@ -305,3 +329,413 @@ def test_safelog_extra_repr(eps, expected_repr):
...
+    # Assertions to verify band_filters and number of bands
+    assert layer.band_filters == expected_band_filters, "Default band_filters not set correctly when band_filters=None."
+    assert layer.n_bands == 9, "Number of bands should be 9 when band_filters=None."
+
+    # Forward pass to ensure output shape is correct
+    output = layer(sample_input)
+    assert output.shape == (sample_input.shape[0], layer.n_bands, n_chans,
+                            sample_input.shape[
+                                2]), "Output shape is incorrect when band_filters=None."
```
tomMoral, line 672:
```suggestion
    assert output.shape == (
        sample_input.shape[0], layer.n_bands, n_chans,
        sample_input.shape[2]
    ), "Output shape is incorrect when band_filters=None."
```

tomMoral (approved):
Here is a second pass on the review.
Nice for the test.

Once these comments are addressed, this LGTM.

bruAristimunha:
Thank you so much for all the revision @qbarthelemy, @sylvchev and @tomMoral.

I feel like I'm learning a lot from you! I loved! Thanks again! I particularly enjoyed making this layer, because it delivered good results and there was no other compatible implementation available.

sylvchev:
You did a very good job and this is a solid foundation that could be reused in many approaches. Kudos Bruno!

PR #659, opened 2024-10-02
[EHN] Including MSVTNet

bruAristimunha:


braindecode/models/msvtnet.py
```diff
@@ -0,0 +1,372 @@
...
+
+    .. figure:: https://raw.githubusercontent.com/SheepTAO/MSVTNet/refs/heads/main/MSVTNet_Arch.png
+       :align: center
+       :alt: MSVTNet Architecture
+
+    Parameters
+    ----------
+    n_filters_list : List[int], optional
+        List of filter numbers for each TSConv block, by default [9, 9, 9, 9].
+    conv1_kernel_sizes : List[int], optional
```
sylvchev, line 211:
```suggestion
    conv1_kernels_size : List[int], optional
```
Nice to add `s` to mark plural and to indicate it is a list of kernel size. I think it is better to put the `s` on kernel.

sylvchev (approved):

bruAristimunha:
thank you @sylvchev

PR #581, opened 2024-03-28
[ENH] Add new model: EEG-SimpleConv

elouayas:
Adding the EEG-SimpleConv net.

```
  EEG-SimpleConv starts with a 1D convolutional layer, where each EEG channel
    enters a separate 1D convolutional channel. This is followed by a series of
    blocks of two 1D convolutional layers. Between the two convolutional layers
    of each block is a max pooling layer, which downsamples the data by a factor
    of 2. Each convolution is followed by a batch normalisation layer and a ReLU
    activation function. Finally, a global average pooling (in the time domain)
    is performed to obtain a single value per feature map, which is then fed
    into a linear layer to obtain the final classification prediction output.
```

PR #661, opened 2024-10-04
EHN: Including pytorch audio

bruAristimunha:


agramfort:
Have you checked it gives equivalent results ?

bruAristimunha:
hey @agramfort!

Yes! The code was exactly the same, as I introduce here: #581. At the time, it didn't make sense to include torch audio as a dependency since it only involved these three functions. The problem was that with FilterBankLayer #656, it would be necessary to include more than a thousand lines (lfilter, filtfilt, fftconvolute, and many private functions).

After thinking about it, it would be better to include torch audio as a dependency, considering that other new models will also use torchaudio functions.

PR #663, opened 2024-10-05
[EHN] Including LMDA

bruAristimunha:
A model that mixes ShallowNet and EEGNet, it was published in NeuroImage. I started this adaptation in February, but I had forgotten to open the PR here.


bruAristimunha:
Small benchmark results with cross-session:

image

bruAristimunha:
Cross-Subject:

image

image

braindecode/models/lmda.py
```diff
@@ -0,0 +1,278 @@
...
+        torch.Tensor
+            Output tensor after applying depth-wise attention.
+
+        """
+        x_pool = self.adaptive_pool(x)
+        x_transpose = x_pool.transpose(-2, -3)
+        y = self.conv(x_transpose)
+        y = self.softmax(y)
+        y = y.transpose(-2, -3)
+        return y * self.num_channels * x
```
martinwimpff, line 67:
```suggestion
        x_pool = self.adaptive_pool(x)  # D' x 1 x num_times
        x_transpose = x_pool.transpose(-2, -3)  # 1 x D' x num_times
        y = self.conv(x_transpose)  # 1 x D' x num_times
        y = self.softmax(y)  # 1 x D' x num_times
        y = y.transpose(-2, -3)  # D' x 1 x num_times
        return y * self.num_channels * x  # D' x num_channels x num_times
```

braindecode/models/lmda.py
```diff
@@ -0,0 +1,278 @@
...
+            Input tensor of shape `(batch_size, n_chans, time)`.
+
+        Returns
+        -------
+        torch.Tensor
+            Output logits of shape `(batch_size, n_outputs)`.
+
+        """
+        x = self.ensure_dim(x)
+        x = torch.einsum("bdcw, hdc->bhcw", x, self.channel_weight)  #
```
martinwimpff, line 270:
```suggestion
        x = self.channel_weight(x)
```
With the convolution layer above, we would simplify to this expression.

martinwimpff (commented):
I haven't read the paper but just from my experience, three aspects are weird:
- spatial layer (channel_weight) before temporal conv 
- two spatial layers: channel_weight and spatial_conv
- depthwise separable convolutions are in the "wrong" order (starting with pw-conv). However, the ordering might still make sense if viewed differently, i.e., the pw-conv is effectively part of the block before. Still, the pw-conv of the spatial_conv would be unnecessary in this case.

PR #666, opened 2024-10-07
[ENH] Including CTNet into braindecode

bruAristimunha:
This model is very interesting! It is a natural evolution of the EEGConformer, published in Scientific Rep. I finished the adaptation for braindecode, but the review by the authors needs to be done.

@eeyhsong and @snailpt  can you review?


bruAristimunha:
Hey @snailpt,

Great that the re-implementation is correct! Thank you so much for your time.

When this PR is integrated, your model will be integrated into the braindecode source; you can use [installing braindecode from the source](https://braindecode.org/stable/install/install_source.html#installing-braindecode-from-the-source), and in our next release, which will be coming soon, you will be able to use it with `pip install braindecode.`

About braindecode.

We are particularly interested in your view of being a new user. We are currently expanding the zoo model, but we have several features, such as data augmentation, self-supervised, and an IO well integrated with mne, for example.

Your contribution as someone new in Github can be either reviewing the documentation, opening an issue indicating things that are not clear, or even integrating things that you think are necessary, such as a new data augmentation, and interpretability.
In my experience, things that can help your research usually lead to more interesting new features.

I sincerely appreciate any time you can dedicate to the library! I think together we can go further!

bruAristimunha:
Missing adjust in the api.rst file

bruAristimunha:
Executing some cross-session training in the competition dataset, I have these results:

image
image

I used moabb, so the pre-processing is different, and there is no data augmentation.

I will test the impact of the position encoding, I remember that for the EEG conformer it does not contribute much

braindecode/models/ctnet.py
```diff
@@ -0,0 +1,411 @@
...
+    """CTNet from Zhao, W et al (2024) [ctnet]_.
+
+    A Convolutional Transformer Network for EEG-Based Motor Imagery Classification
+
+    .. figure:: https://raw.githubusercontent.com/snailpt/CTNet/main/architecture.png
+       :align: center
+       :alt: CTNet Architecture
+
+
+    TODO: fill here..
```
bruAristimunha, line 35: Can you fill here @snailpt?

You will need to suggest.

braindecode/models/ctnet.py
```diff
@@ -0,0 +1,411 @@
...
+    drop_prob_posi : float, default=0.1
+        Dropout probability for the positional encoding.
+    drop_prob_final : float, default=0.5
+        Dropout probability for the final layers.
+
+
+    Notes
+    -----
+    This implementation is adapted from the original CTNet source code
+    [ctnetcode]_ to comply with Braindecode's model standards. More comments.
```
bruAristimunha, line 69:
```suggestion
    [ctnetcode]_ to comply with Braindecode's model standards. More comments.
```

and here too

docs/whats_new.rst
```diff
@@ -53,6 +53,7 @@ Enhancements
 - Adding :class:`braindecode.models.SyncNet` (:gh:`643` by `Bruno Aristimunha`_ )
 - Creating the `FilterBanklayer` module for new models (:gh:`656` by `Bruno Aristimunha`_ )
 - Including PytorchAudio as dependency and remove copied code (:gh:`661` by `Bruno Aristimunha`_)
+- Adding :class:`braindecode.models.CTNet`  (:gh:`640` by `Bruno Aristimunha`_ )
```
bruAristimunha, line 56:
```suggestion
- Adding :class:`braindecode.models.CTNet`  (:gh:`666` by `Bruno Aristimunha`_ )
```

snailpt:
I have already added the corresponding suggestions and comments in the code. Why is the status of my responses still showing as 'pending' in the conversation?

bruAristimunha:
Hey @snailpt,

GitHub is a bit confusing sometimes, you will need to go to the review page again and submit your review @snailpt. 

I think it's a green button at the top of the page.

snailpt:
Hey @bruAristimunha   This is no longer important. Could you please check if I have filled in everything required from my side?

bruAristimunha:
Unfortunately I can't see it, only the person who started the review can see the draft. 

Here is a tutorial that may help you: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request

snailpt:
I finally found the submit button! If there’s anything else I can assist with, please feel free to let me know.

elouayas (changes requested):
The model's inference works well, and I tested it across various scenarios.

I only noticed a small limitation when the input time signal duration is too long. This issue arises from the way the positional encoding is implemented.

Currently, the positional encoding is defined as a fixed parameter with a specified length, which defaults to 100:

`self.encoding = nn.Parameter(torch.randn(1, length, emb_size))`

During the forward pass, only the relevant portion of this encoding (determined by seq_length) is used by indexing the elements.

If the input signal is too long, the length parameter must be manually increased to satisfy the following condition:

`int(n_times / (pool_size_1 * pool_size_2)) > length`

With the default values in the code:

`int(n_times / (8 * 8)) > 100`

With the default values, it means that if n_times (the signal's time duration) exceeds 6463, the inference will crash.

For BCI applications, 6463 should generally be sufficient, given their typical duration and sampling rates. However, it would be beneficial to make the code more flexible.

To address this, the length could be automatically determined based on the input tensor size during the forward pass by calculating:

`length = int(n_times / (pool_size_1 * pool_size_1))`
This adjustment would improve the model's adaptability to different signal lengths.

snailpt:
@bruAristimunha I have tested on the BCI IV-2a and 2b datasets, and found that not applying band-pass filtering to the raw EEG during CTNet data preprocessing yields better results. And looking forward to comparing the experimental results after using S&R data augmentation.

snailpt:
> Executing some cross-session training in the competition dataset, I have these results:
> 
> image image
> 
> I used moabb, so the pre-processing is different, and there is no data augmentation.
> 
> I will test the impact of the position encoding, I remember that for the EEG conformer it does not contribute much

I have tested on the BCI IV-2a and 2b datasets, and found that not applying band-pass filtering to the raw EEG during CTNet data preprocessing yields better results. And looking forward to comparing the experimental results after using S&R data augmentation.

PR #674, opened 2024-10-15
[MNT] TCN and Hybrid are not longer models 

bruAristimunha:
All models in braindecode work end-to-end, except these two. I pinged @robintibor  and @gemeinl and am proposing to move them to the module since they are not models in the end-to-end definition we have here.

I raised this issue in February and during the code sprint last year, and people have already contacted me to find out how these classes work. Which I also don't know either.


bruAristimunha:
The message here is, if it doesn't work in eegclassifier, it must not have EEGmixinModule

bruAristimunha:
The conversation took place via email

robintibor:
Sorry only seeing now, yeah looks good I think @bruAristimunha

PR #677, opened 2024-10-21
[MNT] replacing pip with uv

bruAristimunha:
Experimentally doing a test to see the difference in job speed.


bruAristimunha:
nevermind

PR #680, opened 2024-10-22
[MNT] Eegnetv4 fix v2

bruAristimunha:


bruAristimunha:
The benchmark table @tomMoral 
image
image
image

bruAristimunha:
Hey @tomMoral,

I did a more in-depth study for EEGNet because the benchmark I initially did was slightly different from the original result. (different evaluation)

In the original article, we have interfered from the barplot that the accuracy is 68% +-[70%-67%] (Within-Subject, mixing two sessions and doing cross-validation with four folds (conceptually wrong for me)) and 38% +- [42%-35%] (inter-subjects).

Using benchopt, I investigated the impact of different band frequencies (one of the key differences following our table):
intra_session
inter_session
inter_subject


Using only the closest frequency range [4Hz, 38Hz], we have better results.
| objective_name   | use_linear_layer=False   | use_linear_layer=True   |   original             |
|:-----------------|:-------------------------|:------------------------|:------------------------|
| inter_sessions   | 44.5+-19.96              | 59.38+-15.63            |          |
| inter_subjects   | 51.15+-6.31              | 49.31+-8.51             | 38% +- [42%-35%]         |
| intra_session    | 47.09+-23.21             | 62.87+-20.77            |              |
| within session   |    -                              |                                  | 68% +-[70%-67%]   |


There are several interesting things about this contrast of different evaluations, more or less filtering and domain generalization, and possibly, because of this, the implementation has not been corrected before. Since linear is better when we filter less, I will keep it in my ideas box to explore in the future, somewhat along the lines of our article in the CNS.

We should integrate this code that adds this parameter, and during the benchmark, we validate whether we leave it as linear as default or not.

braindecode/models/eegnet.py
```diff
@@ -265,26 +264,35 @@ def __init__(
...
+                Rearrange("batch x y z -> batch x z y"),
+            )
 
+            module.add_module("squeeze", Expression(squeeze_final_output))
+        else:
+            module.add_module("flatten", nn.Flatten())
+            module.add_module(
+                "linearconstraint",
+                LinearWithConstraint(
+                    in_features=output_shape[1] * output_shape[3],
```
PierreGtch, line 291:
```suggestion
                    in_features=self.F2 * self.final_conv_length,
```
(For consistency with the convolutional layer version)

Plus, in case `n_out_virtual_chans!=1`, it doesn't work. Is it possible to be in this case?
- If yes, we need `in_features=self.F2 * n_out_virtual_chans  * self.final_conv_length`
- Otherwise, we should replace `n_out_virtual_chans` everywere to simplify the code and put an `assert output_shape[2]==1`

braindecode/models/modules.py
```diff
@@ -355,6 +355,22 @@ def _max_norm(self):
...
+
+    def __init__(self, max_norm=1.0, *args, **kwargs):
+        super(LinearWithConstraint, self).__init__(*args, **kwargs)
+        self.max_norm = max_norm
+
+    def forward(self, x):
+        with torch.no_grad():
+            # Apply max-norm constraint
+            self.weight.data = torch.renorm(
+                self.weight.data, p=2, dim=0, maxnorm=self.max_norm
```
PierreGtch, line 369: I think this is correct but not trivial because there is a double inversion between pytorch and tensorflow!
- In torch, the kernel has shape `(out_features, in_features)` and it's the oposite in tensorflow.
- In tensorflow the `axis` parameter of `max_norm`specifies the dim along which the norm is computed, whereas the `dim` parameter of `torch.renorm` specifies the dim along which the sub-tensors are split

I checked just to be sure and it seem correct:
```python
>>> kernel = tf.constant([[0.0,1,2],[3,4,5]])
>>> tf.keras.constraints.max_norm(max_value=2)(kernel)
<tf.Tensor: shape=(2, 3), dtype=float32, numpy=
array([[0.        , 0.48507127, 0.74278134],
       [2.        , 1.9402851 , 1.8569534 ]], dtype=float32)>
```
(see [original EEGNet implementation](https://github.com/vlawhern/arl-eegmodels/blob/4a512e503198db2010848813ead9afbf8cd54c97/EEGModels.py#L152))
```python
>>> kernel = torch.tensor([[0.0,1,2],[3,4,5]]).T # we transpose because the kernel is transposed in torch
>>> torch.renorm(kernel, p=2, dim=0, maxnorm=2)
tensor([[0.0000, 2.0000],
        [0.4851, 1.9403],
        [0.7428, 1.8570]])
```

braindecode/models/eegnet.py
```diff
@@ -38,32 +43,27 @@ class EEGNetv4(EEGModuleMixin, nn.Sequential):
...
-        Dropout probability after the second conv block and before the last layer.
-
-    Notes
-    -----
-    This implementation is not guaranteed to be correct, has not been checked
-    by original authors, only reimplemented from the paper description.
+        Dropout probability.
+    final_layer_linear : bool, default=False
+        If ``False``, uses a convolution-based classification layer. If ``True``,
+        apply a flattened linear layer as the final classification step.
```
PierreGtch, line 64:
```suggestion
    final_layer_with_constraint : bool, default=False
        If ``False``, uses a convolution-based classification layer. If ``True``,
        apply a flattened linear layer with constraint on the weights norm as the final classification step.
```
The fact that it's linear vs. convolutional is not super important, the main difference is the constraint. So I think the name should reflect that

braindecode/models/eegnet.py
```diff
@@ -89,15 +89,15 @@ def __init__(
...
-        pool1_stride_size: int = 4,
         pool2_kernel_size: int = 8,
-        pool2_stride_size: int = 8,
         conv_spatial_max_norm: int = 1,
         activation: nn.Module = nn.ELU,
         batch_norm_momentum: float = 0.01,
         batch_norm_affine: bool = True,
         batch_norm_eps: float = 1e-3,
         drop_prob: float = 0.25,
+        final_layer_linear: bool = False,
```
PierreGtch, line 99: Should we use the "correct" EEGNet implementation as default instead? 
Or maybe a deprecation warning in case `final_layer_linear=False` to say that this is not the correct implementation and that the default value will change in the future

PierreGtch (changes requested):
Nice job @bruAristimunha !! I'm just not sure about the parameter name and the default. WDYT?

braindecode/models/eegnet.py
bruAristimunha, line 99:
```suggestion
        final_layer_with_constraint: bool = False,
```

braindecode/models/eegnet.py
```diff
@@ -116,6 +116,13 @@ def __init__(
         if final_conv_length == "auto":
             assert self.n_times is not None
 
+        if not final_layer_linear:
```
bruAristimunha, line 119:
```suggestion
        if not final_layer_with_constraint:
```

braindecode/models/eegnet.py
```diff
@@ -116,6 +116,13 @@ def __init__(
...
+            warn(
+                "Parameter 'final_layer_linear=False' is deprecated and will be "
```
bruAristimunha, line 121:
```suggestion
                "Parameter 'final_layer_with_constraint=False' is deprecated and will be "
```

bruAristimunha:
Hi @PierreGtch  and @tomMoral,

I appreciate your review @PierreGtch. I have incorporated everything. I really like the new name, thank you for that. As for changing the name, I would rather not. Let's wait for a release to do that.

sylvchev:
Thanks @bruAristimunha !

PR #682, opened 2024-10-26
[MNT] Including figures in the docstring of all models that I could find

bruAristimunha:
Only:

- ContraWR
- SPaRCNet

It doesn't have neural network figures


braindecode/models/__init__.py
```diff
@@ -23,7 +23,7 @@
 from .tidnet import TIDNet
 from .usleep import USleep
 from .util import get_output_shape, to_dense_prediction_model
-from .modules import TimeDistributed, LogActivation
+from .modules import TimeDistributed, LogActivation, GeneralizedGaussianFilter
```
bruAristimunha, line 26: This is a small fix in the docstring

bruAristimunha:
Since this only affects the documentation and is literally the inclusion of figures inside the docstring, I will apply the merge when everything is green.

PR #687, opened 2025-01-02
Check if self._description is None before set_description.

Visvaria:
The original code does not work when WindowsDataset object is created by **def _create_description**, etc.
The function above creates objects with a NoneType self._description, which causes error when using method **set_description**.


agramfort:
can you update a test? are you using _create_description directly? if so it starts with _ so it's a private function meaning you should not be using directly this function. Can it explain the problem you're facing?

bruAristimunha:
Close for lack of response, feel free to reopen and continue if you want

PR #694, opened 2025-02-17
[FIX] Making the MOABB dataset work with BIDS cache

bruAristimunha:
Code to test:
```python
cache_config = dict(
    use=True,
    path="~/mne_data",
    save_raw=True,
    save_epochs=False,
    save_array=False,
    overwrite_raw=False,
    overwrite_epochs=False,
    overwrite_array=False,
)

dataset = MOABBDataset("AlexMI", subject_ids=[1], dataset_load_kwargs={'cache_config': cache_config})

windows = create_windows_from_events(dataset)
```

### DIFF
--- a/braindecode/datasets/moabb.py
+++ b/braindecode/datasets/moabb.py
@@ -56,8 +56,13 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids=None, dataset_load_kwargs=
 
 
 def _annotations_from_moabb_stim_channel(raw, dataset):
-    # find events from stim channel
-    events = mne.find_events(raw)
+    # find events from the stim channel
+    stim_channels = mne.utils._get_stim_channel(None, raw.info, raise_error=False)
+    if len(stim_channels) > 0:
+        # returns an empty array if none found
+        events = mne.find_events(raw, shortest_event=0, verbose=False)
+    else:
+        events, _ = mne.events_from_annotations(raw, verbose=False)
 
     # get annotations from events
     event_desc = {k: v for v, k in dataset.event_id.items()}
@@ -141,7 +146,7 @@ class MOABBDataset(BaseConcatDataset):
         dataset_load_kwargs: dict[str, Any] | None = None,
     ):
         # soft dependency on moabb
-        from moabb import __version__ as moabb_version
+        from moabb import __version__ as moabb_version  # type: ignore[attr-defined]
 
         if moabb_version == "1.0.0":
             warnings.warn(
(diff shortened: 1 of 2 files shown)


docs/whats_new.rst
```diff
@@ -75,6 +75,7 @@ Bugs
 - Deprecate moabb version 1.0.0 because of incorrect epoching (:gh:`627` by `Pierre Guetschel`_)
 - Fixing tutorial benchmark lazy eager loagin (:gh:`` by `Bruno Aristimunha`_ and `Aphel`_)
 - Improve doc build's time with better caching (:gh:`693` by `Thomas Moreau`_)
+- Fixing the MOABBDataset to work with the cache (:gh:`XXX` by `Bruno Aristimunha`_)
```
bruAristimunha, line 78:
```suggestion
- Fixing the MOABBDataset to work with the cache (:gh:`694` by `Bruno Aristimunha`_)
```

braindecode/datasets/moabb.py
```diff
@@ -56,8 +56,13 @@ def _fetch_and_unpack_moabb_data(dataset, subject_ids=None, dataset_load_kwargs=
...
 def _annotations_from_moabb_stim_channel(raw, dataset):
-    # find events from stim channel
-    events = mne.find_events(raw)
+    # find events from the stim channel
+    stim_channels = mne.utils._get_stim_channel(None, raw.info, raise_error=False)
+    if len(stim_channels) > 0:
+        # returns an empty array if none found
+        events = mne.find_events(raw, shortest_event=0, verbose=False)
+    else:
+        events, _ = mne.events_from_annotations(raw, verbose=False)
```
PierreGtch, line 65:
```suggestion
    events, _ = mne.events_from_annotations(raw, verbose=False)
```
I think we should always get the events using the annotations (with https://github.com/NeuroTechX/moabb/pull/696,  the annotations are always set in raw objects obtained from moabb)

PierreGtch (changes requested):

braindecode/datasets/moabb.py
bruAristimunha: For some strange reason, it does not happen in all the datasets. I can try to run without this and fix this on the moabb side.

bruAristimunha:
Hey Pierre,

It is falling already with the fake dataset, as I need this to run some stuff. For me, we merge this PR and open one issue on moabb to solve this in the future. It is a problem on the moabb side.

PierreGtch (approved):
@bruAristimunha I'm not sure I understand why this PR is now needed; what changed on the MOABB/MNE side? 
But let's merge if you think it is best. 
We should take a deep look at versions compatibility with MOABB before making the next braindecode release

braindecode/preprocessing/windowers.py
```diff
@@ -546,10 +546,20 @@ def _create_windows_from_events(
         # We could also just say we just assume window size=trial size
         # in case not given, without this condition...
         # but then would have to change functions overall
-        # to deal with varying window sizes hmmhmh
-        assert np.all(this_trial_sizes == window_size_samples), (
-            "All trial sizes should be the same if you do not supply a window " "size."
-        )
+        # to deal with varying windevents[:, -1]
+        this_trial_sizes[-1] = this_trial_sizes[-1] - 1
```
bruAristimunha, line 550:
```suggestion
```

bruAristimunha:
I will take some time to merge this pr, don't worry

PR #695, opened 2025-02-19
Distributed samplers for recording and relative positioning sampling

dungscout96:
Adding support for distributed sampling by creating new classes `DistributedRecordingSampler` (which subclass `torch.utils.data.distributed.DistributedSampler`), and `DistributedRelativePositioningSampler` which subclass the recording sampler. The split into two classes for distributed vs non-distributed cases is not ideal but torch's distributed sampler doesn't take another torch.utils.data.sampler as input and only takes a Dataset (reference [source code](https://github.com/pytorch/pytorch/blob/main/torch/utils/data/distributed.py)). But maybe there's a better way to consolidate them.

### DIFF
--- a/braindecode/samplers/__init__.py
+++ b/braindecode/samplers/__init__.py
@@ -1,7 +1,7 @@
 """Classes to sample examples."""
 
 from .base import RecordingSampler, SequenceSampler, BalancedSequenceSampler, DistributedRecordingSampler
-from .ssl import RelativePositioningSampler
+from .ssl import RelativePositioningSampler, DistributedRelativePositioningSampler
 
 __all__ = [
     "RecordingSampler",
@@ -9,4 +9,5 @@ __all__ = [
     "BalancedSequenceSampler",
     "RelativePositioningSampler",
     "DistributedRecordingSampler",
+    "DistributedRelativePositioningSampler",
 ]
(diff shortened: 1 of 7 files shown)


bruAristimunha:
Hey @dungscout96,

Thank you for your contribution! Can you please apply the pre-commit, fill in the what's new file and create some tests? What I expected here is the new sampler matches the results from the original version with tests.

bruAristimunha:
hey @dungscout96, do you need some help?

dungscout96:
I added some test cases and updated the documentation. Let me know if I missed anything.

bruAristimunha:
Looks good @dungscout96! My super small comment it is to replace the print to warning warn. Then I can apply the merge when everything it is green :)

If you have any specific research agenda that you would like to integrate to braindecode, please let's use know ;)

dungscout96:
Awesome, thanks @bruAristimunha! I'm scaling SSL training on a large public EEG dataset from the Child Mind Institute of ~2,600 subjects that has recently been formatted to BIDS ([Shirazi et. al, 2024](https://www.biorxiv.org/content/10.1101/2024.10.03.615261v1)). I just finished setting up a torch lightning framework and will be experimenting with different encoder architectures, SSL tasks, and representation quality metrics. Braindecode has been such an elegant framework to work with EEG sampling and I plan to contribute models and SSL samplers to the codebase as I test them out. Would be happy to collaborate on the research side as well if anyone is interested!

bruAristimunha:
Once you finish your paper, we would be very happy to receive a tutorial with the discoveries you made ;)

dungscout96:
Thanks @bruAristimunha! I'll follow up with you re collaboration. The BIDS dataloader sounds neat! Can you give me some pointers?

PR #701, opened 2025-02-28
Add `BIDSDataset`

PierreGtch:
This PR adds the possibility to load any BIDS dataset.

In addition, it adds the possibility to load degenerate BIDS containing `mne.Epochs` data (such as the cache from MOABB), which are not officially supported.

### DIFF
--- a/braindecode/datasets/bids.py
+++ b/braindecode/datasets/bids.py
@@ -162,7 +162,7 @@ class BIDSDataset(BaseConcatDataset):
         # (argument ignore_json only available in mne-bids>=0.16)
         bids_paths = [
             bids_path for bids_path in bids_paths if bids_path.extension != ".json"
-            ]
+        ]
         all_base_ds = Parallel(n_jobs=self.n_jobs)(
             delayed(self._get_dataset)(bids_path) for bids_path in bids_paths
         )
--- a/braindecode/datasets/bids.py
+++ b/braindecode/datasets/bids.py
@@ -99,11 +99,6 @@ class BIDSDataset(BaseConcatDataset):
         (default), the ``.check`` attribute of the returned
         :class:`mne_bids.BIDSPath` object will be set to ``True`` for paths that
         do conform to BIDS, and to ``False`` for those that don't.
-    ignore_json : bool
-        If ``True``, ignores json files. Defaults to ``False``.
-    ignore_nosub : bool
-        If ``True``, ignores all files that are not of the form ``root/sub-*``.
-        Defaults to ``False``.
     preload : bool
         If True, preload the data. Defaults to False.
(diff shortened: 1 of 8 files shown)


examples/datasets_io/plot_bids_dataset_example.py
```diff
@@ -0,0 +1,37 @@
...
+
+from braindecode.datasets import BIDSDataset
+
+
+###############################################################################
+# First, we download a collection of (fake/empty) BIDS datasets.
+
+# import tempfile
+# data_dir = tempfile.mkdtemp()
+data_dir = Path("~/data_local/openneuro/").expanduser()
```
bruAristimunha, line 24:
```suggestion
data_dir = Path("~/mne_data/openneuro/").expanduser()
```

braindecode/datasets/bids.py
```diff
@@ -0,0 +1,210 @@
...
+            ".set",
+            ".edf",
+            ".bdf",
+            ".EDF",
+            ".snirf",
+            ".cdt",
+            ".mef",
+            ".nwb",
+        ]
+    )
```
bruAristimunha, line 140: Can we get this information from mne_bids?

braindecode/datasets/bids.py
```diff
@@ -0,0 +1,210 @@
...
+    ignore_nosub: bool = False
+    preload: bool = False
+    n_jobs: int = 1
+
+    def __post_init__(self):
+        from mne_bids import __version__ as mne_bids_version
+
+        assert (
+            mne_bids_version >= "0.16"
+        ), f"mne-bids version must be >= 0.16, got {mne_bids_version}"
```
bruAristimunha, line 153: Should be function

bruAristimunha (approved):

bruAristimunha:
Looks good for me!

bruAristimunha:
Just testing and release at moabb side.

braindecode/datasets/bids.py
PierreGtch: no I don't think it is available

braindecode/datasets/bids.py
PierreGtch: Maybe it's not even necessary actually to check this. I already specified this in `pyproject.toml`. For MOABB it's different because one version can lead to a silent bug but here the code itself would just fail because of a missing argument. I will remove these lines

PierreGtch:
@bruAristimunha why do we require mne-bids<0.15 in MOABB?
I need mne-bids>=0.16 here or we would have to implement the `ignore_json` argument by hand. It's easy but redundant. Can we change the dep in moabb?

bruAristimunha (approved):

PR #704, opened 2025-03-04
Small fixes on BIDS datasets

PierreGtch:
Issues I found while using these datasets:
- the ‘suffix’ should be `”epo”`, not the `datatype`
- not useful to have `datatypes=“eeg”` by default
- sometimes the annotations are not set in _epo.fif files but we can still recover the event names using the `events` and `event_id` attributes


bruAristimunha:
Does releasing moabb after merge solve the problem? It is only affected epoch data?

bruAristimunha:
I looked again the file, it all the cache...

PierreGtch:
> Does releasing moabb after merge solve the problem? It is only affected epoch data?

It affects epochs and array epochs

bruAristimunha (approved):

PR #709, opened 2025-03-21
Add type hints to `augmentation/functional.py`

itsaphel:
TODOs:
* A couple of places where mypy/pyright get confused, due to reusing the same variable name with a different type.
* Maybe a couple of places where the type should be tightened?


braindecode/augmentation/functional.py
```diff
@@ -853,15 +906,21 @@ def _rotate_signals(X, rotations, sensors_positions_matrix, spherical=True):
...
 
     if isinstance(angle, (float, int, np.ndarray, list)):
         angle = torch.as_tensor(angle)
+    # TODO: else?
 
     if degrees:
         angle = angle * np.pi / 180
 
+    # TODO: mypy is confused about types here
     device = angle.device
```
itsaphel, line 924: Maybe need some stronger type assertions here

bruAristimunha:
Can you fix the tests @itsaphel?

bruAristimunha:
Good question! hmmmmm, let me think... 

I think it is okay to create some assertions to create the type if it is super necessary. I would recommend building or using some util function that informs the type or something and checks for you if this will make mypy happy. Very similar to the scikit-learn approach.

braindecode/augmentation/functional.py
bruAristimunha: hmmm. okay got it your point.

braindecode/augmentation/functional.py
```diff
@@ -853,15 +906,21 @@ def _rotate_signals(X, rotations, sensors_positions_matrix, spherical=True):
...
+def _make_rotation_matrix(
+    axis: Literal["x", "y", "z"],
+    angle: float | int | np.ndarray | list,
+    degrees: bool = True,
+) -> torch.Tensor:
     assert axis in ["x", "y", "z"], "axis should be either x, y or z."
 
     if isinstance(angle, (float, int, np.ndarray, list)):
         angle = torch.as_tensor(angle)
+    # TODO: else?
```
bruAristimunha, line 918:
```suggestion
    if isinstance(angle, (float, int, np.ndarray, list)):
        angle = torch.as_tensor(angle)
    else:
        raise ValueError("angle should be either a float, int, np.ndarray or list.")
```

braindecode/augmentation/functional.py
bruAristimunha: I made one suggestion for this case, I did not understand in the beginning.

PierreGtch:
> For the tests, can fix. I think it's unhappy with Python 3.10+ syntax that I used, so I can fall back to `typing.Union` (etc)

@itsaphel no need to fall back to the `typing.Union` syntax; you can simply add `from __future__ import annotations` to the imports (see [forum](https://stackoverflow.com/questions/73879925/python-type-hints-conditional-from-future-import-annotations))!

braindecode/augmentation/functional.py
```diff
@@ -853,22 +923,29 @@ def _rotate_signals(X, rotations, sensors_positions_matrix, spherical=True):
         return transformed_X
 
 
-def _make_rotation_matrix(axis, angle, degrees=True):
+def _make_rotation_matrix(
+    axis: Literal["x", "y", "z"],
+    angle: float | int | np.ndarray | list,
```
PierreGtch, line 928:
```suggestion
    angle: float | int | np.ndarray | list | torch.Tensor,
```

PierreGtch (changes requested):
Thanks @itsaphel for this work! Just a recuring comment regarding the random state

itsaphel:
Thanks @PierreGtch - I took the randomstate type from the docstring. Should the docstring change here as well?

PierreGtch:
> Thanks @PierreGtch - I took the randomstate type from the docstring. Should the docstring change here as well?

Yes, the docstring should be updated for consistency, thanks!

> Looks like `np.random.RandomState` is the legacy generator? (according to the doc you linked) though I guess if that's what sklearn uses...

Indeed, sklearn uses the leagacy `RandomState` and passing it the new `Generator` will fail:
```python
from sklearn.utils import check_random_state
import numpy as np
random_state = np.random.default_rng(1) # numpy.random.Generator instance
check_random_state(random_state)
```
```
  File "<stdin>", line 1, in <module>
  File “~/miniforge3/envs/mne-bids-pipeline/lib/python3.12/site-packages/sklearn/utils/validation.py", line 1518, in check_random_state
    raise ValueError(
ValueError: Generator(PCG64) at 0x13A8166C0 cannot be used to seed a numpy.random.RandomState instance
```
So we must type accordingly.

bruAristimunha:
Hey @itsaphel,

I will solve the model part on the #714. Do you need help here?

bruAristimunha:
@copilot, please finish this PR

braindecode/augmentation/functional.py
```diff
@@ -90,7 +99,13 @@ def _new_random_fft_phase_odd(batch_size, c, n, device, random_state):
     )
 
 
-def _new_random_fft_phase_even(batch_size, c, n, device, random_state):
+def _new_random_fft_phase_even(
+    batch_size: int,
+    c: int,
+    n: int,
+    device: torch.device,
+    random_state: int | np.random.Generator | None,
```
bruAristimunha:
```suggestion
    random_state: int | np.random.RandomState | None,
```

braindecode/augmentation/functional.py
```diff
@@ -75,7 +78,13 @@ def sign_flip(X, y):
     return -X, y
 
 
-def _new_random_fft_phase_odd(batch_size, c, n, device, random_state):
+def _new_random_fft_phase_odd(
+    batch_size: int,
+    c: int,
+    n: int,
+    device: torch.device,
+    random_state: int | np.random.Generator | None,
```
bruAristimunha:
```suggestion
    random_state: int | np.random.RandomState | None,
```

braindecode/augmentation/functional.py
```diff
@@ -175,7 +196,9 @@ def ft_surrogate(X, y, phase_noise_magnitude, channel_indep, random_state=None):
     return transformed_X, y
 
 
-def _pick_channels_randomly(X, p_pick, random_state):
+def _pick_channels_randomly(
+    X: torch.Tensor, p_pick: float, random_state: int | np.random.Generator | None
```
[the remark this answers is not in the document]
bruAristimunha:
```suggestion
    X: torch.Tensor, p_pick: float, random_state: int | np.random.RandomState | None
```

braindecode/augmentation/functional.py
```diff
@@ -188,7 +211,12 @@ def _pick_channels_randomly(X, p_pick, random_state):
     return torch.sigmoid(1000 * (unif_samples - p_pick))
 
 
-def channels_dropout(X, y, p_drop, random_state=None):
+def channels_dropout(
+    X: torch.Tensor,
+    y: torch.Tensor,
+    p_drop: float,
+    random_state: int | np.random.Generator | None = None,
```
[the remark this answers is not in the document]
bruAristimunha:
```suggestion
    random_state: int | np.random.RandomState | None = None,
```

braindecode/augmentation/functional.py
```diff
@@ -222,7 +250,9 @@ def channels_dropout(X, y, p_drop, random_state=None):
     return X * mask.unsqueeze(-1), y
 
 
-def _make_permutation_matrix(X, mask, random_state):
+def _make_permutation_matrix(
+    X: torch.Tensor, mask: torch.Tensor, random_state: int | np.random.Generator | None
```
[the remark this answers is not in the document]
bruAristimunha:
```suggestion
    X: torch.Tensor, mask: torch.Tensor, random_state: int | np.random.RandomState | None
```

braindecode/augmentation/functional.py
```diff
@@ -241,7 +271,12 @@ def _make_permutation_matrix(X, mask, random_state):
     return batch_permutations
 
 
-def channels_shuffle(X, y, p_shuffle, random_state=None):
+def channels_shuffle(
+    X: torch.Tensor,
+    y: torch.Tensor,
+    p_shuffle: float,
+    random_state: int | np.random.Generator | None = None,
```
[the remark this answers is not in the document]
bruAristimunha:
```suggestion
    random_state: int | np.random.RandomState | None = None,
```

braindecode/augmentation/functional.py
```diff
@@ -280,7 +315,12 @@ def channels_shuffle(X, y, p_shuffle, random_state=None):
     return torch.matmul(batch_permutations, X), y
 
 
-def gaussian_noise(X, y, std, random_state=None):
+def gaussian_noise(
+    X: torch.Tensor,
+    y: torch.Tensor,
+    std: float,
+    random_state: int | np.random.Generator | None = None,
```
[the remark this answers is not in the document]
bruAristimunha:
```suggestion
    random_state: int | np.random.RandomState | None = None,
```

braindecode/augmentation/transforms.py
```diff
@@ -160,7 +160,7 @@ class ChannelsDropout(Transform):
     ----------
     probability: float
         Float setting the probability of applying the operation.
-    proba_drop: float | None, optional
+    p_drop: float | None, optional
         Float between 0 and 1 setting the probability of dropping each channel.
         Defaults to 0.2.
     random_state: int | numpy.random.Generator, optional
```
bruAristimunha:
```suggestion
    random_state: int | numpy.random.RandomState, optional
```

bruAristimunha (approved):

PR #719, opened 2025-04-11
Include summary.csv in package data and update MANIFEST.in

Samuel-Boehm:
-> models/util.py wants to load models/summary.csv but its not downloaded using pip as it was not included in MANIFEST.in and setup.cfg. 

Now it should be possible to use `pip install git+git@github.com:braindecode/braindecode.git` to install current master branch version again and use it without putting summary.csv there manually.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,5 +2,7 @@ include README.rst
 include LICENSE.txt
 include NOTICE.txt
 
+include braindecode/models/summary.csv
+
 recursive-include docs *.ipynb *.rst conf.py Makefile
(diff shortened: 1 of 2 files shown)


bruAristimunha:
Thank you so much @Samuel-Boehm, I faced this issue yesterday,y and I didn't understand.

PierreGtch:
@bruAristimunha do you think we have the same problem in MOABB with the dataset CSV tables? These tables are used to automatically fill the datasets' docstrings

PierreGtch:
good, thanks @bruAristimunha thanks @lucas-heck !

PR #700, opened 2025-02-26
[FIX] Update preprocess.py for Windows compatibility

arnodelorme:
Ensure Windows compatibility -- otherwise, Windows return a write permission error

PR #784, opened 2025-09-29
fix(preprocess): set prefer option to "threads". 

wurining:
Loky(Backend) requires all tasks to be pickleable. For enhanced compatibility, threads are employed as the default parallelism mechanism for I/O-intensive tasks.


bruAristimunha:
I think it is from the competition. Would it be okay to change this? Another user has already reported this issue.

tomMoral:
Is `_preprocess` not doing any computation at all?

IMO, this seems to be not really a good default as many `_preprocess` task will have to perform some computations. But this is probably something to evaluate (check the example, to see if putting `thread` instead of process don't slow down dataloading.
Given that pytorch dataloaders are using `multiprocesses`, I guess in many cases it makes more sense to use processes for parallelization instead of threads.

What are the typical workflow that are not picklable?

wurining:
Here, I add a default behavior for compatibility, as well as an extra slot for feeding parameters passed by joblib.  @bruAristimunha

braindecode/preprocessing/preprocess.py
```diff
@@ -153,8 +158,12 @@ def preprocess(
 
     parallel_processing = (n_jobs is not None) and (n_jobs != 1)
 
-    job_prefer = "threads" if platform.system() == "Windows" else None
-    list_of_ds = Parallel(n_jobs=n_jobs, prefer=job_prefer)(
+    parallel_params = {} if parallel_kwargs is None else dict(parallel_kwargs)
+    parallel_params.setdefault(
+        "prefer", "threads" if platform.system() == "Windows" else None
+    )
```
tomMoral, line 164: Once again, I would not use this as a default as in most cases, the tasks are picklable and multiprocessing will give you better performance if there is some computation to do.

Can you describe what type of tasks are not picklable?

bruAristimunha: Using threads on Windows was a problem noted by @arnodelorme and the San Diego team a while back, and approved by @agramfort, here the pr: https://github.com/braindecode/braindecode/pull/700

We couldn't find the cause, and we didn't have enough human resources to investigate why some operations weren't pickled. At the time, we suspected that this was happening when we defined custom preprocessing functions, or when we tried to use the lighting itself, which made it very difficult to find the exact root.

I tried to investigate, but after two days, I couldn't make any progress.

bruAristimunha: I opened an issue to address this in the future

wurining:
Sorry, I overlooked a key detail in the error. After more detailed testing, I found that this error occurs when:

1. Using debugpy for debugging
2. Using wandb sweep's agent node to launch code that imports this package

However, the error does NOT occur when launching directly from the command line, e.g., `python [entry_point]`

Since the error is easier to reproduce using debugpy, I launched the program with debugpy and conducted more detailed testing. By setting the `n_jobs` parameter of the preprocess function to `2`, I found that the error is triggered at:
<https://github.com/joblib/joblib/blob/fb0b1012d3bab7cbeec87cab2215a1f9905b7ee3/joblib/externals/loky/process_executor.py#L690>

Here, `result_item` is a `_RemoteTraceback` instance, which provides the error message:

```plaintext
  File "/local/data/conda_envs/eeg2025/lib/python3.11/site-packages/joblib/externals/loky/process_executor.py", line 453, in _process_worker
    call_item = call_queue.get(block=True, timeout=timeout)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/data/conda_envs/eeg2025/lib/python3.11/multiprocessing/queues.py", line 122, in get
    return _ForkingPickler.loads(res)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/local/data/projects/eeg_competition_2025_nclab/braindecode/braindecode/preprocessing/__init__.py", line 1, in <module>
    from .mne_preprocess import (  # type: ignore[attr-defined]
  File "/local/data/projects/eeg_competition_2025_nclab/braindecode/braindecode/preprocessing/mne_preprocess.py", line 11, in <module>
    from braindecode.preprocessing.preprocess import Preprocessor
  File "/local/data/projects/eeg_competition_2025_nclab/braindecode/braindecode/preprocessing/preprocess.py", line 31, in <module>
    from braindecode.datasets.base import (
  File "/local/data/projects/eeg_competition_2025_nclab/braindecode/braindecode/datasets/__init__.py", line 11, in <module>
    from .sleep_physio_challe_18 import SleepPhysionetChallenge2018
  File "/local/data/projects/eeg_competition_2025_nclab/braindecode/braindecode/datasets/sleep_physio_challe_18.py", line 25, in <module>
    from braindecode.preprocessing.preprocess import _preprocess
ImportError: cannot import name '_preprocess' from partially initialized module 'braindecode.preprocessing.preprocess' (most likely due to a circular import) (/local/data/projects/eeg_competition_2025_nclab/braindecode/braindecode/preprocessing/preprocess.py)
```

Clearly, this is an error caused by a circular import. As mentioned in the commit, the fix is to move [`from braindecode.preprocessing.preprocess import _preprocess`](https://github.com/wurining/braindecode/blob/a8868dbcd4cd0d1d23c130036d449a15b98ad37d/braindecode/datasets/sleep_physio_challe_18.py#L25) to after [`if preproc is not None:`](https://github.com/wurining/braindecode/blob/a8868dbcd4cd0d1d23c130036d449a15b98ad37d/braindecode/datasets/sleep_physio_challe_18.py#L409), using lazy loading to avoid the circular import.

This happens because when debugpy (and I believe wandb agent has the same reason) launches the program, the required modules are already loaded in the main process, and then they are loaded again in the subprocess, resulting in a circular import.

- Only 2 references to `_preprocess` were found:
  - [`braindecode/braindecode/preprocessing/preprocess.py`](https://github.com/braindecode/braindecode/blob/b45dafb0bc1971423f1c7721cf1669968de22d36/braindecode/preprocessing/preprocess.py#L158)
  - [`braindecode/braindecode/datasets/sleep_physio_challe_18.py`](https://github.com/braindecode/braindecode/blob/b45dafb0bc1971423f1c7721cf1669968de22d36/braindecode/datasets/sleep_physio_challe_18.py#L25)

Therefore, I believe that changing the import location can solve this problem and avoid compatibility issues.

About the modifications already submitted, in a way that does not affect compatibility, developers can have more control over parallel objects, so I believe the modifications can be retained.

@bruAristimunha @tomMoral 

> Tips:
>
> The issue wasn't initially discovered because when the instance was printed to the console, the `__str__` method was called, and the `__cause__` field of the raised exception object was ignored, causing the critical error information to be missed.
> At <https://github.com/joblib/joblib/blob/fb0b1012d3bab7cbeec87cab2215a1f9905b7ee3/joblib/parallel.py#L884>, the error object is assigned to `self._result`.
> It is subsequently raised directly via `raise self._result` and caught at <https://github.com/joblib/joblib/blob/fb0b1012d3bab7cbeec87cab2215a1f9905b7ee3/joblib/parallel.py#L773>.
> However, the specific error information is actually stored in the `__cause__` field of the `_RemoteTraceback` instance.

braindecode/preprocessing/preprocess.py
tomMoral: I still think this does not make sense to put `thread` as you are putting constraints on all users for a few use case  that fails...

And I don't see any approval, just a question by @agramfort.

tomMoral:
I agree that moving the `_preprocess` to avoid circular import make sense.

For `joblib`, all parameters to pass to `Parallel`  can be set in a context [`parallel_config`](https://joblib.readthedocs.io/en/stable/generated/joblib.parallel_config.html) external to the library, so there is no real need for passing extra parameters. But this is not really harmful to put extra ways to pass arguments if you think it makes it more explicit.

For setting `threads` as default, I think this is a bad choice but that's for the other discussion.

wurining:
> I agree that moving the `_preprocess` to avoid circular import make sense.
> 
> For `joblib`, all parameters to pass to `Parallel` can be set in a context [`parallel_config`](https://joblib.readthedocs.io/en/stable/generated/joblib.parallel_config.html) external to the library, so there is no real need for passing extra parameters. But this is not really harmful to put extra ways to pass arguments if you think it makes it more explicit.
> 
> For setting `threads` as default, I think this is a bad choice but that's for the other discussion.

For sure, about under the Windows env., I didn't test it. I agree set the default backend to process-level parallel, like loky.

bruAristimunha:
Thank you for your contribution @wurining and the quick review @tomMoral, we can discuss this again later. I will possibly do a quick release of braindecode for the final phase of the competition.

FYI @sylvchev and @arnodelorme.

PR #792, opened 2025-10-28
Fix issues and enhance Labram model functionality

bruAristimunha:
Address various bugs and improve the Labram model by allowing pre-loading of weights. Include tests for models and update the documentation to reflect recent changes.

### DIFF
--- a/braindecode/models/labram.py
+++ b/braindecode/models/labram.py
@@ -78,7 +78,7 @@ class Labram(EEGModuleMixin, nn.Module):
         The size of the patch to be used in the patch embedding.
     emb_size : int
         The dimension of the embedding.
-    in_channels : int
+    in_conv_channels : int
         The number of convolutional input channels.
     out_channels : int
         The number of convolutional output channels.
@@ -149,7 +149,7 @@ class Labram(EEGModuleMixin, nn.Module):
         input_window_seconds=None,
         patch_size=200,
         emb_size=200,
-        in_channels=1,
+        in_conv_channels=1,
         out_channels=8,
         n_layers=12,
         att_num_heads=10,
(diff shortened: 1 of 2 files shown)


braindecode/models/labram.py
```diff
@@ -228,7 +242,7 @@ def __init__(
                 _PatchEmbed(
                     n_times=self.n_times,
                     patch_size=patch_size,
-                    in_channels=in_channels,
+                    in_channels=self.n_chans,
                     emb_dim=self.emb_size,
```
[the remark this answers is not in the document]
bruAristimunha: I am not super sure about this point, dear @codex, I think maybe the original implementation adaptation was wrong

braindecode/models/labram.py
PierreGtch: There is a problem with `in_channels`. Later it is ignored when `neural_tokenizer=True`.  Should we have `in_channels==n_chans` except if `neural_tokenizer=True`?

I suggest we just remove the `in_channels` parameter

test/unit_tests/models/test_lbm_models.py
```diff
@@ -0,0 +1,436 @@
...
+
+    # This might not raise immediately but could cause issues
+    # depending on how the model is implemented
+    try:
+        output = model_tokenizer(x)
+        # If it doesn't raise, the shape might be unexpected
+        assert output is not None
+    except RuntimeError:
+        # Expected behavior
+        pass
```
PierreGtch, line 453: Should we add a test for loading the weights? 

It might be a slow test. Maybe there is a way to congigure it such that it is only run sometimes (ex: only when we merge a PR)

bruAristimunha: it is okay, at the end the model weight it is 22 mb

braindecode/models/labram.py
bruAristimunha: about this, I think we needed this convolutional for the VQVAE, at least the author study this parameters if I remember correctly

PierreGtch: Ok I see we should not remove it. 

But the name is confusing I think. What about `in_conv_channels`?

bruAristimunha: renamed @PierreGtch

PierreGtch (approved):
Awesome! Thanks @bruAristimunha

PR #796, opened 2025-10-29
Add eegprep preprocessor

chkothe:
This PR adds a set of preprocessors based on the [eegprep](https://github.com/sccn/eegprep) package to complement the existing MNE preprocessors. The main functionality is the `EEGPrep` `Preprocessor` class, which is an end-to-end pipeline for removing artifacts from EEG with reasonable defaults and not too many knobs that one has to worry about tuning. It's based on the much older [clean_rawdata](https://github.com/sccn/clean_rawdata) pipeline popularized by [EEGLAB](https://eeglab.org/). The reference that this implements is:

Mullen, T. R., Kothe, C. A., Chi, Y. M., Ojeda, A., Kerth, T., Makeig, S., ... & Cauwenberghs, G. (2015). Real-time neuroimaging and cognitive monitoring using wearable dry EEG. *IEEE Transactions on Biomedical Engineering*, 62(11), 2553-2567.

There's a simple example script ([plot_bcic_iv_2a_moabb_trial_eegprep.py](https://github.com/chkothe/braindecode/blob/add-eegprep-preprocessor/examples/model_building/plot_bcic_iv_2a_moabb_trial_eegprep.py)) that shows how this can be dropped into an existing workflow.

Besides that class, the individual processing stages are also made available as separate preprocessors so that users can customize the pipeline down to minute details and omit/insert steps beyond what's exposed by the end-to-end pipeline. 

I've added a battery of unit tests that mainly test the interop framework between `braindecode` and `eegprep`, along with several of the algorithms that are straightforward enough to test. The calling convention into `eegprep` is exactly the same for all algorithms, so the algorithm-specific lines of code on the `braindecode` side mainly just the parameter-passing (which is trivial). In addition, EEGPrep itself has a fairly extensive test battery that covers whether the other underlying algorithms work as expected (by establishing correspondence to the original (ancient) battle-tested MATLAB code). 

There's also a fair amount of documentation in the individual classes and parameters. I haven't yet tried to link it into the `.rst` docs (they generally don't build for me right now).


braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1277 @@
...
+
+
+def _eegprep_missing_error() -> ImportError:
+    """Return error class when EEGPrep is not installed."""
+    return ImportError(
+        "The eegprep package is required to use the "
+        "EEGPrep preprocessor.\n"
+        "Please install it via 'pip install eegprep eeglabio', "
+        "or (if you use uv) 'uv pip install eegprep eeglabio'."
+    )
```
bruAristimunha:
```suggestion
from mne.utils import _soft_import
eegprep = _soft_import(
    "eegprep", "EEG Prep", strict=False
)

HAS_EEGPrep = eegprep is not False
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1277 @@
...
+    try:
+        # noinspection PyUnresolvedReferences
+        from threadpoolctl import threadpool_limits
+
+        # this is done because some processing steps might cause churn otherwise
+        return threadpool_limits(limits=nthreads, user_api="blas")
+    except ImportError:
+        # should not happen since eegprep implicitly pulls it in
+        log.info("threadpoolctl not installed, using default thread limits.")
+        return contextlib.nullcontext()
```
bruAristimunha: I didn't understand why we needed this. Can you clarify the rationale here?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1277 @@
...
+    if not isinstance(EEG["chaninfo"], dict) or not EEG["chaninfo"]:
+        EEG["chaninfo"] = {"nosedir": "+X"}
+
+    # actual conversion
+    proc = eegprep.eeg2mne(EEG)
+
+    # restore the .info['description'] field, if it was present
+    proc.info["description"] = EEG["etc"].get("mne_description", "") or None
+
+    return proc
```
bruAristimunha: these functions should be inside eeg prep itself

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1277 @@
...
+
+    return result
+
+
+def _off_if_none(x, *, switch: Any = -1):
+    """Helper to convert None to 'off' (as expected by some functions in eegprep).
+    Can also accept a second variable, which when None, results in 'off' to be
+    emitted."""
+    return "off" if x is None or switch is None else x
+
```
bruAristimunha, line 38: All the functions above should be inside the eegprep, IHMO

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1277 @@
...
+                # actual operation happens here
+                EEG = self.apply_eeg(EEG)
+
+                if self.force_dtype is not None:
+                    EEG["data"] = EEG["data"].astype(self.force_dtype)
+
+            finally:
+                # clear again so we don't keep a reference around
+                self._raw = None
```
bruAristimunha: I think this is will be a little complicated because we have to use fork, process, and thread, but we can discuss in the second iteration of the PR

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1277 @@
...
+            transition=self.transition,
+            attenuation=self.attenuation,
+            method=self.method,
+            # (no other args)
+        )
+
+        return EEG
+
+
+class Resample(EEGPrepBasePreprocessor):
```
bruAristimunha: This function name will collide with the other function name we have here.

examples/model_building/plot_bcic_iv_2a_eegprep_cleaning.py
```diff
@@ -0,0 +1,383 @@
...
+#
+# .. [2] Jayaram, Vinay, and Alexandre Barachant.
+#        "MOABB: trustworthy algorithm benchmarking for BCIs."
+#        Journal of neural engineering 15.6 (2018): 066011.
+#
+# .. [3] Schirrmeister, R.T., Springenberg, J.T., Fiederer, L.D.J., Glasstetter, M.,
+#        Eggensperger, K., Tangermann, M., Hutter, F., Burgard, W. and Ball, T. (2017),
+#        Deep learning with convolutional neural networks for EEG decoding and visualization.
+#        Hum. Brain Mapping, 38: 5391-5420. https://doi.org/10.1002/hbm.23730.
+#
```
bruAristimunha, line 423: You can put your references here too

examples/model_building/plot_bcic_iv_2a_eegprep_cleaning.py
```diff
@@ -0,0 +1,383 @@
...
+# Now we apply preprocessing like bandpass filtering to our dataset.
+# You can either apply functions provided by :class:`mne.io.Raw` or
+# :class:`mne.Epochs` or apply your own functions, either to the
+# MNE object or the underlying numpy array.
+#
+# .. note::
+#    Generally, braindecode prepocessing is directly applied to the loaded
+#    data, and not applied on-the-fly as transformations, such as in
+#    PyTorch-libraries like `<torchvision_>`_.
+#
```
bruAristimunha, line 87: can you write a little more about the eeg prep itself, I think it is necessary here

pyproject.toml
```diff
@@ -65,6 +65,10 @@ documentation = "https://braindecode.org/stable/index.html"
 
 [project.optional-dependencies]
 moabb = ["moabb>=1.2.0"]
+eegprep = [
+    "eegprep>=0.1.0",
+    "eeglabio>=0.1.2",
```
bruAristimunha:
```suggestion
```

eeglabio should be dep of eegprep.

braindecode/preprocessing/eegprep_preprocess.py
bruAristimunha: I think it better to use the _soft_import from mne, will take of the test for all the import stuff

braindecode/preprocessing/eegprep_preprocess.py
bruAristimunha: I got it. I saw it in the code later. This is going to be complicated. Let's discuss this in the next revision. I will ask a parallel expert to shed some light on this. 

We don't need this here because we already depend on Joblib, which can handle this.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Thanks, I've replaced it with a soft import that will raise the following error when a user attempts to use EEGPrep without having installed it:
```pytb
  File "/home/chkothe/repos/ArnoPro/braindecode/braindecode/preprocessing/eegprep_preprocess.py", line 96, -in _mne2eegprep
    EEG = eegprep.mne2eeg(mne)
          ^^^^^^^^^^^^^^^
  File "/home/chkothe/repos/ArnoPro/braindecode/braindecode/preprocessing/eegprep_preprocess.py", line 42, -in __getattr__
    raise ImportError(
ImportError: The eegprep package is required to use the EEGPrep preprocessor.
  Please install braindecode with the [eegprep] extra added as in
  'pip install braindecode[eegprep]' to use this functionality,
  or run 'pip install eegprep[eeglabio]' directly.
```

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Got it, removed that (had escaped me that joblib has been doing the equivalent of that for a while, nice).

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Eliminated the unnecessary utility functions (having them as functions in the first place was a leftover from a variant that predated the common base class).

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Folded that feature into eegprep (0.1.1)

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Removed the use of the `self._raw` mutable state, assuming that's the concern.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Renamed that to Resampling.

examples/model_building/plot_bcic_iv_2a_eegprep_cleaning.py
chkothe: Done

examples/model_building/plot_bcic_iv_2a_eegprep_cleaning.py
chkothe: Added a fair amount of prose to that now; hope that'll do.

pyproject.toml
chkothe: Folded into eegprep as an extra.

chkothe:
Thanks, I think I've applied all the refactors you mentioned. Hope we're in good shape now. Since the doc build failed for me locally I was flying blind there with the tutorial, but maybe you can see whether that passes muster on your end (feel free to suggest corrections).

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        if data:
+            # Still have other keys, update the marker
+            new_marker = _encode_payload(data)
+            new_description = _MARKER_PATTERN.sub(new_marker, description, count=1)
+        else:
+            # No more keys, remove the entire marker
+            new_description = _MARKER_PATTERN.sub("", description, count=1).strip()
+        raw.info["description"] = new_description
+
+    return result
```
chkothe: Btw these two private utilities don't really fit in eegprep, since what they do is store/retrieve custom metadata in an MNE data structure, and eegprep doesn't depend on or interact with MNE (eeglabio does but neither do we maintain that, nor is that a great place for that feature either). 

What this is used for here is for making sure that the data after any of the individual channel removal steps is annotated with the original coordinates (and order) of the removed channels, so that those can later be reinterpolated by the Reinterpolate processor, if the user wants that. This is mainly needed to cater to braindecode's declarative (list-of-classes) preprocessor paradigm where you need to track that type of metadata through the data object (another reason why that functionality is more braindecode specific and not really an eegprep thing, since in an imperative MNE or eegprep preprocessing script, the user would normally just save that into a variable before channel removal and pass it back in later).

That said, the code is a generic metadata store (something that seems sorely needed in the MNE data structure from what I could tell) so theoretically it could also go into braindecode's `util.py`. But maybe it could migrate there if/when a second need for that arises (i.e., someone else's preprocessors need to first annotate and later look up something in the data).

The good thing though is that these are private functions of the module, underscored and not in `__all__`, so people will normally not have to see or think about those.

bruAristimunha: It's okay to create a util in preprocessing folder and move this function. In the top it would be nice to have the __all__ and combine with the soft import

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+        def default(self, obj):
+            # note we don't expect other numpy objects here besides np.array([])
+            if isinstance(obj, np.ndarray) and obj.size == 0:
+                return {_EMPTY_ARRAY_KEY: True}
+            return super().default(obj)
+
+    json_str = json.dumps(data, cls=NumpyEmptyEncoder)
+    encoded = base64.b64encode(json_str.encode("utf-8")).decode("ascii")
+    return f"{_MARKER_START} {encoded} {_MARKER_END}"
```
bruAristimunha: It is strange this for me, an function that define class that define a function. Can you deconvolute and flat this?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+
+def _numpy_empty_decoder(dct):
+    """JSON decoder that reconstructs empty numpy arrays."""
+    if dct.get(_EMPTY_ARRAY_KEY):
+        return np.array([])
+    return dct
+
+
+def _encode_payload(data: dict) -> str:
```
bruAristimunha: Is there no native function on mne that can do this?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+    """
+
+    # payload key under which we store our original channel locations
+    _chanlocs_key = "original_chanlocs"
+
+    def __init__(
+        self,
+        *,
+        can_change_duration: str | bool | None = None,
```
bruAristimunha:
```suggestion
        can_change_duration: str | bool = False,
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            can_change_duration = self.__class__.__name__
+        elif can_change_duration is False:
+            can_change_duration = None
+        self.can_change_duration = can_change_duration
+        self.record_orig_chanlocs = record_orig_chanlocs
+        self.force_dtype = np.dtype(force_dtype) if force_dtype is not None else None
+
+    def apply_eeg(self, eeg: dict[str, Any], raw: BaseRaw) -> dict[str, Any]:
+        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
+        raise NotImplementedError("Subclasses must implement apply_eeg().")
```
bruAristimunha:
```suggestion
    @abstractmethod
    def apply_eeg(self, eeg: dict[str, Any], raw: BaseRaw) -> dict[str, Any]:
        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+    def _apply_op(self, raw: BaseRaw) -> None:
+        """Internal method that does the actual work; this is called by Preprocessor.apply()."""
+        opname = self.__class__.__name__
+        if isinstance(raw, BaseEpochs):
+            raise ValueError(
+                f"{opname} is meant to be used on Raw (continuous) data, "
+                f"not Epochs. Use before epoching."
+            )
```
bruAristimunha: This is complicated in braindecode since 0.8. everything is raw.

 do we have indication that the raw object is used as epoch @PierreGtch ?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+        # preserve the data description for restoration later
+        description = raw.info["description"]
+
+        # split off non-EEG channels since apply_eeg() expects only EEG channels
+        chn_order = raw.ch_names.copy()
+        eeg = raw.copy().pick_types(eeg=True, exclude=[])
```
bruAristimunha: Question here, why not using this? 

https://mne.tools/stable/generated/mne.io.Raw.html#mne.io.Raw.apply_function

The problem I see is that this copy will duplicate the raw object in memory, which, when done in parallel, can cause memory to overflow in large datasets.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        if len(eeg.ch_names) < len(raw.ch_names):
+            non_eeg = raw.drop_channels(eeg.ch_names)
+        else:
+            non_eeg = None
```
bruAristimunha: This whole selection of channel can be performed with function apply I think

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        # remove per-channel DC offset (can be huge)
+        eeg["data"] -= np.median(eeg["data"], axis=1, keepdims=True)
+
+        # optional resampling
+        if (srate := self.resample_to) is not None:
+            eeg = eegprep.resample(eeg, srate)
+
+        # do a check if the data has a supported sampling rate
+        if self.burst_removal_cutoff is not None:
+            from eegprep.utils import round_mat
```
bruAristimunha: Only global imports because of pep8, can you move all the import for the top?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+            supported_rates = (100, 128, 200, 250, 256, 300, 500, 512)
```
bruAristimunha, line 428: Can you put the reference on top of this line?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+_MARKER_END = "-->"
+# Marker key for empty numpy arrays
+_EMPTY_ARRAY_KEY = "__empty_array__"
+
+
+def _numpy_empty_decoder(dct):
+    """JSON decoder that reconstructs empty numpy arrays."""
+    if dct.get(_EMPTY_ARRAY_KEY):
+        return np.array([])
+    return dct
```
bruAristimunha: please move to utils inside the pre-processing

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    cross-session prediction; when using this with a within-session split, there is
+    a risk of data leakage since the artifact removal will be calibrated on statistics
+    of the entire session (and thus test sets). In practice the effect may be minor,
+    unless your downstream analysis is strongly driven by artifacts (e.g., if you
+    are trying to decode eye movements or muscle activity), but paper reviewers may
+    not be convinced by that.
+
+    Parameters
+    ----------
+    (main processing parameters, corresponding to processing stages)
```
bruAristimunha:
```suggestion
```

this will break the documentation generation

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        max_mem_mb: int = 64,
+    ):
+        can_change_duration = " and ".join(
+            opname
+            for opname in (
+                "resample" if resample_to else "",
+                "bad time window removal" if bad_window_max_bad_channels else "",
+            )
+            if opname
+        )
```
bruAristimunha, line 378: I would prefer to be only bool here, or not mix str, bool and None

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+                proc.add_channels([non_eeg], force_update_info=True)
+                if proc.ch_names != chn_order:
+                    proc.reorder_channels(chn_order)
+
+        # write result back into raw, discard proc (_apply_op() is in-place)
+        if not proc.preload:
+            proc.load_data()
+        raw.__dict__ = proc.__dict__
+
+        ...  # can place breakpoint here to inspect raw after processing
```
bruAristimunha:
```suggestion
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        noisy channels. Lower is more aggressive. Typical values are 0.15 (quite
+        aggressive) to 0.3 (quite lax).
+    bad_channel_reinterpolate : bool
+        Whether to reinterpolate bad channels that were detected and removed. Usually
+        required when doing cross-session analysis (to have a consistent channel set).
+    common_avg_ref : bool
+        Whether to apply a common average reference after processing. Recommended
+        when doing cross-study analysis to have a consistent referencing scheme.
+
+    (additional tuning parameters)
```
bruAristimunha:
```suggestion

```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            NumSamples=num_samples,
+            SubsetSize=subset_size,
+            NoLocsChannelCriterion=bad_channel_nolocs_threshold,
+            NoLocsChannelCriterionExcluded=bad_channel_nolocs_exclude_frac,
+            MaxMem=max_mem_mb,
+            # Distance='euclidian',
+            # Channels=None,
+            # Channels_ignore=None,
+            # availableRAM_GB=None,
+            # (no other args)
```
bruAristimunha:
```suggestion
            # Distance='euclidian',
            # Channels=None,
            # Channels_ignore=None,
            # availableRAM_GB=None,
            # (no other args)
```

Can you adjust this, remove it, or make it available? The command lie is not super intuitive for me, the same for # burstRejection on the top

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        self.max_flatline_duration = max_flatline_duration
+        self.max_allowed_jitter = max_allowed_jitter
+
+    def apply_eeg(self, eeg: dict[str, Any], raw: BaseRaw) -> dict[str, Any]:
+        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
+        eeg = eegprep.clean_flatlines(
+            eeg,
+            max_flatline_duration=self.max_flatline_duration,
+            max_allowed_jitter=self.max_allowed_jitter,
+            # (no other args)
```
bruAristimunha:
```suggestion
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    characteristics of the hardware) can have such a large DC offset that typical
+    highpass filters do not fully remove it.
+
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def apply_eeg(self, eeg: dict[str, Any], raw: BaseRaw) -> dict[str, Any]:
+        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
```
bruAristimunha:
```suggestion
        """Apply the preprocessor to an EEGLAB EEG structure."""
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        self.method = method
+
+    def apply_eeg(self, eeg: dict[str, Any], raw: BaseRaw) -> dict[str, Any]:
+        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
+        eeg = eegprep.clean_drifts(
+            eeg,
+            transition=self.transition,
+            attenuation=self.attenuation,
+            method=self.method,
+            # (no other args)
```
bruAristimunha:
```suggestion
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            max_allowed_jitter=self.max_allowed_jitter,
+            # (no other args)
+        )
+
+        return eeg
+
+
+class RemoveDCOffset(EEGPrepBasePreprocessor):
+    """Remove the DC offset from the EEG data by subtracting the per-channel median.
+
```
bruAristimunha, line 529: Can you put the link to the example figure, please?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
+        # note this might as well be implemented directly on the MNE data structure,
+        # but was kept this way since we have the EEGPrep machinery here already.
+        eeg["data"] -= np.median(eeg["data"], axis=1, keepdims=True)
+        return eeg
+
+
+class RemoveDrifts(EEGPrepBasePreprocessor):
+    """Remove drifts from the EEG data using a forward-backward high-pass filter.
+
```
bruAristimunha, line 562: Can you put the link to the example figure, please?

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    Note that MNE has its own suite of filters for this that offers more choices; use
+    this filter if you are specifically interested in matching the EEGLAB and EEGPrep
+    behavior, for example if you're building an EEGPrep-like pipeline from individual
+    steps, e.g., to customize parts that are not exposed by the top-level EEGPrep
+    preprocessor.
+
+    NOTE: If your method involves causal analysis, either with applications to real-time
```
bruAristimunha: The note will not be formatted correctly with the note, needed to follow the rst/pydata theme formation

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            method=self.method,
+            # (no other args)
+        )
+
+        return eeg
+
+
+class Resampling(EEGPrepBasePreprocessor):
+    """Resample the data to a specified rate. Included to equivalence to EEGPrep.
+
```
bruAristimunha, line 627: Can you put the link to the example figure, please? Or just clarify a little more the difference with MNE filtering. Mne Filtering has a big page about filtering, so it would be nice to mention what the difference is with more details for the user.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        """Apply the preprocessor to an EEGLAB EEG structure. Overridden by subclass."""
+        if self.sfreq is not None:
+            eeg = eegprep.resample(eeg, self.sfreq)
+
+        return eeg
+
+
+class RemoveBadChannels(EEGPrepBasePreprocessor):
+    """Removes EEG channels with problematic data; variant that uses channel locations.
+
```
bruAristimunha, line 693: Can you put a figure example link please? it would be also necessary when I create the docstring for the eegprep. 

```suggestion
    .. figure:: https://....
       :align: center
       :alt: Remove Bad Channels
      
      Caption
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            # (no other args)
+        )
+
+        return eeg
+
+
+class RemoveBadChannelsNoLocs(EEGPrepBasePreprocessor):
+    """Remove EEG channels with problematic data; variant that does not use channel
+    locations.
+
```
bruAristimunha, line 804: Figure here

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            # (no other args)
+        )
+
+        return eeg
+
+
+class RemoveBursts(EEGPrepBasePreprocessor):
+    """Run the Artifact Subspace Reconstruction (ASR) method on EEG data to remove
+    burst-type artifacts.
+
```
bruAristimunha, line 889: Figure here

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            maxmem=self.maxmem,
+            # (no other args)
+        )
+
+        return eeg
+
+
+class RemoveBadWindows(EEGPrepBasePreprocessor):
+    """Remove periods with abnormally high-power content from continuous data.
+
```
bruAristimunha, line 1005: Figure here.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        Fractional overlap between consecutive windows (0-1). Higher overlap
+        finds more artefacts but is slower. Default is 0.66 (about 2/3 overlap).
+    max_dropout_fraction : float
+        Maximum fraction of windows that may have arbitrarily low amplitude
+        (e.g. sensor unplugged). Default is 0.1.
+    min_clean_fraction : float
+        Minimum fraction of windows expected to be clean (essentially
+        uncontaminated EEG). Default is 0.25.
+
+    (miscellaneous parameters that normally need not be changed)
```
bruAristimunha:
```suggestion
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+
+class ReinterpolateRemovedChannels(EEGPrepBasePreprocessor):
+    """Reinterpolates EEG channels that were previously removed via one of the EEGPrep
+    channel removal operations, in their original order. This is useful for ensuring
+    a consistent channel set across recordings after automated channel removal; this
+    step is ideally performed after all other artifact removal steps. If no channel
+    locations were recorded, this preprocessor has no effect.
+
+    """
```
bruAristimunha: Format this a little better, put figure, and reference.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+
+    def __init__(self):
+        super().__init__()
+
```
bruAristimunha, line 1154:
```suggestion
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    recordings.
+
+    Note MNE has a very sophisticated rereferencing implementation; this preprocessor
+    is included to simplify building custom EEGPrep-like pipelines.
+
+    """
+
+    def __init__(self):
+        super().__init__()
+
```
bruAristimunha, line 1199:
```suggestion
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            eeg = eegprep.eeg_interp(eeg, orig_chanlocs)
+
+        return eeg
+
+
+class RemoveCommonAverageReference(EEGPrepBasePreprocessor):
+    """Subtracts the common average reference (mean across EEG channels) from the
+    EEG data. This is useful for having a consistent referencing scheme across
+    recordings.
+
```
bruAristimunha: Figure here please

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    Note MNE has a very sophisticated rereferencing implementation; this preprocessor
+    is included to simplify building custom EEGPrep-like pipelines.
```
bruAristimunha: I think it better to explaining a little more the difference

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+        # optionally apply common average reference
+        if self.common_avg_ref:
+            eeg = eegprep.reref(eeg, [])
+
+        return eeg
+
+
+class RemoveFlatChannels(EEGPrepBasePreprocessor):
+    """Removes EEG channels that flat-line for extended periods of time.
+
```
bruAristimunha, line 465: Figure here.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+                raise NotImplementedError(
+                    f"The dataset has an uncommon sampling rate of {sr_current} Hz,"
+                    f" which is not supported by the EEGPrep Preprocessor"
+                    f" implementation. Please enable resampling to"
+                    f" resample the data to one of the supported rates"
+                    f" ({', '.join(str(r) for r in supported_rates)})."
+                )
+
+        # preserve input channel locations for reinterpolation later
+        orig_chanlocs = [cl.copy() for cl in eeg["chanlocs"]]
```
bruAristimunha:
```suggestion
        orig_chanlocs = [channel_loc.copy() for channel_loc in eeg["chanlocs"]]
```

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    pipeline involves the stages:
+     * DC offset subtraction
+     * Optional resampling
+     * Flatline channel detection and removal
+     * High-pass filtering
+     * Bad channel detection and removal using correlation and HF noise criteria
+     * Burst artifact removal using ASR (Artifact Subspace Reconstruction)
+     * Detection and removal of residual bad time windows
+     * Optional reinterpolation of removed channels
+     * Optional common average referencing
```
bruAristimunha: Needed to be better adjusted for when we generate the webpage, but I can do it for you. Just need figures about each operation, and the reference for each operation.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+            new_description = _MARKER_PATTERN.sub(new_marker, description, count=1)
+        else:
+            # No more keys, remove the entire marker
+            new_description = _MARKER_PATTERN.sub("", description, count=1).strip()
+        raw.info["description"] = new_description
+
+    return result
+
+
+class EEGPrepBasePreprocessor(Preprocessor):
```
bruAristimunha, line 40: Here, I think we use a little more of the native function from mne raw to apply, select channels, and reduce by half the amount of code.

braindecode/preprocessing/eegprep_preprocess.py
```diff
@@ -0,0 +1,1139 @@
...
+    class EEGPrepMissing(ModuleType):
+        def __getattr__(self, _: str) -> Any:
+            raise ImportError(
+                "The eegprep package is required to use the EEGPrep preprocessor.\n"
+                "  Please install braindecode with the [eegprep] extra added as in\n"
+                "  'pip install braindecode[eegprep]' to use this functionality,\n"
+                "  or run 'pip install eegprep[eeglabio]' directly."
+            )
+
+    eegprep = EEGPrepMissing("eegprep")
```
bruAristimunha: I would prefer something simpler than this.... but I can do in the last round of the revisions

bruAristimunha (changes requested):
Great, we're getting better and better :)

I think general comments on this review are now more focused on small details and documentation. Since you're having trouble generating the Braindecode documentation, I can adjust the RST file for you, but I need you to include a link for each method to visually explain the process and a reference for each operation.

One very important thing is that for the functions that already exist here with the MNE API, more contrast is necessary, otherwise it will confuse the user.

bruAristimunha:
It is missing what's new file log register and pre-commit too :)

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Moved the nested class out of the function and moved to a utility module.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: We can't use that because the `fn` there will then only see the raw numpy array, but most of the preprocessors need to see a full view of the data object incl. metadata, and may furthermore change the shape of the array (remove or interpolate channels).

However, I changed it so that the temporary copy is only made when there actually are non-EEG channels, and added a note to the tutorial that it's more efficient if users remove non-EEG channels beforehand if they don't want them (which presumably is a common case). It'd be nice if MNE had a `split_channels()` function, and maybe this could be optimized when it eventually gains one.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: I looked and I didn't see anything in MNE that can do the split more efficiently, since ops are by default destructive, though I may have overlooked something.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Eliminated the import

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Which reference did you mean there? On which sampling rates it supports natively?
(that'd be https://github.com/sccn/eegprep/blob/develop/src/eegprep/utils/asr.py#L104)

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Done; I've also made it so that it supports non-empty numpy arrays too and added tests (it'd be weird if it supported numpy types but only if they're empty).

braindecode/preprocessing/eegprep_preprocess.py
chkothe: I changed it to str | bool. The str is for the error message so users know what they need to disable to get the preprocessor to not fail. However, I could require `str` only and users would then need to put the class name or some other human-readable name for their preproc step there. I had that at an earlier point but it felt duplicative, onerous, and copy-paste error prone (when making new preprocessors).

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Clarified that now.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Sorry, I don't have figures for these on hand (let alone in a consistent styling) and won't have time to make 9 of them at this time. In the meantime one could put the reference in the module-level docstring and refer users to that one; it does have a flowchart in it IIRC and also one composite figure that shows the effect of the overall Pipeline.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Fixed it

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Added some more detail there on the differences and whether it matters.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Thanks!

braindecode/preprocessing/eegprep_preprocess.py
chkothe: I assume you're referring to the separation of non-EEG channels? I think we could leave some of that for future takers who want to streamline the code further, max out efficiency, and so forth.

braindecode/preprocessing/eegprep_preprocess.py
chkothe: Simplified it to a minimal try/catch and a single check in the first place where eegprep is needed in the codebase. Note that `_soft_import()` doesn't do much other than setting it to `False` in the except case. It's arguably more customary to set it to `None,` but more importantly, it masks the import from code inspection tools and IDEs etc, so you don't get autocomplete, type checks, and so forth when working with functions/etc in `eegprep`.

chkothe:
Made it through your comments; I've also added the what's new and authors entry as requested. `pre-commit run --files <new files>` passes. 

I think the only thing we don't have at this time is figures. I don't think I can supply them at this time -- you may have to go with a less comprehensive web documentation for this module for the time being. There is one figure that comes to mind that one could associate with the module or the EEGPrep class documentation. You can find it on slide 3 in [this](https://sccn.ucsd.edu/githubwiki/files/asr-final-export.pdf) presentation, attached also for reference. I don't think I still have a higher resolution version of it.

asr-figure

braindecode/preprocessing/eegprep_preprocess.py
bruAristimunha: Reference for each preprocessing, looks a dummy step, but as the documentation is created based on this, it is better if we have the reference.

braindecode/preprocessing/eegprep_preprocess.py
bruAristimunha: Can you please use the claude/copilot to generate? Like, now that it is pre-processing should not be complicated, one good matplot, it is really important visual components, the pages that have visual components are much more visited in braindecode.

bruAristimunha:
@chkothe , I just saw the CI issue, to fix, you need to change this line:

https://github.com/braindecode/braindecode/blob/master/.github%2Fworkflows%2Fdocs.yml#L41

I'll be away from the computer until tomorrow morning, but just switch to all. I'm giving you permission in the repository to activate the trigger to run the CI. 

Once the points are in place, we can apply merge :)

bruAristimunha:
This might help you:

Example reading the plot part:

https://github.com/hubertjb/dl-eeg-review/blob/master/code%2Futils.py#L570-L592

And I like this matplotlib function to do the plot:

https://github.com/hubertjb/dl-eeg-review/blob/master/code%2Fanalysis.py#L1450-L1535

The only thing that would be needed is an EEG with a lot of noise, and to run it before and after each step.

bruAristimunha:
And for the rst, you need to just include the classes names here: https://github.com/braindecode/braindecode/blob/master/docs%2Fapi.rst#L357-L377

bruAristimunha:
hi @chkothe, fix the rst, it is just missing the figures and one reference for each eegprep method

bruAristimunha:
@arnodelorme, do you want to review?

chkothe:
Thanks; I'll see what I can do there. Plotting something meaningful for some of these stages (e.g., re-referencing) etc. will require some creativity there and it's not just 9x the same call to matplotlib. Will let you know.

chkothe:
Btw I saw you pulled the individual stages into the main preprocessing namespace. That seems ok and will enrich the general artifact removal toolkit outside that specific EEGPrep pipeline. Couple things though: 

1. If users encounter these while flipping through the top-level preprocessing namespace (outside their original pipeline context), there's a high risk that they'll apply them in a less-than-ideal order or are missing certain required (precondition) stages entirely. 

2. There will be confusion from having some classes sitting side-by-side that do basically the same thing (just the MNE and the EEGLAB way), i.e., `Resample`/`Resampling`, `SetEEGReference`/`RemoveCommonAverageReference`, and to a lesser extent `Filter`/`RemoveDrifts`.
 
So at minimum we'll need to enhance the per-step docs to clarify pre/post (mainly pre) requirements. E.g., most of them require `RemoveDrifts` having been called beforehand.

 I'd also suggest that we remove `RemoveCommonAverageReference` and `Resampling` from the top level and keep them only in the submodule namespace to avoid the confusion, while users who start out with a use of `EEGPrep` and move from there to wanting to split it up into something finer-grained (but still equivalent) can be directed to where they find the respective classes (the `EEGPrep` class could document that).

One could make a case for keeping `RemoveDrifts` at the top level since it's easier to use correctly than `Filter` for this and it documents some decoding-specific caveats that `Filter` doesn't document (_side note_: I think you'd need a mechanism to inject some braindecode-specific docs into those MNE-inherited filter stages since otherwise you'd have no way to document those things where it matters).

bruAristimunha:
Hmmm, I think I got it, but I am not super sure, maybe small sinc it would be good? I am at discord, aristimunha.

bruAristimunha:
If you don't have a figure/diagram, a simple equation in the docstring is enough for me. 

> Thanks; I'll see what I can do there. Plotting something meaningful for some of these stages (e.g., re-referencing) etc. will require some creativity there and it's not just 9x the same call to matplotlib. Will let you know.

bruAristimunha:
But some figure it would be good!

I like the one you posted, it could be the image from eegprep here.

> If you don't have a figure/diagram, a simple equation in the docstring is enough for me. 
> 
> > Thanks; I'll see what I can do there. Plotting something meaningful for some of these stages (e.g., re-referencing) etc. will require some creativity there and it's not just 9x the same call to matplotlib. Will let you know.
> 
>

chkothe:
Done touching up docstrings. Each preprocessor class now has a citation and either a figure, or an equation.

bruAristimunha (approved):

bruAristimunha:
Many thanks @chkothe

chkothe:
You're welcome; I also fixed the test coverage blocker. 

Re deduplicating the top-level preprocessors (i.e., `braindecode.preprocessing`): I clarified the class docstrings (first sentence) so that it's more obvious that these are alternative strategies for doing the same thing. Perhaps it's fine. 

As far as I'm concerned, we should be good to go now (docs still building but I think they're fine now - edit:confirmed).

bruAristimunha:
Many thanks @chkothe ! We gonna try to do the release for Braindecode very soon to announce all the new feature after competition. I'll potentially send you some details about the promotional artwork for eegprep+ Braindecode.

chkothe:
No problem -- btw. (minor thing) in looking at the generated docs I saw that the classes that had the trivial (no-arguments) constructors, that constructor should probably be put back -- reason is that otherwise these classes will accept arguments of the base class, which aren't necessarily meant to be set by the user (ex.: [RemoveDCOffset](https://braindecode.org/dev/generated/braindecode.preprocessing.RemoveDCOffset.html#braindecode.preprocessing.RemoveDCOffset) shouldn't invite the user to set `can_change_duration`). Arguably one could re-expose/document the `force_dtype` argument, but that could be done in a future enhancement.

PR #806, opened 2025-10-31
Better inheritance between the different dataset classes

PierreGtch:
At the moment, for handling a single file, we have classes 
- `BaseDataset`: handles raw files/objects
- `EEGWindowsDataset`: handles raw files/objects but returns windows, inherits and completely overwrites `BaseDataset`
- and `WindowsDataset`: handles epochs files/objects, inherits and completely overwrites `BaseDataset`

In addition, we have common methods regarding description and transform that are copy/pasted three times.

This is messy and makes type checking fail. Instead, one parent class `RecordDataset` with the common methods, and three children: `RawDataset`, `EEGWindowsDataset` and `WindowsDataset`.

We keep `BaseDataset`, deprecated and identical as `RawDataset`, for backward compatibility

In addition, `BaseConcatDataset` can now be typed with the type of datasets it contains (only raw or only only epochs)

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -666,22 +666,23 @@ class BaseConcatDataset(ConcatDataset, Generic[T]):
                 ds.set_description({key: value_}, overwrite=overwrite)
 
     def save(self, path: str, overwrite: bool = False, offset: int = 0):
-        """Save datasets to files by creating one subdirectory for each dataset:
-        path/
-            0/
-                0-raw.fif | 0-epo.fif
-                description.json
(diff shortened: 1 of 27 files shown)


braindecode/datasets/base.py
```diff
@@ -416,7 +454,10 @@ def set_description(self, description: dict | pd.Series, overwrite: bool = False
         self._description = pd.concat([self.description, description])
 
 
-class BaseConcatDataset(ConcatDataset):
+T = TypeVar("T", bound=RecordDataset)
```
bruAristimunha: Can you move this to the top, and can you make it more descriptive?

Issue #778, opened 2025-09-10
removing the annotation warning


bruAristimunha:
If you know a way to remove the print Used Annotations descriptions: [np.str_('stimulus_anchor')] that is generated by MNE through the Braindecode create_windows_from_events , it will be nice

MatthewChen37:
I believe we can close this issue now based on PR #814 ?

PR #814, opened 2025-11-03
Modify verbosity of create_windows_from_events to clean printed outputs

MatthewChen37:
Currently, braindecode does not control the verbosity of the function call to `mne.events_from_annotations`, leading to inconveniences such as https://github.com/braindecode/braindecode/issues/778. This small change + unit test ensures that the end user can control the verbosity and that it is propagated directly to MNE. 

For example, a common use case was during the (EEG Challenge 2025)[https://eeg2025.github.io/], where the message: 

```
Used Annotations descriptions: [np.str_('stimulus_anchor')]
```

Was printed every time the data was loaded. This can be specifically attributed to (the MNE line in this commit)[https://github.com/mne-tools/mne-python/blob/7cfcc6ba0bd409e2ec2bb76598853288f61d6dc0/mne/annotations.py#L1924]. 

Instead of specifying the verbosity parameter, it used the default, which caused the logger message to print every time. 

To ensure it was properly fixed, I wrote a unit test. Note that (mne's logger set propagate to false)[https://github.com/mne-tools/mne-python/blob/7cfcc6ba0bd409e2ec2bb76598853288f61d6dc0/mne/utils/_logging.py], so the unit test temporarily changes the logger's propagation properties for the sake of testing. 

Outside of those, I noticed that there were some formatting issues (violated flake8) in the files I was working on (often because the lines were too long) so I also decided it would be better to clean the formatting while I was working on the file.

### DIFF
--- a/braindecode/preprocessing/windowers.py
+++ b/braindecode/preprocessing/windowers.py
@@ -11,6 +11,7 @@
 #          Maciej Sliwowski <maciek.sliwowski@gmail.com>
 #          Mohammed Fattouh <mo.fattouh@gmail.com>
 #          Robin Schirrmeister <robintibor@gmail.com>
+#          Matthew Chen <matt.chen42601@gmail.com>
 #
 # License: BSD (3-clause)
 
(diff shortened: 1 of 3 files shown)


bruAristimunha:
this is suuuuper nice!

bruAristimunha:
OMG, super yes, you crushed it with that PR! Many thanks!!!

PR #818, opened 2025-11-04
Add SIENA and CHB_MIT datasets

bruAristimunha:
Include the SIENA dataset and CHB_MIT dataset, updating relevant documentation and file structures to support these additions.


bruAristimunha:
hey @danjjl

can you review and take a look? We can discuss this during the meeting.

braindecode/datasets/chb_mit.py
```diff
@@ -0,0 +1,206 @@
...
+                )
+
+            except PermissionError:
+                raise PermissionError(
+                    f"Please rename {Path(path).parent / unzip_file_name}"
+                    + f"manually to {path} and try again."
+                )
+
+    # Check if the BIDS_Siena folder exists inside the path
+    bids_siena_path = os.path.join(path, "BIDS_CHB-MIT")
```
danjjl: This is supposed to refer the CHB-MIT, I believe the variable name should be corrected.

braindecode/datasets/siena.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class SIENA(BIDSDataset):
+    """The Siena EEG Dataset.
+
+    The database consists of EEG recordings of 14 patients acquired at the Unit of Neurology
+    and Neurophysiology of the University of Siena.
+
+    Subjects include 9 males (ages 25-71) and 5 females (ages 20-58).
+    Subjects were monitored with a Video-EEG with a sampling rate of 512 Hz,
```
danjjl: In the BIDS version of the dataset, which is used here, the sample rate is fixed to 250Hz

jon-dan:
Generally looks fine. Two comments that still need to be resolved.
! I did not download and run the code.

bruAristimunha:
Nice @danjjl!

I was wondering, can you install and check if the pytorch dataset will match your needs?

```bash
pip install "git+https://github.com/braindecode/braindecode.git@siena-dataset"
```

```python
from braindecode.datasets import CHBMIT
dataset = CHBMIT()
```
If you could comment on how you create the window based on the dataset, or point that out in the code where you do it, I can show how to use the create windows dataset, and we can iterate more.

braindecode/datasets/chb_mit.py
```diff
@@ -0,0 +1,206 @@
...
+         SzCORE: seizure community open-source research evaluatio
+         framework for the validation of electroencephalography-based
+         automated seizure detection algorithms. Epilepsia, 66, 14-24.
+    """
+
+    def __init__(self, root=None, *args, **kwargs):
+        # correct the path if needed
+        path_root = root
+        if path_root is not None:
+            list_tsv = glob.glob(f"{path_root}/**/participants.tsv", recursive=True)
```
PierreGtch:
```suggestion
            if not Path(f"{path_root}/participants.tsv").exists():
                raise ValueError("bad root...") 
```
I think it can be confusing and even lead to errors to try to auto-correct the root. For example, if you have a directory with multiple BIDS datasets and so multiple participants.tsv files, you will not necessarily find the correct dataset...

braindecode/datasets/siena.py
```diff
@@ -0,0 +1,198 @@
...
+        SzCORE: seizure community open-source research evaluatio
+        framework for the validation of electroencephalography-based
+        automated seizure detection algorithms. Epilepsia, 66, 14-24.
+    """
+
+    def __init__(self, root=None, *args, **kwargs):
+        # correct the path if needed
+        path_root = root
+        if path_root is not None:
+            list_tsv = glob.glob(f"{path_root}/**/participants.tsv", recursive=True)
```
PierreGtch: same, no correction

braindecode/datasets/chb_mit.py
```diff
@@ -0,0 +1,206 @@
...
+
+        super().__init__(
+            *args,
+            extensions=".edf",
+            check=False,
+            **kwargs,
+        )
+
+
+def _correct_path(path: str):
```
PierreGtch: This could be a staticmethod if the dataset if it only applies to this one. WDYT?

braindecode/datasets/siena.py
```diff
@@ -0,0 +1,198 @@
...
+
+        super().__init__(
+            *args,
+            extensions=".edf",
+            check=False,
+            **kwargs,
+        )
+
+
+def _correct_path(path: str):
```
PierreGtch: same, static method?

braindecode/datasets/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+import os
+from pathlib import Path
+
+
+def correct_dataset_path(
+    path: str, archive_name: str, subfolder_name: str | None = None
+) -> str:
+    """
+    Check if the path is correct and rename the file if needed.
```
PierreGtch: Maybe this should be a private function

Plus the description is not very clear. Could you improve it?

Issue #725, opened 2025-04-17
Improve the initil parameter in the filterbank layer


bruAristimunha:
ok for now but generally not super happy with this one.
Technically the filter_parameters would not be an empty dict but rather the default parameters.

_Originally posted by @martinwimpff in https://github.com/braindecode/braindecode/pull/724#discussion_r2048871516_

MatthewChen37:
Is this issue resolved by #819 now?

bruAristimunha:
Super yes @MatthewChen37, many thanks :)

PR #819, opened 2025-11-05
Fix documentation + interface of FBC models

MatthewChen37:
- Clean up the constructor parameterizations of FBMSNet so they are consistent with FBCNet.
- Added documentation for constructors
- Removed "padding" documentation from FilterBankLayer as it is not used. Note that the pad parameter was added in [MNE v 0.15](https://mne.tools/stable/generated/mne.filter.filter_data.html) to the filter_data function, but it is never called in the filter bank layer. Instead, external modules that call FilterBankLayer, such as FBMSNet and FBCNet, pad after FilterBankLayer's forward pass.
- Also added additional unit tests to `test_models.py` since there were no unit tests for FBMSNet in that file.
    - Note that the [FBMSNet paper](https://ieeexplore.ieee.org/document/9837422) does not provide a calculation of total parameters, so no unit test mirrors the FBCNet's parametrization unit test.
    - Otherwise, in general, the new unit tests mirror FBCNet's unit tests.
- With regards to the issue, https://github.com/braindecode/braindecode/issues/725, I originally wanted the constructor filter parameter arguments to be like: 

```
filter_parameters: dict[Any, Any] | None = {}
```
which would simplify the line:

```
self.filter_parameters = filter_parameters or {} -> self.filter_parameters = filter_parameters
```
But it seems like `mypy` complains about it even though it is valid Python syntax. Therefore, I chose to keep the initialization as:

```
filter_parameters: dict[Any, Any] | None = Mpme
self.filter_parameters = filter_parameters or {} 
```

Also, since I worked on these files, I formatted them because I noticed some weren't following flake8 conventions (line too long, etc.).

### DIFF
--- a/braindecode/models/fbcnet.py
+++ b/braindecode/models/fbcnet.py
@@ -69,7 +69,9 @@ class FBCNet(EEGModuleMixin, nn.Module):
     linear_max_norm : float, default=0.5
         Maximum norm for the final linear layer.
     filter_parameters: dict, default None
-        Parameters for the FilterBankLayer
+        Dictionary of parameters to use for the FilterBankLayer.
+        If None, a default Chebyshev Type II filter with transition bandwidth of
+        2 Hz and stop-band ripple of 30 dB will be used.
 
     References
     ----------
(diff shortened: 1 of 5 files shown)


bruAristimunha:
many thanks @MatthewChen37 🙏🏽

PR #841, opened 2025-11-17
[DOC] updating logo and fixing mismatch in the pyproject

bruaris:
Enhance project ownership by adding new authors and updating institutional affiliations. Revise logo and styling for improved presentation on the homepage. Adjust project configuration for better compatibility.

### DIFF
--- a/docs/_static/style.css
+++ b/docs/_static/style.css
@@ -329,4 +329,6 @@ html[data-theme="dark"] .braindecode-hero__title {
 .braindecode-hero img.logo.mainlogo {
     display: block;
     margin: 0 auto;
+    max-width: 500px;
+    width: 100%;
 }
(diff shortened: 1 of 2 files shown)


bruAristimunha:
@arnodelorme, FYI, I am putting USCD as the support institution

PierreGtch:
Sorry I missed it before merge

It looks a bit strange now, the letters don't always have the same font:
image

I think it would be best to include the letters in the logo as vectorised paths, so that it's not machine-dependant

bruAristimunha:
The tricky part is that when we embed the font within the .svg, it does not always load correctly. But I think I am not super expert in inkspace to export within the font embed

bruAristimunha:
Also, the figure on the left top is not fixed yet:

image

Is this the figure you're talking about?

PierreGtch:
We can convert the font to a path and it will be stable. I will make a quick PR

PR #842, opened 2025-11-17
Vectorize logo text

PierreGtch:
Follow-up of #841


PierreGtch:
Let's check the generated doc before merge

PierreGtch:
All vectorized and consistent now 🤗
image

PR #848, opened 2025-11-18
EEG data augmentations

brunaafl:
I added two EEG data augmentation techniques (Amplitude Scale and Channel Recombination/Rereferencing from Mohsenvand et al. (2020) [1], originally implemented for another project, but that could be useful to Braindecode.

[1] Mohsenvand, M.N., Izadi, M.R., & Maes, P. (2020). Contrastive Representation Learning for Electroencephalogram Classification. PMLR 136:238–253.

### DIFF
--- a/braindecode/augmentation/functional.py
+++ b/braindecode/augmentation/functional.py
@@ -1,6 +1,7 @@
 # Authors: Cédric Rommel <cedric.rommel@inria.fr>
 #          Alexandre Gramfort <alexandre.gramfort@inria.fr>
 #          Gustavo Rodrigues <gustavenrique01@gmail.com>
+#          Bruna Lopes <brunajaflopes@gmail.com>
 #
 # License: BSD (3-clause)
 
@@ -1196,7 +1197,7 @@ def mask_encoding(
     return X, y  # Return the masked tensor and labels
 
 
-def channels_recomb(
+def channels_rereference(
     X: torch.Tensor,
     y: torch.Tensor,
     random_state: int | np.random.RandomState | None = None,
@@ -1227,7 +1228,7 @@ def channels_recomb(
     .. [1] Mohsenvand, M.N., Izadi, M.R. &amp; Maes, P.. (2020). Contrastive
         Representation Learning for Electroencephalogram Classification. Proceedings
         of the Machine Learning for Health NeurIPS Workshop, in Proceedings of Machine
-        Learning Research 136:238-253 Available from https://proceedings.mlr.press/v136/mohsenvand20a.html.
+        Learning Research 136:238-253
 
     """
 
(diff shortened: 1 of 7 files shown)


braindecode/augmentation/transforms.py
```diff
@@ -1271,3 +1273,76 @@ def get_augmentation_params(self, *batch):
             "segment_length": segment_length,
             "n_segments": self.n_segments,
         }
+
```
bruAristimunha, line 1277:
```suggestion


```

bruAristimunha:
pre-commit and whats_new file @brunaafl is lacking

braindecode/augmentation/functional.py
```diff
@@ -1194,3 +1194,100 @@ def mask_encoding(
     X[mask] = 0
 
     return X, y  # Return the masked tensor and labels
+
+
+def channels_recomb(
```
bruAristimunha: Can we call re reference method? I think it be more intuitive

braindecode/augmentation/functional.py
```diff
@@ -1194,3 +1194,100 @@ def mask_encoding(
...
+
+    """
+
+    rng = check_random_state(random_state)
+    batch_size, n_channels, _ = X.shape
+
+    ch = rng.randint(0, n_channels, size=batch_size)
+
+    X_ch = X[torch.arange(batch_size), ch, :]
+    X = X - X_ch.unsqueeze(1)
```
bruAristimunha, line 1241:
```suggestion
    X = X - X_ch.unsqueeze(1)
```
Why do we need the unsqueeze here?

bruAristimunha: It is more an question

braindecode/augmentation/functional.py
```diff
@@ -1194,3 +1194,100 @@ def mask_encoding(
...
+    """
+
+    rng = check_random_state(random_state)
+    batch_size, n_channels, _ = X.shape
+
+    s = torch.from_numpy(
+        rng.uniform(*scale, size=(batch_size, n_channels, 1))
+        ).to(X.dtype).to(X.device)
+
+    X = s*X
```
bruAristimunha: Can you be more descriptive with s and unpack more the operations

bruAristimunha: This isn't very good from a deep learning perspective. Torch<-numpy you will basically swap GPU to cpu memory and losing reference to any gradient.

bruAristimunha: We need to replace this for pure torch

bruAristimunha (approved):
Minor comments Bruna, the rest looks good for me.

The to-do here:
1) expose in the API page the new method,
2) address the comment of the revision,
3) precommit and what's new file

braindecode/augmentation/functional.py
brunaafl: I think so... X is (batch, channel, time) while X_ch is (batch,time), since it represents one channel (the reference one) / trial. So the unsqueeze transforms it into (batch, 1, time)

braindecode/augmentation/functional.py
brunaafl: s is just the samples scaling parameter, sampled for each channel and trial on the batch

bruAristimunha:
Thank you @brunaafl <3

PR #850, opened 2025-11-19
Models configs via pydantic

PierreGtch:
This adds pydantic config classes that are automatically generated for every model!

pydantic is a soft dependancy. @bruAristimunha can you check how I implemented the soft dep? Not sure this is the best solution

Use case:
```python
from braindecode.models.config import EEGNetConfig

cfg = EEGNetConfig(n_chans=16, n_outputs=1, n_times=1000)
cfg  # EEGNetConfig(n_chans=16, n_outputs=1, n_times=1000, final_conv_length='auto', pool_mode='mean', F1=8, D=2, F2=None, kernel_length=64, depthwise_kernel_length=16, pool1_kernel_size=4, pool2_kernel_size=8, conv_spatial_max_norm=1, activation=<class 'torch.nn.modules.activation.ELU'>, batch_norm_momentum=0.01, batch_norm_affine=True, batch_norm_eps=0.001, drop_prob=0.25, final_layer_with_constraint=False, norm_rate=0.25, chs_info=None, input_window_seconds=None, sfreq=None)
```

Models configs benefit from all the nice [pydantic functionalities](https://docs.pydantic.dev/latest/), including easy serialisation for example:
```python
cfg.model_dump()
```
```python
{'n_chans': 16,
 'n_outputs': 1,
 'n_times': 1000,
 'final_conv_length': 'auto',
 'pool_mode': 'mean',
 'F1': 8,
 'D': 2,
 'F2': None,
 'kernel_length': 64,
 'depthwise_kernel_length': 16,
 'pool1_kernel_size': 4,
 'pool2_kernel_size': 8,
 'conv_spatial_max_norm': 1,
 'activation': torch.nn.modules.activation.ELU,
 'batch_norm_momentum': 0.01,
 'batch_norm_affine': True,
 'batch_norm_eps': 0.001,
 'drop_prob': 0.25,
 'final_layer_with_constraint': False,
 'norm_rate': 0.25,
 'chs_info': None,
 'input_window_seconds': None,
 'sfreq': None}
```

Then the model can be instantiated with the `create_instance()` method:
```python
model = cfg.create_instance()
type(model) # braindecode.models.eegnet.EEGNet
```

### DIFF
--- a/braindecode/models/eegconformer.py
+++ b/braindecode/models/eegconformer.py
@@ -411,7 +411,12 @@ class _TransformerEncoder(nn.Sequential):
     """
 
     def __init__(
-        self, num_layers, emb_size, num_heads, att_drop, activation: type[nn.Module] = nn.GELU
+        self,
+        num_layers,
+        emb_size,
+        num_heads,
+        att_drop,
+        activation: type[nn.Module] = nn.GELU,
     ):
         super().__init__(
             *[


PierreGtch:
TODO :
- [ ] include in the doc, but how? I don't think we want to "duplicate" the list of models in the doc. Maybe just a link in the models' docstrings?  @bruAristimunha WDYT?
- [ ] create a tutorial with a fully-configured exca experiment. Other PR

braindecode/models/config.py
```diff
@@ -0,0 +1,217 @@
...
+
+import numpy as np
+from numpydantic import NDArray, Shape
+from typing_extensions import TypedDict
+
+from braindecode.models.base import EEGModuleMixin
+from braindecode.models.util import SigArgName, models_dict, models_mandatory_parameters
+
+
+class ChsInfoType(TypedDict, total=False):
```
bruAristimunha: We have more fields in the chs info:

https://github.com/mne-tools/mne-python/blob/09905e5745b2e5f3c81b442e2ce36558af181c54/mne/_fiff/tests/test_meas_info.py#L928

braindecode/models/config.py
```diff
@@ -0,0 +1,217 @@
...
+    origin = get_origin(type_hint)
+    args = get_args(type_hint)
+    if origin is type or origin is Callable or type_hint is Callable:
+        return pydantic.ImportString
+    if origin is None:
+        return type_hint
+    replaced_args = tuple(_replace_type_hints(arg) for arg in args)
+    if origin is UnionType:
+        origin = Union
+    return origin[replaced_args]
```
bruAristimunha, line 49: i loved

braindecode/models/util.py
```diff
@@ -115,6 +132,129 @@ def _init_models_dict():
     "SignalJEPA",
 ]
 
+################################################################
```
bruAristimunha, line 135: I would prefer that we have some renaming for this logic, and kill the need for the user to fill the table above, but future PR

braindecode/models/config.py
PierreGtch: This is just the `"chs"` key of info. I took all the fields described in the [doc](https://mne.tools/stable/generated/mne.Info.html) (see Notes section)

braindecode/models/util.py
PierreGtch: not easy to implement... At the moment, to detect which arguments are mandatory, we would need to try to instantiate the model multiple times, and each time catch the error (missing argument `n_times`, etc.). 
It would be heavy...

Or you have another idea?

PierreGtch:
@bruAristimunha thanks for pushing the soft import. Regarding this, the configs are located in :
```python
from braindecode.models.config import EEGNetConfig
```
You think it's ok? maybe we could put them directly in :
```python
from braindecode.models import EEGNetConfig
```

bruAristimunha:
Let's keep within config, I would like to iterate with the brain & AI team about this within the Exca tutorial.

PierreGtch:
I will start a PR on the exca tutorial when this is merged

PierreGtch:
@bruAristimunha it looks a bit strange to have the whole file indented :/

bruAristimunha:
thanks @PierreGtch 🙏🏽

PierreGtch:
thanks for your help!

Issue #823, opened 2025-11-05
Temporal generalization Tutorial


bruAristimunha:
Create a tutorial showing the use of generalized time slice windows with braindecode. Basically, it's necessary to train a small model EEGNeX, for example (it could be for BCI, the default model), and apply https://mne.tools/stable/generated/mne.decoding.SlidingEstimator.html to demonstrate Jean Remi King's (@kingjr) logic using BrainDecode.

Once trained, I believe we can utilize Scikit-learn's FrozenEstimator to avoid retraining multiple times.

The general idea is to visualize the curves shown in this tutorial, illustrating activation importance over time.

<img width="640" height="480" alt="Image" src="https://github.com/user-attachments/assets/8acc4b55-4900-4325-bc71-270791794b60" />

This could be considered a core tutorial or an applied one.

You can check this mne tutorial:

https://mne.tools/stable/auto_tutorials/machine-learning/50_decoding.html#temporal-decoding

These two parts are important:
```python
# We will train the classifier on all left visual vs auditory trials on MEG

clf = make_pipeline(StandardScaler(), LogisticRegression(solver="liblinear"))

time_decod = SlidingEstimator(clf, n_jobs=None, scoring="roc_auc", verbose=True)
# here we use cv=3 just for speed
scores = cross_val_multiscore(time_decod, X, y, cv=3, n_jobs=None)

# Mean scores across cross-validation splits
scores = np.mean(scores, axis=0)

# Plot
fig, ax = plt.subplots()
ax.plot(epochs.times, scores, label="score")
ax.axhline(0.5, color="k", linestyle="--", label="chance")
ax.set_xlabel("Times")
ax.set_ylabel("AUC")  # Area Under the Curve
ax.legend()
ax.axvline(0.0, color="k", linestyle="-")
ax.set_title("Sensor space decoding")
```

----
```python
# define the Temporal generalization object
time_gen = GeneralizingEstimator(clf, n_jobs=None, scoring="roc_auc", verbose=True)

# again, cv=3 just for speed
scores = cross_val_multiscore(time_gen, X, y, cv=3, n_jobs=None)

# Mean scores across cross-validation splits
scores = np.mean(scores, axis=0)

# Plot the diagonal (it's exactly the same as the time-by-time decoding above)
fig, ax = plt.subplots()
ax.plot(epochs.times, np.diag(scores), label="score")
ax.axhline(0.5, color="k", linestyle="--", label="chance")
ax.set_xlabel("Times")
[...]

PR #855, opened 2025-11-21
Temporal Generalization Tutorial

MatthewChen37:
This PR adds a temporal generalization tutorial based on https://github.com/braindecode/braindecode/issues/823. Compared to the MNE tutorial, we use a deep learning model (Simple 3-layer MLP) and leverage components from the Braindecode library. 

A rendering of what it would look like is below: 
[TemporalGeneralizationRendering.pdf](https://github.com/user-attachments/files/23684458/TemporalGeneralizationRendering.pdf)

Also, I made a change to the `eegneuralnet.py` implementation because I realized it was overwriting the initialization of the `NeuralNet` skorch module. I don't know if this is intended, but the module's kwargs are overwritten by the signal's shape during initialization, potentially overwriting any user-specified kwargs. So I added logic that made user-specified kwargs the "last word" on what the module's kwargs should look like when initialized.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -43,10 +43,7 @@ Enhancements
 - Adding :class:`braindecode.models.EEGSym` model (:gh:`697` by `Pierre Guetschel`_ and `Bruno Aristimunha`_ )
 - Adding tutorial for :func:`braindecode.preprocessing.windowers.create_fixed_length_windows` (:gh:`746` by `Aniela Bulicz` and `Arina Schippers` and `Aron van Zuijlen`)
 - Including interface with Hugging Face Hub for dataset sharing and loading (:gh:`820` by `Kuntal Kokate`_ and `Bruno Aristimunha`_ )
-<<<<<<< HEAD
 - Added a tutorial for temporal generalization (:gh:`855` by `Matthew Chen`_)
-=======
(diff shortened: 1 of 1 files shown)


MatthewChen37:
@bruAristimunha. Also, I'm aware that we shouldn't merge this immediately as is. There might need to be a couple of revisions, because I am not sure if we should squash the commits since half of the commits in this PR are "merge" commits and also maybe we could also include a more complicated model like an EEGNeX model to compare against the 3-layer MLP.

MatthewChen37:
Build docs seems to be failing because the environment doesn't have the `shap` package: https://shap.readthedocs.io/en/latest/overviews.html. I included as an optional part of the tutorial, and it ran locally because I downloaded it to my local environment. Should we include the `shap` package?

PierreGtch:
Hi @MatthewChen37, thank you for this nice tutorial!
you can add the `shap` dependency in the `pyproject.toml` file, along with the other `docs` dependencies

MatthewChen37:
After editing the `pyproject.toml` file it seems like the tests specifically for macos-latest, 3.11/3.12 are failing for it. Is this expected?

** EDIT ** 

I also think I royally screwed up the git history of this pr, should I just make a new PR?

PR #857, opened 2025-11-22
Implement Temporal Generalization Tutorial

MatthewChen37:
### Copying from previous closed PR: https://github.com/braindecode/braindecode/pull/855

This PR adds a temporal generalization tutorial based on https://github.com/braindecode/braindecode/issues/823. Compared to the MNE tutorial, we use a deep learning model (Simple 3-layer MLP) and leverage components from the Braindecode library.

Also, I made a change to the eegneuralnet.py implementation because I realized it was overwriting the NeuralNet initialization in the skorch module. I don't know if this is intended, but the module's kwargs are overwritten by the signal's shape during initialization, potentially overwriting any user-specified kwargs. So I added logic that made user-specified kwargs the "last word" on what the module's kwargs should look like when initialized.

Side note: Git is hard: 😵😵😵


bruAristimunha:
the issue within the CI is unreleated, is suuuuper hard, but get easy over time ;)

MatthewChen37:
I am now done with the entire tutorial. I just added a section at the end extending the tutorial to the `EEGSimpleConv` model. For some reason, though, building the docs fails because there is no storage left on the device, which is strange—I am not sure which machine this runs on, but I thought it was GitHub’s servers. 

Anyways, please let me know what else I should do before this PR gets merged.

braindecode/eegneuralnet.py
byneuro, line 1: @PierreGtch, can you look this part

.github/workflows/tests.yml
byneuro, line 1: here, i was necessary because of numba

bruAristimunha:
thank you @MatthewChen37 :)

PR #861, opened 2025-11-27
Convert Python examples to Jupyter notebooks for Colab integration

bruAristimunha:
Update configuration to include automatic generation of Jupyter notebooks from Python examples and enhance documentation with Colab buttons for tutorials.

### DIFF
--- a/docs/_templates/sg_launcher_links.html
+++ b/docs/_templates/sg_launcher_links.html
@@ -1,0 +1,29 @@
+{#
+  Custom launcher links for Sphinx-Gallery example pages.
+  This adds an "Open in Colab" button that opens the notebook version
+  of the current example.
+#}
+
+{% if pagename.startswith('auto_examples/') and not pagename.endswith('/index') and pagename != 'auto_examples/index' %}
+
+{#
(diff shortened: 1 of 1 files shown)


PierreGtch:
This is soo cool!! 

Do you think there is a way to also convert the Sphynx formatted cells to markdown?

PierreGtch:
Another thing to take into account: the "open in Colab" link and widget are only compatible with plain GitHub links and Gist links: https://github.com/googlecolab/open_in_colab

Maybe we could have another repo containing only tutorials as notebooks, and the CI would automatically push the new notebook versions, but this seems a bit overkill...

bruAristimunha:
everything is converted! I copied the logic from deep inverse. I saw last week. Basically, we'll have a link to a Jupyter notebook on the webpage that generates our webpage.

I'm just unsure if it will work perfectly because we have a version control system here, but we'll test it and adjust it if necessary.

PierreGtch:
Nice! I argot that the website was deployed on a git repo.

But it's strange, at the moment the link contains version "1.3" instead of "dev"

PR #866, opened 2025-12-08
Add REVE model

jonathanlys01:
This is a first attempt at integrating the REVE model to braindecode.
More documentation will follow once the implementation and usage are validated by the team.

To verify model instantiation and Hugging Face loading, you can run the following (a separate PR will add proper tests if this proves sufficient):

```python
from braindecode.models import REVE
import torch

n_chans = 64
n_times = 1000
batch_size = 8

eeg = torch.randn(batch_size, n_chans, n_times)
pos = torch.randn(batch_size, n_chans, 3)

# Test with random model
model = REVE()
out = model(eeg, pos)

# Load pretrained and use positions
model: REVE = REVE.from_pretrained("brain-bzh/reve-base", cache_dir="./cache")
channel_names = [f"E{i + 1}" for i in range(n_chans)]
pos = model.get_positions(channel_names).expand(eeg.size(0), -1, -1)  # (batch_size, channels, 3)
out = model(eeg, pos)
```

### DIFF
--- a/docs/_static/style.css
+++ b/docs/_static/style.css
@@ -332,3 +332,49 @@ html[data-theme="dark"] .braindecode-hero__title {
     max-width: 500px;
     width: 100%;
 }
+
+/* ********************************************************** Announcement banner */
+
+.bd-header-announcement {
+  background-color: #DAEBF1 !important;  /* Your chosen background */
+  color: #002B5C !important;            /* A strong contrasting text colour */
+  border-left: 5px solid #007CBA !important; /* A refined accent colour */
+}
+
+.bd-header-announcement__content {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: center;
+  gap: 0.6rem 0.75rem;
+  padding: 0.75rem 1rem;
+  font-size: 1.05rem;
+  font-weight: 650;
+  line-height: 1.25;
+}
+
+.bd-header-announcement__content a {
+  color: inherit;
+  text-decoration: underline;
+  text-underline-offset: 3px;
+  text-decoration-thickness: 2px;
+}
+
+.bd-header-announcement__content a.braindecode-announcement-cta {
+  background-color: #007CBA;
+  color: #ffffff !important;
+  text-decoration: none;
+  padding: 0.35rem 0.85rem;
+  border-radius: 999px;
+  font-weight: 800;
+  box-shadow: 0 10px 22px rgba(0, 124, 186, 0.25);
+  transition: background-color 200ms ease, transform 200ms ease;
+}
+
+.bd-header-announcement__content a.braindecode-announcement-cta:hover {
+  background-color: #005a8c;
+  transform: translateY(-1px);
+}
+
+.braindecode-announcement-secondary {
+  font-weight: 550;
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,5 @@
 """Our local Sphinx configuration file."""
+
 #!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 #
@@ -336,6 +337,13 @@ html_theme_options = {
     "footer_start": ["copyright"],
     # 'pygment_light_style': 'default',
     "analytics": dict(google_analytics_id="G-7Q43R82K6D"),
+    "announcement": (
+        "<strong>Using Braindecode in academic work?</strong> "
+        "<a class='braindecode-announcement-cta' href='cite.html'>Cite Braindecode</a> "
+        "<span class='braindecode-announcement-secondary'>"
+        "DOI: <a href='https://doi.org/10.5281/zenodo.16279624'>10.5281/zenodo.16279624</a>"
+        "</span>"
+    ),
 }
 
 # The name of an image file (relative to this directory) to place at the top
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -44,7 +44,7 @@ Enhancements
 - Adding :class:`braindecode.models.EEGSym` model (:gh:`697` by `Pierre Guetschel`_ and `Bruno Aristimunha`_ )
 - Adding tutorial for :func:`braindecode.preprocessing.windowers.create_fixed_length_windows` (:gh:`746` by `Aniela Bulicz` and `Arina Schippers` and `Aron van Zuijlen`)
 - Including interface with Hugging Face Hub for dataset sharing and loading (:gh:`820` by `Kuntal Kokate`_ and `Bruno Aristimunha`_ )
-- Adding a CITATION.cff file, allowing researchers to directly cite the Braindecode library via a Zenodo DOI (:gh:`847` by `Pierre Guetschel`_ )
+- Adding a CITATION.cff file, allowing researchers to directly cite the Braindecode library via a Zenodo DOI (:gh:`847` and :gh:`872` by `Pierre Guetschel`_ )
 - Add automatically-generated pydantic models configs (:gh:`850` by `Pierre Guetschel`_ )
 - Refactor: standardized parameter names in transformer and LabRam modules (:gh:`850` by `José Mauricio`_)
 - Added a tutorial for temporal generalization (:gh:`857` by `Matthew Chen`_)


braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+    print(
+        "flash_attn not found, install it with `pip install flash_attn` if you want to use it"
+    )
+
+
+class REVE(EEGModuleMixin, nn.Module):
+    """
+    REVE (Representation for EEG with versatile embeddings).
+
+    Model described in [elouahidi2025reve]_.
```
bruAristimunha: A big description is necessary here.

Good examples:
* https://braindecode.org/dev/generated/braindecode.models.PBT.html
* https://braindecode.org/dev/generated/braindecode.models.BENDR.html
* https://github.com/braindecode/braindecode/blob/master/braindecode/models/medformer.py

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+
+try:
+    import flash_attn  # type: ignore
+
+    FLASH_AVALIABLE = True
+except ImportError:
+    FLASH_AVALIABLE = False
+    print(
+        "flash_attn not found, install it with `pip install flash_attn` if you want to use it"
+    )
```
bruAristimunha: you can put this as braindecode dependecies

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+        Size of each patch for patch embedding.
+    patch_overlap : int
+        Overlap size between patches.
+
+    References
+    ----------
+    .. [elouahidi2025reve] Yassine El Ouahidi · Jonathan Lys · Philipp Thölke · Nicolas Farrugia · Bastien Pasdeloup · Vincent Gripon · Karim Jerbi · Giulia Lioi (2025).
+       REVE: A Foundation Model for EEG - Adapting to Any Setup with Large-Scale Pretraining on 25,000 Subjects.
+       NeurIPS 2025.
+       Online: `https://arxiv.org/abs/2510.21585`
```
bruAristimunha: We try to follow APA at braindecode, otherwise the formation when you generation the webpage didn't work super well

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+        self.to_patch_embedding = patch_embedding(self.embed_dim, self.patch_size)
+        self.fourier4d = FourierEmb4D(self.embed_dim, freqs=self.freqs)
+        self.mlp4d = mlp_pos_embedding(self.embed_dim)
+        self.ln = nn.LayerNorm(self.embed_dim)
+
+        self.final_layer = nn.Identity()
+
+        self.cls_query_token = nn.Parameter(torch.randn(1, 1, self.embed_dim))
+
+        self._position_bank = None
```
bruAristimunha: The order of appearance of the layers is important to make it more intuitive for the user who is trying to read the code.

Try to minimize the cognitive load as much as possible.

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+            try:
+                from transformers import AutoModel
+
+                self._position_bank = AutoModel.from_pretrained(
+                    "brain-bzh/reve-positions", trust_remote_code=True
+                )
+            except ImportError:
+                raise ImportError(
+                    "Please install transformers to use the REVE position bank: pip install transformers"
+                )
```
bruAristimunha: For the positional encoding, it would be better to put it somewhere else, in a Zenodo repository, or even add an option to not load it and set it as the default. Otherwise, every time someone instantiates the model, you have to go to Hugging Face.

And also, here at braindecode, we are using the Hugging Face library, so pushing and pulling becomes simpler.

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+        """
+        Forward pass of the model.
+        Args:
+            eeg (torch.Tensor): Input EEG tensor of shape (batch_size, channels, sequence_length).
+            pos (torch.Tensor): Position tensor of shape (batch_size, channels, 3) representing (x, y, z) coordinates.
+            return_output (bool, optional): If True, returns the output from the transformer directly.
+                If False, applies the final layer and returns the processed output. Default is False.
+        Returns:
+            Union[torch.Tensor, list[torch.Tensor]]: The output tensor(s) from the model. If `return_output` is True,
+                returns the transformer output; otherwise, returns the output after the final layer.
```
bruAristimunha: we follow numpy convention

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+        """
+
+        eeg = eeg.float()
```
bruAristimunha:
```suggestion
```

you can assume as true

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+        patches = eeg.unfold(
+            dimension=2, size=self.patch_size, step=self.patch_size - self.overlap_size
+        )
+        _b, c, h, _p = patches.shape
```
bruAristimunha:
```suggestion
        _batch, channel, heights, _n_patches = patches.shape
```

bruAristimunha: more descriptive name variables

braindecode/models/reve.py
```diff
@@ -0,0 +1,525 @@
...
+        self.embed_dim = embed_dim
+        self.depth = depth
+        self.heads = heads
+        self.head_dim = head_dim
+        self.mlp_dim_ratio = mlp_dim_ratio
+        self.use_geglu = use_geglu
+        self.freqs = freqs
+        self.noise_ratio = noise_ratio
+        self.patch_size = patch_size
+        self.patch_overlap = patch_overlap
```
bruAristimunha:
```suggestion
```

bruAristimunha: it not needed

braindecode/models/reve.py
jonathanlys01: Working on it, will do a second PR soon with the expanded doc.

braindecode/models/reve.py
jonathanlys01: I'm personally against putting `flash-attn` as a dependency, as it has proven to be quite difficult to install, and because its build is cuda version/torch version dependent. In some cases (I'm thinking about older GPUs like Volta or Tesla generation), it's not even possible to use flash attention.

The current setup works, in case the user doesn't have flash attention installed, the attention backend switches to (in order) flash attention (implemented by torch), efficient attention and MATH. A naive torch backend can also be triggered if people want to extract the attention matrix.

braindecode/models/reve.py
jonathanlys01: A JSON version of the position bank has been pushed to HF at https://huggingface.co/brain-bzh/reve-positions/blob/main/positions.json. It will be loaded by the REVE model, only if positions are required (through a `get_positions` fn). In the future, it is important to make sure that there is a 1-1 correspondance between the position bank as a transformers model and the JSON version as multiple instances of the position data might exist. 

Regarding the use of Hugging Face, I think you are using is `huggingface_hub `. Our original rationale for using the `transformers` library is that you can instantiate our model (and any new version that we push to the model hub with only 2 lines):

```python
from transformers import AutoModel
model = AutoModel.from_pretrained("brain-bzh/reve-base", trust_remote_code=True)
```

As mentioned in the comment on the position bank, this multiplies the number of versions of the same model, and calls for additional integrity tests for the correspondance between the braindecode version and the transformers version.

braindecode/models/reve.py
jonathanlys01: I didn't get this one, you mean in the output format?

braindecode/models/reve.py
jonathanlys01: Not sure, our training pipeline used half precision through the AMP context manager (for faster training and flash attention compatibility), this line tells torch to keep the EEG data in float (i.e. float32).

braindecode/models/reve.py
bruAristimunha: Sorry, one PR for model

braindecode/models/reve.py
bruAristimunha: My point here is that it's not the way to make dependencies optional.

braindecode/models/reve.py
bruAristimunha: It is just .json file. Sorry, I'm not convinced that an extra library is needed for this.

braindecode/models/reve.py
bruAristimunha: Documentation formation

bruAristimunha: Documents formation

braindecode/models/reve.py
bruAristimunha: Like, with some function for hugging library you should be able to load the json.

braindecode/models/reve.py
jonathanlys01: Ended up deleting the flash attention component.

bruAristimunha:
Can you fill the what's new file please and export the model parameters within the utils?

bruAristimunha:
Almost all the comments for the copilot are validated and need to be addressed, except for the test that we already covered with the integration test, which follows the Timm style.

There are some more style details about layer names and how to become fully compatible with braindecode and still be able to load the weights, but I can address them when everything is resolved from the copilot initial review and the tests are green.

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+        )  # Shape: (B, C, num_patches, 4)
+        # Reshape to (B, C * num_patches, 4)
+        pos_with_time = pos_with_time.view(B, C * num_patches, 4)
+
+        return pos_with_time
+
+
+def patch_embedding(embed_dim, patch_size):
+    to_patch_embedding = nn.Sequential(nn.Linear(patch_size, embed_dim))
+    return to_patch_embedding
```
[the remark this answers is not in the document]
jonathanlys01: I tested this, it silently breaks the parameter loading of the patch embedding parameters.
I again highly support the introduction of tests that will check for the equality of the output with the hugging face model and the braindecode version.
 Something like:
```python
from braindecode.models import REVE
import torch
from transformers import AutoModel


batch_size = 2
n_chans = 32
n_times = 1000
n_outputs = 10
cache_dir = "./cache"

ch_list = [f"E{i + 1}" for i in range(n_chans)]
eeg_input = torch.randn(batch_size, n_chans, n_times)

model_braindecode: REVE = REVE.from_pretrained(
    "brain-bzh/reve-base",
    cache_dir=cache_dir,
    n_times=n_times,
    n_chans=n_chans,
    n_outputs=n_outputs,
)
pos_braindecode = model_braindecode.get_positions(ch_list)
pos_braindecode = pos_braindecode.unsqueeze(0).repeat(batch_size, 1, 1)

model_hf = AutoModel.from_pretrained(
    "brain-bzh/reve-base",
    cache_dir=cache_dir,
    trust_remote_code=True,
)
pos_bank_hf = AutoModel.from_pretrained(
    "brain-bzh/reve-positions",
    cache_dir=cache_dir,
    trust_remote_code=True,
)
pos_hf = pos_bank_hf(ch_list)
pos_hf = pos_hf.unsqueeze(0).repeat(batch_size, 1, 1)

# return_output is True to bypass the final layer
output_braindecode = model_braindecode(eeg_input, pos_braindecode, return_output=True)
output_hf = model_hf(eeg_input, pos_hf, return_output=True)

print("positions allclose:", torch.allclose(pos_braindecode, pos_hf, atol=1e-6))
print("outputs allclose:", torch.allclose(output_braindecode[-1], output_hf[-1], atol=1e-6))
```

jonathanlys01:
I agree with all copilot comments about the style, there's just the small issue about the patch embedding layer being a Sequential with a single element, but replacing it with a plain linear silently breaks the loading of the parameters through hugging face (see [this comment](https://github.com/braindecode/braindecode/pull/866#discussion_r2609009597)).

bruAristimunha:
I am supporting of the new test, please let me know if you need any help. It should be quite straightforward to do with Copilot, including transformer as a dependency for the test, and a test following PyTest style. Please let me know if you need any help.

jonathanlys01:
Added the test, you will probably have to add a hugging face token in the GitHub secrets because the repository for REVE is gated.

bruAristimunha:
Hey @jonathanlys01, I tried, and it didn't work. 

Can you find a way to block only the weights? Or maybe host in another repository. 

And btw, we are planning to launch next Friday, or when MNE-bids release. If we don't integrate it by next Friday, it will be included in the next release, possibly at the end of February. I'm working on the promotional materials this next week, canvas, video, and LinkedIn post.

jonathanlys01:
The goal is specifically to ensure the weights integrity, I think it works if you manage to add an HF token as a secret then load it as an env variable and pass it to the from_pretrained method through the `token` argument.

bruAristimunha:
Sorry, I couldn't do it. I haven't found a way to do this without linking my personal Hugging Face permission, which would require an additional key that we would have to maintain. Perhaps there's a way to grant permission within the BrainDecode organization.

I still think it's a **_terrible_** idea to implement this gatekeeping. It is fundamentally against the principle of accessibility to more [FAIR](https://en.wikipedia.org/wiki/FAIR_data) science. This is everything we are historically against in Moabb. However, we are at braindecode, so I don't want to block the merge, so I'm granting you the temporary maintenance permission to add you to the secret.

bruAristimunha:
here is the braindecode org at hugging face: https://huggingface.co/braindecode

jonathanlys01:
Hello @bruAristimunha , I've worked on the integration a bit:
- added the test for the model loading but the secret can't be loaded since it's an external PR, it works locally though.
- added REVE to the summary.csv, let me know if you agree on the labels (currently `Foundation Model` and `Attention/Transformer`.

Let me know if there is anything left to do before merging.

bruAristimunha:
Can you just double check the comments from copilot first revision? If you think it is okay, just resolved.

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+        loc = (
+            positions[..., 0] * p_x
+            + positions[..., 1] * p_y
+            + positions[..., 2] * p_z
+            + positions[..., 3] * p_w
+        ).view(*U, -1)
+        if self.dimension != 512:  # noqa
+            _, _, hd = loc.shape
+            diff = hd - self.dimension // 2
+            loc = loc[:, :, :-diff]
```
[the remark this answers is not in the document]
jonathanlys01: Removed the magic number and implement safer slicing.

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+            The output tensor(s) from the model. If `return_output` is True,
+            returns the transformer output; otherwise, returns the output after the final layer.
+        """
+
+        patches = eeg.unfold(
+            dimension=2,
+            size=self.patch_size,
+            step=self.patch_size - self.overlap_size,
+        )
+        batch_size, channel, heights, _n_patches = patches.shape
```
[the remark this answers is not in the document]
jonathanlys01: Done.

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+class Attention(nn.Module):
+    """
+    Multi-head self-attention layer with RMSNorm.
+    """
+
+    def __init__(self, dim: int, heads: int = 8, head_dim: int = 64):
+        super().__init__()
+        inner_dim = head_dim * heads
+        self.heads = heads
+        self.scale = head_dim**-0.5
```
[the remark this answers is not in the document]
jonathanlys01: Done

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+        ) == 0:
+            n_patches += 1
+
+        flat_dim = self.n_chans * n_patches * self.embed_dim
+        return flat_dim
+
+    def get_positions(self, channel_names: list[str]) -> torch.Tensor:
+        """Fetch channel positions from the position bank.
+
+        Downloads and caches the position bank on first call.
```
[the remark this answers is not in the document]
jonathanlys01: Done.

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+        return_output : bool, optional
+            If True, returns the output from the transformer directly.
+            If False, applies the final layer and returns the processed output. Default is False.
+
+        Returns:
+        -------
+        Union[torch.Tensor, list[torch.Tensor]]
+            The output tensor(s) from the model. If `return_output` is True,
+            returns the transformer output; otherwise, returns the output after the final layer.
+        """
```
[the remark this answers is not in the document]
jonathanlys01: Added to the doc.

braindecode/models/reve.py
```diff
@@ -0,0 +1,568 @@
...
+            mlp_dim=int(self.embed_dim * self.mlp_dim_ratio),
+            geglu=self.use_geglu,
+        )
+
+        final_dim = self._get_flattened_output_dim()
+        self.final_layer = nn.Sequential(
+            nn.Flatten(),
+            nn.LayerNorm(final_dim),
+            nn.Linear(final_dim, self.n_outputs),
+        )
```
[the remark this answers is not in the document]
jonathanlys01: Done

jonathanlys01:
I think we are ready to merge @bruAristimunha !

braindecode/models/reve.py
```diff
@@ -0,0 +1,618 @@
...
+                nn.Linear(final_dim, self.n_outputs),
+            )
+
+        self.cls_query_token = nn.Parameter(torch.randn(1, 1, self.embed_dim))
+
+        self._position_bank = RevePositionBank()
+
+        self.default_pos = None
+        if chs_info is not None:
+            self.default_pos = self.get_positions([ch["ch_name"] for ch in chs_info])
```
PierreGtch: Why not directly using the positions provided in chs_info? Where did the positions come from in your channel bank?
Something like this:
```suggestion
            self.default_pos = self.get_positions([ch["loc"][:3] for ch in chs_info])
```

Or maybe a hybrid solution such that we also cover the case where the channel names are non-informative but positions are provided

PierreGtch:
Super nice to have your model in Braindecode, thanks @jonathanlys01 !!

braindecode/models/reve.py
jonathanlys01: The positions have been manually realigned to be consistent across montages. It's still possible to provide raw positions through the `pos` argument of the forward, but the recommended way to get the 3D coordinates is to use the `get_positions`, that takes a list of channel names as an input.

bruAristimunha:
@codex review please

bruAristimunha:
It's okay for me @jonathanlys01!

However, I made several modifications to the documentation. Can you check the render documentation @jonathanlys01, @glioi, and @elouayas?

To download the documentation locally, you will need to click on this link:

https://github.com/braindecode/braindecode/actions/runs/20288163405/artifacts/4893342966

I'm waiting for your OK to apply the merge.

bruAristimunha:
I made one small change when I was trying to create the redirect link for the doc. Now the PR got moved for this: #876, but @jonathanlys01 still has the writing access. 

Sorry for my git issue.

PR #908, opened 2025-12-29
Add EEGPT

dungscout96:
Add `EEGPT` model to the codebase and adds corresponding unit test that verifies forward pass with 3D input. Pretrained weights can be downloaded with the provided link.

### DIFF
--- a/braindecode/models/eegpt.py
+++ b/braindecode/models/eegpt.py
@@ -32,8 +32,8 @@ class EEGPT(EEGModuleMixin, nn.Module):
         b) Local spatio-temporal embedding maps patches to tokens.
         c) Use of dual self-supervised learning with Spatio-Temporal Representation Alignment and Mask-based Reconstruction.
 
-    **EEGPT** is a 10-million-parameter pretrained transformer model designed for universal EEG
-    feature extraction. It addresses challenges like low SNR and inter-subject variability by employing
+    **EEGPT** is a pretrained transformer model designed for universal EEG feature extraction.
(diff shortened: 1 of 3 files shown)


agramfort:
is there a way to evaluate / compare this model with the REVE model that just landed? As more and more models are added here it would be great to ease the comparisons (eg by making APIs as similar as possible eg to load pretrained weights etc.)

dungscout96:
> is there a way to evaluate / compare this model with the REVE model that just landed? As more and more models are added here it would be great to ease the comparisons (eg by making APIs as similar as possible eg to load pretrained weights etc.)

That’s a great point. We’ve been thinking along similar lines, especially as the number of foundation models in this space continues to grow.

We’re developing a standardized framework to make comparisons across models more fair and reproducible. The current plan is to enforce parity through public BIDS-formatted datasets, consistent preprocessing, and identical adapter architectures (e.g., linear probing, MLPs, LoRA) with a unified hyperparameter sweep protocol. Models would be evaluated alongside established baselines such as EEGNet and EEGConformer.

We're still working on the library, but we plan to interface it with Braindecode so that new model contribution can be evaluated automatically. Once the core pieces are in place, we’d greatly value your thoughts on the API design and any best practices you think would improve cross-model comparisons.

bruAristimunha:
Thanks for the context @dungscout96, I pushed the fixes that I think are necessary. I feel we should focus a little on coverage of this model, but with minor details.

bruAristimunha:
@codex review please

PR #914, opened 2026-01-10
Add CBraMod model

PierreGtch:


braindecode/models/base.py
```diff
@@ -397,9 +397,18 @@ def get_torchinfo_statistics(
         torchinfo.ModelStatistics
             ModelStatistics generated by torchinfo.summary.
         """
+        try:
```
bruAristimunha, line 400: 👀 ?

braindecode/models/util.py
```diff
@@ -132,6 +132,7 @@ def _init_models_dict():
             "chs_info": [{"ch_name": f"E{i + 1}", "kind": "eeg"} for i in range(19)],
         },
     ),
+    ("CBraMod", [], None),
```
bruAristimunha: no parameter? O=

docs/api.rst
```diff
@@ -62,16 +62,16 @@ a common interface for all EEG models and derivate variables names if necessary.
...
+    .. currentmodule:: braindecode.models
+
+    - :class:`BIOT` - Foundation model with pre-trained weights
+    - :class:`CBraMod` - Criss-Cross Transformer model with pre-trained weights
+    - :class:`Labram` - Large Brain Model with pre-trained weights
+    - :class:`REVE` - EEG foundation model with pre-trained weights
+    - :class:`LUNA` - Universal EEG embedding model with pre-trained weights
+    - :class:`BENDR` - Foundation model with pre-trained weights
+    - :class:`SignalJEPA` - Self-supervised learning model with pre-trained weights
+    - :class:`EEGPT` - Pretrained transformer for universal EEG
```
bruAristimunha, line 74: can you revert here?

test/unit_tests/models/test_foundation_models.py
```diff
@@ -974,3 +1013,26 @@ def test_reve_model_outputs_match():
...
+
+# ==============================================================================
+# Tests for CBraMod Model
+# ==============================================================================
+
+
+def test_cbramod_load_weights():
+    model = CBraMod()
+    state_dict = torch.hub.load_state_dict_from_url(
+        "https://huggingface.co/weighting666/CBraMod/resolve/main/pretrained_weights.pth",
```
bruAristimunha: can you rehost the pre-trained weights? The license allows

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+
+    CBraMod has been comprehensively evaluated on **10 downstream BCI tasks across 12 public datasets**
+    sourced from institutions different from the pretraining dataset, demonstrating strong
+    generalization capabilities:
+
+    - Motor imagery (MI) classification
+    - Emotion recognition
+    - Seizure detection
+    - Sleep staging
+    - And others
```
bruAristimunha: can you link the datasets classes here?

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+
+    Consistently outperforms strong baselines including EEGNet, EEGConformer, BIOT, and LaBraM.
```
bruAristimunha:
```suggestion
```

We know the reproducibility issues from this model...

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+    dim_feedforward : int, default=800
+        Dimension of the feedforward network in Transformer layers.
+    n_layer : int, default=12
+        Number of Transformer layers.
+    nhead : int, default=8
+        Number of attention heads.
+    activation : type[nn.Module], default=nn.GELU
+        Activation function used in Transformer feedforward layers.
+    emb_dim : int, default=200
+        Output embedding dimension.
```
bruAristimunha, line 137: missing some parameters here

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+       In The Thirteenth International Conference on Learning Representations (ICLR 2025).
+       https://arxiv.org/abs/2412.07236
+
+    Notes
+    -----
+    - Pretraining uses masked EEG patch reconstruction on TUEG corpus
+    - Model supports arbitrary EEG channel counts through flexible positional encoding
+    - Compatible with various EEG reference schemes
+    - The criss-cross attention mechanism provides interpretability through visualizable
+      attention weights showing spatial and temporal patterns learned by the model
```
bruAristimunha:
```suggestion
```

I think you already talked about this before.

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+
+    def forward(self, x, mask=None):
+        x = self.rearrange(x)
+        patch_emb = self.patch_embedding(x, mask)
+        feats = self.encoder(patch_emb)
+        out = self.proj_out(feats)
+        return out
+
+
+class PatchEmbedding(nn.Module):
```
bruAristimunha:
```suggestion
class _PatchEmbedding(nn.Module):
```

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+                padding=(0, 1),
+            ),
+            nn.GroupNorm(5, 25),
+            nn.GELU(),
+        )
+        self.spectral_proj = nn.Sequential(
+            nn.Linear(101, d_model),
+            nn.Dropout(drop_prob),
+        )
+
```
bruAristimunha, line 284: I think the hardcoded value will make our life hard to fine-tune, I would expose everything.

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+        mask_x = mask_x.contiguous().view(bz * ch_num * patch_num, patch_size)
+        spectral = torch.fft.rfft(mask_x, dim=-1, norm="forward")
+        spectral = torch.abs(spectral).contiguous().view(bz, ch_num, patch_num, 101)
+        spectral_emb = self.spectral_proj(spectral)
+
+        patch_emb = patch_emb + spectral_emb
+
+        positional_embedding = self.positional_encoding(patch_emb.permute(0, 3, 1, 2))
+        positional_embedding = positional_embedding.permute(0, 2, 3, 1)
```
bruAristimunha: Can you replace einops to become something more explicit?

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+            mask_x = x.clone()
+            mask_x[mask == 1] = self.mask_encoding
+
+        mask_x = mask_x.contiguous().view(bz, 1, ch_num * patch_num, patch_size)
+        patch_emb = self.proj_in(mask_x)
+        patch_emb = (
+            patch_emb.permute(0, 2, 1, 3)
+            .contiguous()
+            .view(bz, ch_num, patch_num, self.d_model)
+        )
```
bruAristimunha: Can you replace einops to become something more explicit?

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def _weights_init(m):
+    if isinstance(m, nn.Linear):
+        nn.init.kaiming_normal_(m.weight, mode="fan_out", nonlinearity="relu")
+    elif isinstance(m, nn.Conv1d):
+        nn.init.kaiming_normal_(m.weight, mode="fan_out", nonlinearity="relu")
+    elif isinstance(m, nn.BatchNorm1d):
+        nn.init.constant_(m.weight, 1)
+        nn.init.constant_(m.bias, 0)
```
bruAristimunha: I would put within the model, for make more clear the relation

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+class TransformerEncoder(nn.Module):
```
bruAristimunha:
```suggestion
class _TransformerEncoder(nn.Module):
```

mostly to not need to write documentation about the function

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+        self,
+        encoder_layer,
+        num_layers,
+        norm=None,
+        enable_nested_tensor=True,
+        mask_check=True,
+    ):
+        super().__init__()
+        torch._C._log_api_usage_once(f"torch.nn.modules.{self.__class__.__name__}")
+        self.layers = _get_clones(encoder_layer, num_layers)
```
bruAristimunha: I never saw this command. Can you explain with some comments?

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+        for mod in self.layers:
+            output = mod(output, src_mask=mask)
+        if self.norm is not None:
+            output = self.norm(output)
+        return output
+
+
+class CrissCrossTransformerEncoderLayer(nn.Module):
+    __constants__ = ["norm_first"]
+
```
bruAristimunha: I would put some attention within the documentation, putting the figure from the paper, it is nice representation of the crissCrossAttention.

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+        nhead: int,
+        dim_feedforward: int = 2048,
+        dropout: float = 0.1,
+        activation: type[nn.Module] = nn.GELU,
+        layer_norm_eps: float = 1e-5,
+        batch_first: bool = False,
+        norm_first: bool = False,
+        bias: bool = True,
+        device=None,
+        dtype=None,
```
bruAristimunha:
```suggestion
```

you usually don't expose the device and the type

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+    ) -> None:
+        factory_kwargs = {"device": device, "dtype": dtype}
```
bruAristimunha: same here
```suggestion
```

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+    ) -> Tensor:
+        x = src
+        x = x + self._sa_block(
+            self.norm1(x), src_mask, src_key_padding_mask, is_causal=is_causal
+        )
+        x = x + self._ff_block(self.norm2(x))
+        return x
+
+    # self-attention block
+    def _sa_block(
```
bruAristimunha:
```suggestion
    def _self_attention_block(
```

little more explicit

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+    ) -> Tensor:
+        bz, ch_num, patch_num, patch_size = x.shape
+        xs = x[:, :, :, : patch_size // 2]
+        xt = x[:, :, :, patch_size // 2 :]
+        xs = (
+            xs.transpose(1, 2)
+            .contiguous()
+            .view(bz * patch_num, ch_num, patch_size // 2)
+        )
+        xt = xt.contiguous().view(bz * ch_num, patch_num, patch_size // 2)
```
bruAristimunha: Einops everywhere here

braindecode/models/cbramod.py
bruAristimunha: As this layer is the main contribution of the model, I think it needs to document a little more about this part.

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+
+    # feed forward block
+    def _ff_block(self, x: Tensor) -> Tensor:
+        x = self.linear2(self.dropout(self.activation(self.linear1(x))))
+        return self.dropout2(x)
+
+
+def _get_clones(module, N):
+    # FIXME: copy.deepcopy() is not defined on nn.module
+    return nn.ModuleList([copy.deepcopy(module) for i in range(N)])
```
bruAristimunha:
```suggestion
def _get_clones(module, N):
    # FIXME: copy.deepcopy() is not defined on nn.module
    return nn.ModuleList([copy.deepcopy(module) for i in range(N)])
```

I am sure that pytorch have some way to do this, without this util function

braindecode/models/cbramod.py
```diff
@@ -0,0 +1,439 @@
...
+        super().__init__(
+            n_outputs=n_outputs,
+            n_chans=n_chans,
+            chs_info=chs_info,
+            n_times=n_times,
+            input_window_seconds=input_window_seconds,
+            sfreq=sfreq,
+        )
+        del n_chans, chs_info, n_times, input_window_seconds, sfreq, n_outputs
+        self.rearrange = Rearrange("b c (n p) -> b c n p", p=patch_size)
```
bruAristimunha:
```suggestion
        self.rearrange = Rearrange("batch nchans (npath patchsize) -> batch nchans npath patchsize", p=patch_size)
```

bruAristimunha: This is a little more like a patch that rearranges, right?

braindecode/models/cbramod.py
bruAristimunha, line 1: I think it is missing how to load the weights

braindecode/models/base.py
bruAristimunha: why?

PierreGtch: Some models are invariant to n_chans or n_times. It's annoying to have the print function fail in case these parameters are not provided. One test was failing because of this

docs/api.rst
PierreGtch: The line `.. currentmodule:: braindecode.models` is enough to create links to the models (see generated doc), no need to have the full path for each model. 

You're sure you want to revert? 

(the diff highlight is wrong)

bruAristimunha: yeah, i checked now, okay this stuff

braindecode/models/base.py
bruAristimunha: ah, in this case

braindecode/models/base.py
```diff
@@ -397,9 +397,18 @@ def get_torchinfo_statistics(
...
+            n_chans = self.n_chans
+        except ValueError:
+            n_chans = 1
+        try:
+            n_times = self.n_times
+        except ValueError:
+            n_times = 1000
```
bruAristimunha:
```suggestion
            n_times = 200
```

bruAristimunha: 1 seconds

braindecode/models/cbramod.py
PierreGtch: this is the same method name as in `nn.TransformerEncoderLayer`, I would keep it identical for consistency
https://github.com/pytorch/pytorch/blob/main/torch/nn/modules/transformer.py#L961

braindecode/models/cbramod.py
PierreGtch: moved to `modules` to expose it and added a docstring

braindecode/models/cbramod.py
PierreGtch: yes but they use strictly non-overlapping patches so rearrange works. You think we should handle the case when signal duration is not a multiple of the patch size? For example with some 0-padding. At the moment, it will fail if not a multiple

braindecode/models/cbramod.py
PierreGtch: it was part of copy/paste. I removed it for simplicity

braindecode/models/cbramod.py
bruAristimunha: hey, sorry, i was talking more about the variable name, not the logic. I am okay with the logic

braindecode/models/cbramod.py
bruAristimunha: hm, okay.

PierreGtch:
Thanks for the review @bruAristimunha ! All implemented, except the pending comments. Let's see if the tests pass

braindecode/models/cbramod.py
PierreGtch: I prefer 1-letter var names in rearrange haha, lines become too long otherwise

bruAristimunha: let's be explicit please, it is really better for reading

PierreGtch: hmmm ok

PR #918, opened 2026-01-23
Mini changes in `BIDSDataset` and epochs

PierreGtch:
- Expose additional arguments of `mne_bids.read_raw_bids` in `braindecode.datasets.BIDSDataset` 
- Populate epochs metadata with the `mne.Annotations`'s `extras` entries


bruAristimunha:
Why not kwargs here? It would allow full compatibility with all the parameters

bruAristimunha:
And okay for the extra until get merge until mne bids

PierreGtch:
> Why not kwargs here? It would allow full compatibility with all the parameters

because `mne_bids.read_raw_bids` also does not use kwargs, and that's the only two relevant arguments. If users pass `return_event_dict`, it breaks the code

PR #931, opened 2026-02-05
[EHN] LaBraM automatic channel reordering and multiple fixes

dungscout96:
Adds automatic channel name handling and reordering to the LaBraM model, ensuring input EEG channels are processed in the correct order expected by pretrained weights.

**Key changes:**
- Add `LABRAM_CHANNEL_ORDER` constant defining the 98 standard 10-20 electrode positions used by pretrained LaBraM
- Add `ch_names` parameter to `Labram.__init__()` for explicit channel name specification
- Automatically extract channel names from `chs_info` (MNE format) when available
- Automatically reorder input channels to match the canonical LaBraM order during forward pass
- Support case-insensitive channel name matching
- Warn users about unmatched channels that will be excluded

## Test plan

- [x] Run existing LaBraM tests to ensure no regressions
- [x] Add new tests for channel reordering functionality:
  - Channel order constant export
  - Initialization with `ch_names`
  - Forward pass with reordering
  - Correct reordering order verification
  - Case-insensitive matching
  - Warning for unmatched channels
  - `chs_info` extraction
  - Gradient flow through reordering
  - GPU device compatibility

### DIFF
--- a/test/unit_tests/models/test_foundation_models.py
+++ b/test/unit_tests/models/test_foundation_models.py
@@ -253,36 +253,26 @@ def test_labram_neural_decoder_forward_pass_single_sample(
 
 
 def test_labram_can_load_pretrained_weights():
-    """Ensure that Labram can load pre-trained weights via torch.hub convenience."""
-    chs_info = [{"ch_name": ch_name} for ch_name in LABRAM_CHANNEL_ORDER]
-    model = Labram(n_times=1600, chs_info=chs_info, n_outputs=4)
-    url = "https://huggingface.co/braindecode/Labram-Braindecode/resolve/main/braindecode_labram_base.pt"
+    """Ensure that Labram can load pre-trained weights from HuggingFace Hub."""
+    mne_data_dir = mne.get_config("MNE_DATA")
+    if mne_data_dir is None:
+        mne_data_dir = str(Path.home() / "mne_data")
+    cache_dir = Path(mne_data_dir) / "labram_pretrained"
+    cache_dir.mkdir(parents=True, exist_ok=True)
+    cache_dir = str(cache_dir)
 
     try:
-        state_dict = torch.hub.load_state_dict_from_url(
-            url,
-            progress=True,
-            map_location="cpu",
-            file_name="braindecode_labram_base_resolved.pt",
+        model = Labram.from_pretrained(
+            "braindecode/labram-pretrained",
+            cache_dir=cache_dir,
         )
     except (URLError, OSError) as err:
         pytest.skip(f"Could not download pretrained Labram checkpoint: {err}")
-    # Align a few known naming/shape differences between hub checkpoint and
-    # current implementation, while still validating that weights can load.
-    if "fc_norm.weight" in state_dict and "norm.weight" not in state_dict:
-        state_dict["norm.weight"] = state_dict.pop("fc_norm.weight")
-    if "fc_norm.bias" in state_dict and "norm.bias" not in state_dict:
-        state_dict["norm.bias"] = state_dict.pop("fc_norm.bias")
 
-    state_dict.pop("patch_embed.segment_patch.patcher.weight", None)
-    state_dict.pop("patch_embed.segment_patch.patcher.bias", None)
-
-    if state_dict["position_embedding"].shape != model.position_embedding.shape:
-        state_dict.pop("position_embedding")
-
-    load_result = model.load_state_dict(state_dict, strict=False)
-    assert "position_embedding" in load_result.missing_keys
-    assert not load_result.unexpected_keys
+    # Verify model was loaded and can run a forward pass
+    x = torch.randn(1, model.n_chans, model.n_times)
+    output = model(x)
+    assert output.shape[0] == 1
 
 
 def test_labram_neural_decoder_forward_features_all_tokens(


braindecode/models/labram.py
```diff
@@ -154,6 +290,15 @@ class Labram(EEGModuleMixin, nn.Module):
...
         module class like ``nn.ReLU`` or ``nn.ELU``. Default is ``nn.GELU``.
+    ch_names : list of str, optional
+        List of channel names corresponding to the input EEG channels.
+        If provided, channels will be automatically reordered to match
+        the LaBraM pretrained model's expected order (LABRAM_CHANNEL_ORDER).
+        Channel names are matched case-insensitively.
+        If not provided but ``chs_info`` is available (from MNE), channel
+        names will be extracted from ``chs_info``.
+        If neither is provided, no channel reordering is performed and
+        the model uses the input channel order as-is.
```
PierreGtch:
```suggestion
```
This parameter is redundant with `chs_info`

there are already a lot of tests and utilitaries in place for chs_info, let's only use this one

braindecode/models/labram.py
```diff
@@ -399,6 +549,146 @@ def _init_weights(layer):
...
+
+        # Try to extract from chs_info (MNE format)
+        try:
+            chs_info = self._chs_info
+            if chs_info is not None:
+                return [ch.get("ch_name", "") for ch in chs_info]
+        except (ValueError, AttributeError):
+            pass
+
+        return None
```
PierreGtch:
```suggestion
        try:
            chs_info = self.chs_info
        except ValueError:
            return None

        return [ch.get("ch_name", "") for ch in chs_info]
```

We could even require `chs_info` to be provided by removing the try/except. Just to be sure that Labram is not missused. 
WDYT @dungscout96 @bruAristimunha ?

bruAristimunha:
I have mixed feelings about this, but I am ok to merge because we need to; we are dropping the channels that are not within the intersection when the ch_info is informed. 

I think it made me realize that we don't have a convention of what to do or how to handle things when the channels don't align with pre-trained.

Last week, @hubertjb challenged me to be more consistent with reading weights and handling these in-pair channel details. This certainly goes beyond this PR, and we will probably need to look at how other libraries do this and implement the same logic across models.

bruAristimunha:
@PierreGtch, your input is welcome :)

PierreGtch:
Ideally, the channels used to pre-train the model should be contained in the weights file and loaded along with the weights

Super ideally, the model does not rely on a fixed set of channel names but directly uses their positions (`pos = [ch['loc'] for ch in chs_info]`), like Signal-JEPA 😇

In any case, if we drop channels, we should raise a warning or error. Maybe we could add a flag:
```python
on_unknown_ch: Literal['raise', 'warn', 'ignore']
# Behaviour when pre-trained weights are loaded,
# and channels that were not used to pre-train are passed 
```

braindecode/models/labram.py
```diff
@@ -533,8 +817,12 @@ def forward(
...
             The input data with shape (batch, n_chans, n_times)
             or (batch, n_chans, n_patches, patch size).
-        input_chans: int
-            An input channel to select some dimensions
+        input_chans: torch.Tensor or list of int, optional
+            Indices for selecting position embeddings. If not provided and
+            channel information was specified during model initialization
+            (via ``chs_info``), channels will be automatically reordered to
+            match LABRAM_CHANNEL_ORDER and the corresponding position
+            embedding indices will be used.
```
PierreGtch: This feels hacky and model-specific... The users have to look up the position of their channels in `braindecode.models.labram.LABRAM_CHANNEL_ORDER`. Not so nice when you want to run a benchmark for example...

We could eventually allow to pass a list `ch_names` to `forward()` which overwrites the channels in `self.chs_info` but not indices. 

And if neither `__init__(chs_info)`, not `forward(ch_names)` are provided, we must raise an error

PierreGtch:
I implemented my comments regarding the `ch_name` and  `on_missing_ch` parameters.

However... the weights loading test fails because the positional embedding weights have 64 channels, but the `LABRAM_CHANNEL_ORDER` lists 120 channels.

@dungscout96 do you know which exact channels were used to obtain the retrained weights in https://huggingface.co/braindecode/Labram-Braindecode/resolve/main/braindecode_labram_base.pt ?

I could not find this information in the original repo 🙃

dungscout96:
> I implemented my comments regarding the `ch_name` and `on_missing_ch` parameters.
> 
> However... the weights loading test fails because the positional embedding weights have 64 channels, but the `LABRAM_CHANNEL_ORDER` lists 120 channels.
> 
> @dungscout96 do you know which exact channels were used to obtain the retrained weights in https://huggingface.co/braindecode/Labram-Braindecode/resolve/main/braindecode_labram_base.pt ?
> 
> I could not find this information in the original repo 🙃

You would hope they make it easier to find these information. After some digging:
- The 1020 channel list with their expected order is defined here: https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/utils.py#L42
- It's used in [get_input_chans](https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/utils.py#L713) to get the corresponding indices to pass to the [forward pass](https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/engine_for_finetuning.py#L19)
- The model used fixed positional embeddings which was [initialized](https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/modeling_finetune.py#L283) with 128 channels (plus a [cls] token channel)
- In the forward pass, the passed in indices is used to [look up](https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/modeling_finetune.py#L358) the corresponding channel embeddings

SO the pretrained weights should have 128 channels?? It looks like in their paper they selected a 64-channel subset to index into the channel positional embeddings. But still it should be 128 dimension.

Ok I checked the official checkpoint of Labram, it's 128 channel.
image

How did we get the weights in braindecode?

bruAristimunha:
I don't remember, I think we have the script at hugging face GitHub historical

bruAristimunha:
But let's regenerate please

PierreGtch:
get_input channels points to a list with 136 channels.... it's a mess
https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/utils.py#L42-L57

I will try to figure it out and regenerate our huggingface weights

PierreGtch:
# Test Plan
This requires cloning the original Labram repo, so it will not be included in our test suit

```python
# %% ######################################################
# Imports constants and setup
#######################################################
import sys
import torch
import re
from pathlib import Path
import difflib

import torch
from torch import nn
from braindecode.models.labram import Labram, LABRAM_CHANNEL_ORDER

# Add LaBraM repo to path
# > cd ~/Projects/
# > git clone git@github.com:935963004/LaBraM.git
labram_path = Path("~/Projects/LaBraM").expanduser()
sys.path.insert(0, str(labram_path))

from timm.models import create_model
import modeling_finetune  # Register the models to timm

DROPPED_KEYS = ["mask_token", "lm_head.weight", "lm_head.bias"]

# Load checkpoint from GitHub
print("Loading checkpoint from GitHub...")
checkpoint_url = (
    "https://github.com/935963004/LaBraM/raw/main/checkpoints/labram-base.pth"
)
checkpoint = torch.hub.load_state_dict_from_url(
    checkpoint_url,
    map_location="cpu",
    progress=True,
    check_hash=True,
    weights_only=False,
)

# Extract state dict
state_dict = {
    key[8:]: value
    for key, value in checkpoint["model"].items()
    if key.startswith("student.")
}
# Drop unused keys
for k in DROPPED_KEYS:
    del state_dict[k]


# %% ######################################################
# Instantiate the original LaBraM model with pre-trained weights.
#######################################################

# Create model
print("Creating model...")
model = create_model(
    "labram_base_patch200_200",
    # arguments form
    # https://github.com/935963004/LaBraM/blob/c431221e6cfd23dbfa9950e0180682fb322b0548/run_class_finetuning.py#L196C1-L208C6
    # and readme
    pretrained=False,
    num_classes=0,
    drop_rate=0,
    drop_path_rate=0.1,
    attn_drop_rate=0.0,
    drop_block_rate=None,
    use_mean_pooling=False,
    init_scale=0.001,
    use_rel_pos_bias=False,
    use_abs_pos_emb=True,
    init_values=0.1,
    qkv_bias=False,
)
model.eval()

# Load pre-trained weights
model.load_state_dict(state_dict, strict=True)


# %% ######################################################
# Instantiate the Braindecode implementation of LaBraM, rename the weights and load the pre-trained weights.
######################################################

# Instantiate model
chs_info = [{"ch_name": ch_name} for ch_name in LABRAM_CHANNEL_ORDER]
model_braindecode = Labram(
    n_times=3000,
    chs_info=chs_info,
    n_outputs=0,
)
_ = model_braindecode.eval()


# Map keys from original LaBraM to Braindecode implementation
braindecode_state_dict = {}
key_mappings = {}
for k, v in state_dict.items():
    new_k = k

    # simple renames
    if new_k == "pos_embed":
        new_k = "position_embedding"
    elif new_k == "time_embed":
        new_k = "temporal_embedding"

    # blocks mlp fc1/fc2 -> mlp.0 / mlp.2
    new_k = re.sub(r"blocks\.(\d+)\.mlp\.fc1", r"blocks.\1.mlp.0", new_k)
    new_k = re.sub(r"blocks\.(\d+)\.mlp\.fc2", r"blocks.\1.mlp.2", new_k)

    # patch_embed conv/norm mappings:
    m = re.match(r"patch_embed\.conv(\d+)(\..+)$", k)
    if m:
        n = int(m.group(1))
        suffix = m.group(2)
        # map convs to temporal_conv.conv{n}
        new_k = f"patch_embed.temporal_conv.conv{n}{suffix}"

    m2 = re.match(r"patch_embed\.norm(\d+)(\..+)$", k)
    if m2:
        n = int(m2.group(1))
        suffix = m2.group(2)
        new_k = f"patch_embed.temporal_conv.norm{n}{suffix}"

    # fallback: use the simple/new_k computed above
    braindecode_state_dict[new_k] = v
    if new_k != k:
        key_mappings[k] = new_k

print("Key mappings:")
max_len = max(len(old_key) for old_key in key_mappings.keys())
for old_key, new_key in key_mappings.items():
    print(f"{old_key.ljust(max_len)} -> {new_key}")

# Load mapped state dict into model
model_braindecode.load_state_dict(braindecode_state_dict, strict=True)


# %%########################################################
# Compare the outputs of the original and Braindecode implementations for the same input.
##########################################################

rng = torch.Generator().manual_seed(42)
x = torch.randn(1, 128, 3000, generator=rng)

with torch.no_grad():
    y1 = model(x.view(1, 128, 15, 200), return_all_tokens=True)
    y2 = model_braindecode(x, return_all_tokens=True)  # return_patch_tokens=True)

assert y1.shape == y2.shape
assert (y1 == y2).all()

# %%########################################################
# Push weights to hub
#########################################################

model_braindecode.push_to_hub(
    repo_id="braindecode/labram-pretrained", commit_message="Several fixes (see PR#931)"
)


# %%########################################################
# Now load form the newly pushed HF checkpoint
#########################################################

model_hf = Labram.from_pretrained(
    "braindecode/labram-pretrained",
    cache_dir=Path("~/.cache/huggingface/").expanduser(),
)


# %%########################################################
# Finally, compare the outputs of the original and HF model for the same input.
##########################################################

rng = torch.Generator().manual_seed(42)
x = torch.randn(1, 128, 3000, generator=rng)

with torch.no_grad():
    y1 = model(x.view(1, 128, 15, 200), return_all_tokens=True)
    y2 = model_hf(x, return_all_tokens=True)  # return_patch_tokens=True)

assert y1.shape == y2.shape
assert (y1 == y2).all()


# %%########################################################
# Sandbox
#########################################################

def format_model_str(model):
    return list(
        map(lambda line: line.replace(" ", ""), repr(model).splitlines(keepends=True))
    )


model_str = format_model_str(model)
model_braindecode_str = format_model_str(model_braindecode)

# diff = difflib.unified_diff(
diff = difflib.ndiff(
    model_str,
    model_braindecode_str,
    # fromfile="Original LaBraM",
    # tofile="Braindecode LaBraM",
)
print("".join(diff))
```

PierreGtch:
Thanks for handling the tests updates @bruAristimunha :D

PR #935, opened 2026-02-26
Fix verbose windower test to be resilient to annotation mutation

bruAristimunha:
## Summary

- `test_window_sizes_from_events_with_verbose` was failing in CI because it hardcoded specific annotation class names (`left_hand`, `tongue`)
- The root cause is that `test_single_sample_size_windows` mutates the module-scoped `concat_ds_targets` fixture by calling `underlying_raw.set_annotations(annotations[:3])`, reducing annotations from 4 classes to 2. This mutation persists for all subsequent tests sharing the fixture.
- The test now checks only for the presence of the `"Used Annotations descriptions:"` log prefix rather than specific class names, making it resilient to fixture mutation and annotation ordering differences.
- Also fixed an incorrect comment (`verbose is False, so we expect` → `so we should NOT`).

## Test plan
- [x] `pytest test/unit_tests/preprocessing/test_windowers.py::test_window_sizes_from_events_with_verbose -v` passes locally
- [x] CI passes on all platforms (Ubuntu, macOS, Windows) and Python versions (3.12, 3.13)

### DIFF
--- a/test/unit_tests/preprocessing/test_windowers.py
+++ b/test/unit_tests/preprocessing/test_windowers.py
@@ -945,16 +945,10 @@ def test_window_sizes_from_events_with_verbose(caplog, concat_ds_targets):
         drop_last_window=False,
         verbose=True,
     )
-    options = [
-        "np.str_('feet'), np.str_('left_hand'), np.str_('right_hand'), np.str_('tongue')",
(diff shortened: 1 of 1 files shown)


PierreGtch:
Thanks @bruAristimunha LGTM 🚀

PR #939, opened 2026-03-03
Fix labram

dungscout96:


PierreGtch:
Hi @dungscout96 thanks for pushing these changes
I'm surprised with the Claude comment regarding the dimensionality mismatch on the time dimension because I didn't have it when I ran my script to generate the HF weights in `braindecode/labram` (script here https://github.com/braindecode/braindecode/pull/931#issuecomment-3861882846). And I used load state diet with `strict=True`

bruAristimunha:
I discovered that we need to interpolate the embedding in the same way as they do in Hugging Face.... But we talk later about this.

bruAristimunha:
I am closing for now, because it already be partially address and now the diff is only showing very small (not necessary modifications)

PR #942, opened 2026-03-10
Per-event-type windowing parameters in `create_windows_from_events`

PierreGtch:
### Summary

Allow `trial_start_offset_samples`, `trial_stop_offset_samples`, and `window_stride_samples` to accept `dict[str, int]` in addition to `int`, allowing different windowing parameters per event type.

### Motivation

In many paradigms with very unbalanced classes (ex: seizure detection), we may want different strides for the different events.

### Changes

**windowers.py:**
- **`create_windows_from_events`**: Updated signature to accept `int | dict[str, int]` for `trial_start_offset_samples`, `trial_stop_offset_samples`, and `window_stride_samples`. Added validation: when any parameter is a dict, `mapping` and `window_size_samples` must be explicitly provided, and dict keys must match `mapping` keys exactly. Int parameters are normalized to dicts internally when mixed with dict parameters.
- **`_create_windows_from_events`**: Added a per-event-type code path that groups events by type, calls `_compute_window_inds` per group with the corresponding offsets/stride, maps trial indices back to global event indices, and merges results sorted chronologically by start sample.
- **`_check_windowing_arguments`**: Extended to accept `dict` types for the three parameters, validating that all dict values are integers and stride values are positive.

**test_windowers.py:**
- Added 10 new tests covering:
  - Validation errors: missing `mapping`, missing `window_size_samples`, mismatched dict keys
  - Equivalence: uniform dict values produce identical results to int values
  - Per-type offsets: different `trial_start_offset_samples` and `trial_stop_offset_samples` per event type
  - Per-type stride: different `window_stride_samples` per event type
  - Mixed parameters: some params as int, others as dict
  - Chronological ordering of output windows
  - Compatibility with `use_mne_epochs=True`

### API

```python
# Before (uniform for all events):
create_windows_from_events(ds, trial_start_offset_samples=0, window_stride_samples=50, ...)

# After (per event type):
create_windows_from_events(
    ds,
    trial_start_offset_samples={"left_hand": -100, "right_hand": 0},
    window_stride_samples={"left_hand": 25, "right_hand": 50},
    mapping={"left_hand": 0, "right_hand": 1},
    window_size_samples=500,
    ...
)
```

### Constraints

- If any of the three parameters is a `dict`, `mapping` must be provided (not inferred).
- If any parameter is a `dict`, `window_size_samples` must be explicitly set (not `None`).
- Dict keys must match `mapping` keys exactly.
- `window_size_samples` remains a scalar (`int | None`) — it is not per-event-type.
- Backward compatible: passing `int` values works identically to before.

*With Claude opus 4.6*

### DIFF
--- a/test/unit_tests/preprocessing/test_windowers.py
+++ b/test/unit_tests/preprocessing/test_windowers.py
@@ -1256,3 +1256,58 @@ def test_dict_params_per_event_type(
     if not use_mne_epochs:
         starts = metadata["i_start_in_trial"].values
         assert np.all(np.diff(starts) >= 0)
+
+
+def test_dict_params_bad_trial_dropped_extras_not_misassigned():
+    """Extras must come from the surviving trial, not from the dropped one.
+
+    When one T0 event is too short for the window size and is dropped via
+    accepted_bads_ratio, the returned i_trials from _compute_window_inds index
+    into the post-filter array.  Before the fix, those indices were mapped back
+    through orig_indices (sized for the pre-filter array), so the surviving
+    T0 window got the extras of the *dropped* trial instead of its own.
+    """
+    sfreq = 100
+    # T0 event 0: onset 1 s, duration 0.4 s  → 40 samples  (too short for window=60)
+    # T0 event 1: onset 5 s, duration 1.0 s  → 100 samples (fits)
+    # T1 events: both 1 s long (needed for mapping only)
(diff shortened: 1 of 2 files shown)


test/unit_tests/preprocessing/test_windowers.py
bruAristimunha, line 1: can you please pytest parametrize more?

PR #946, opened 2026-03-11
Fix/remove deprecated irfft and fix multitarget windowing

tayal-sarthak:
### 1. remove deprecated torch.irfft fallback (visualization/gradients.py)

the function compute_amplitude_gradients_for_X had a try/except block that
caught AttributeError to fall back to torch.irfft, which was deprecated in
pytorch 1.7 and removed in later versions. since braindecode requires
torch >= 2.2 in pyproject.toml, the fallback path was unreachable dead code.

removed the try/except and now calls torch.view_as_complex + torch.fft.irfft
directly.

### 2. fix multi-target windowing bug (preprocessing/windowers.py)

_create_windows_from_target_channels only checked target[0, :] (the first
misc channel) for non-nan values when determining window positions. if
multiple misc channels had target values at different timepoints, windows
were only placed where the first channel had data.

replaced target[0, :] with np.any(~np.isnan(target), axis=0) so windows
are placed at the union of all valid target positions across all misc
channels. single channel behavior is unchanged since np.any on one row
is equivalent.

this resolves the TODO comment that was in the code:
"TODO: handle multi targets present only for some events"

### 3. new test for partial multi-target windowing

added test_windower_from_target_channels_partial_targets which creates two
misc channels with values at completely different timepoints and verifies
that windows are produced for all positions from both channels.

## tests

all existing tests pass. the 11 errors in test_windowers.py are pre-existing
and caused by the optional moabb dependency not being installed.

### DIFF
--- a/braindecode/preprocessing/windowers.py
+++ b/braindecode/preprocessing/windowers.py
@@ -12,7 +12,7 @@
 #          Mohammed Fattouh <mo.fattouh@gmail.com>
 #          Robin Schirrmeister <robintibor@gmail.com>
 #          Matthew Chen <matt.chen42601@gmail.com>
-#          Sarthak Tayal <tayal-sarthak@github>
+#          Sarthak Tayal <sarthaktayal2@gmail.com>
 #
 # License: BSD (3-clause)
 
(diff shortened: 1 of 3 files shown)


tayal-sarthak:
hope pre-commit works now :)

bruAristimunha:
Clear and precise for me, thank you very much for your contribution!

PR #952, opened 2026-03-14
Convert `EEGWindowsDataset` to `WindowsDataset`

PierreGtch:
### Feature
This PR adds a new conversion utility that turns an EEGWindowsDataset backed by mne.Raw into a WindowsDataset backed by mne.Epochs via to_epochs_dataset.
The goal is to support a storage-efficient path for saving/loading windowed datasets while preserving dataset semantics (targets, metadata, transforms, and preprocessing history).

### Tests
New unit tests were added to cover both failure and success behavior of the conversion:

**Error handling:**
- Raises when targets are not metadata-based.
- Raises when window sizes are inconsistent.
- Raises when there are no windows to convert.

**Successful conversion:**
- Returns the expected output type (WindowsDataset / mne.Epochs).
- Preserves core properties (length, targets config, transform, description, metadata).
- Keeps data-level consistency between original and converted datasets (X, y, crop indices).
- Copies raw preprocessing kwargs without aliasing the same list instance.

### DIFF
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -650,10 +650,14 @@ class EEGWindowsDataset(RecordDataset):
         window_kwargs.append((self.to_epochs_dataset.__name__, {}))
         setattr(windows, "window_kwargs", window_kwargs)
         if hasattr(self, "raw_preproc_kwargs"):
-            setattr(windows, "raw_preproc_kwargs", getattr(self, "raw_preproc_kwargs"))
+            setattr(
+                windows, "raw_preproc_kwargs", list(getattr(self, "raw_preproc_kwargs"))
+            )
         if hasattr(self, "window_preproc_kwargs"):
             setattr(
-                windows, "window_preproc_kwargs", getattr(self, "window_preproc_kwargs")
+                windows,
+                "window_preproc_kwargs",
+                list(getattr(self, "window_preproc_kwargs")),
             )
 
         return windows


bruAristimunha:
Are we super, super, sure that we are not doing some regression in the speed-time of the dataloader?

A small test is necessary.

PierreGtch:
> Are we super, super, sure that we are not doing some regression in the speed-time of the dataloader?
> 
> A small test is necessary.

Not sure what changed but now datasets using epochs are 5x faster than with raw recordings!

```
Building synthetic EEGWindowsDataset with parameters:
n_chans=32, sfreq=100.0, window_size_samples=200, window_stride_samples=300, n_windows=10000
/Users/Pierre.Guetschel/Projects/braindecode/braindecode/datasets/base.py:696: UserWarning: The provided mne.Epochs object does not meet the requirements for fast epoch access. This may lead to slow data loading from disk. Consider preloading the epochs or checking the conditions in WindowsDataset._can_use_fast_get_epoch_from_raw.
  warnings.warn(
Benchmark: EEGWindowsDataset vs WindowsDataset (from disk, preload=False)
Temporary benchmark directory: /Users/Pierre.Guetschel/Desktop/tmp
Number of windows: 10001

EEGWindowsDataset
  mean iteration time: 1.0021 s (+/- 0.0237)
  throughput: 9979.70 windows/s

WindowsDataset (converted via to_epochs_dataset)
  mean iteration time: 0.1990 s (+/- 0.0003)
  throughput: 50261.66 windows/s

Speed ratio (epochs/eeg): 0.199x
```
[benchmark_to_epochs_dataset.py](https://github.com/user-attachments/files/25997797/benchmark_to_epochs_dataset.py)

PR #955, opened 2026-03-16
fix squeeze and excitation channel reduction

tayal-sarthak:
fix channel reduction in squeezandexcitation to avoid shape mismatches

fixes issue #889

the squeezandexcitation module had a bug in its fc2 layer initialization. it was using reduction_rate as the in_channels parameter instead of sq_channels, which caused a shape mismatch at runtime.

specifically:
- fc1 outputs sq_channels (e.g., 8 when in_channels=32 and reduction_rate=4)
- fc2 was incorrectly expecting in_channels=reduction_rate (e.g., 4)
- this mismatch would cause a runtime error when data passed through fc2

the fix updates attention.py to use sq_channels for fc2's in_channels, ensuring the channel dimensions match through the module.

changes:
- attention.py: line 63 changed in_channels from reduction_rate to sq_channels
- test_modules.py: added test_squeeze_and_excitation_forward to verify correct channel flow
- whats_new.rst: documented the fix in the bugs section

### DIFF
--- a/braindecode/modules/attention.py
+++ b/braindecode/modules/attention.py
@@ -8,6 +8,7 @@ AttentionBaseNet class.
 
 # Authors: Martin Wimpff <martin.wimpff@iss.uni-stuttgart.de>
 #          Bruno Aristimunha <b.aristimunha@gmail.com>
+#          Sarthak Tayal <sarthaktayal2@gmail.com>
 #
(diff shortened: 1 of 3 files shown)


bruAristimunha:
many thanks @tayal-sarthak!

PR #956, opened 2026-03-16
[HOTFIX] update `push_to_hub` usage in examples

PierreGtch:


PierreGtch:
Fixes an example after change in API of `push_to_hub` in #945

PR #959, opened 2026-03-20
fix bandstop filter notch freq range to match docstring

### LINKED ISSUE #548: Confusution between description and the code in the BandStop augmentation
I'm trying to understand this part of the BandStop augmentation. Why in the description write bandwith/2 but in the codes it use bandwith*2?
https://github.com/braindecode/braindecode/blob/436e641ba366ca64d40effcfb611469b88abd8de/braindecode/augmentation/transforms.py#L606C1-L629C1
@agramfort @cedricrommel

tayal-sarthak:
the notch center frequency bounds in BandstopFilter.get_augmentation_params used `2 * bandwidth` when it should be `bandwidth / 2`. the docstring already documented the corect formula (`bandwidth/2 + transition`) but the code was doing `2 * bandwidth + transition`, making the sampling range 4x tighter than intended.

changed lines 602-603 in transforms.py to use `self.bandwidth / 2` on both low and high bounds. added auther line per contribution guidelines.

closes #548


bruAristimunha:
Many thanks again @tayal-sarthak 🙏🏽

PR #989, opened 2026-04-14
handle missing numpydantic in config without crashing on import

### LINKED ISSUE #871: TypeError: type 'Any' is not subscriptable when importing braindecode.models.config without numpydantic
Importing `braindecode.models.config` (or `EEGNetConfig`) fails with:

`TypeError: type 'Any' is not subscriptable`

This happens when `numpydantic` is **not** installed. The fallback in `braindecode/models/config.py` sets `NDArray = Any` and `Shape = Any`, but `ChsInfoType` still uses `NDArray[Shape["12"], np.float64]`, which raises on Python 3.12.

### Minimal example to reproduce

python3.12 -c "from braindecode.models.config import EEGNetConfig"or in an interactive session:

Python 3.12.11 (main, Sep 18 2025, 19:41:45) [Clang 20.1.4] on darwin
>>> from braindecode.models.config import EEGNetConfig
  File "<stdin>", line 1, in <module>
  File "/Users/aman/Desktop/INC/braindecode/braindecode/models/config.py", line 23, in <module>
    class ChsInfoType(TypedDict, total=False, closed=True):  # type: ignore[call-arg]
  File "/Users/aman/Desktop/INC/braindecode/braindecode/models/config.py", line 29, in ChsInfoType
    loc: NDArray[Shape["12"], np.float64]  # type: ignore[misc]
                 ~~~~~^^^^^^
TypeError: type 'Any' is not subscriptable### Relevant code

`braindecode/models/config.py`:

from typing import Any
from typing_extensions import TypedDict
from mne.utils import _soft_import
import numpy as np

pydantic = _soft_import(name="pydantic", purpose="model configuration", strict=False)

try:
    from numpydantic import NDArray, Shape
except ImportError:
    NDArray = Any  # type: ignore
    Shape = Any  # type: ignore


class ChsInfoType(TypedDict, total=False, closed=True):  # type: ignore[call-arg]
    ...
    loc: NDArray[Shape["12"], np.float64]  # type: ignore[misc]
    ...On Python 3.12, doing `NDArray = Any; Shape = Any` and later using `NDArray[Shape["12"], np.float64]` results in `TypeError: type 'Any' is not subscriptable`.

### Expected behavior

If `numpydantic` is not installed, importing `braindecode.models.config` should still succeed (possibly with less precise typing / validation), without raising a `TypeError` at import time.

### Environment

- **OS**: macOS (darwin 24.6.0), Apple Silicon
- **Python**: 3.12.11
- **Braindecode version**: built from source from `master` (as of 2025‑12‑16)  

### Workaround

Installing `numpydantic` (or `braindecode[typing]`) avoids the error, since `NDArray` and `Shape` are then provided by `numpydantic`:

pip install "braindecode[typing]"

or

pip install numpydantic 

After installing `numpydantic`, the import:
from braindecode.models.config import EEGNetConfigworks as expected.

### Possible fix

One possible direction would be to guard the use of `NDArray` / `Shape` when they are `Any` (e.g. avoid subscripting them in `ChsInfoType` when `numpydantic` is not available), make numpydantic a required dependency, or to adjust the fallback to a plain `np.ndarray` annotation to prevent the `TypeError` on Python 3.12.

tayal-sarthak:
when numpydantic is not installed, importing braindecode.models.config crashes with TypeError because the fallback sets NDArray and Shape to typing.Any, which is not subscriptable on python 3.12+ in certain builds and python 3.14.

the fix precomputes the loc type annotation inside the try/except block. if numpydantic is available it uses the precise NDArray type, otherwise it falls back to np.ndarray. this avoids subscripting Any at annotation evaluation time.

closes #871

### DIFF
--- a/braindecode/models/config.py
+++ b/braindecode/models/config.py
@@ -1,3 +1,4 @@
+# Authors: Sarthak Tayal <sarthaktayal2@gmail.com>
 from collections.abc import Callable
 from inspect import signature
 from types import UnionType
@@ -17,10 +18,10 @@ from braindecode.models.util import (
 
 try:
     from numpydantic import NDArray, Shape
+
+    _loc_type = NDArray[Shape["12"], np.float64]
 except ImportError:
-    # we can't use soft import for numpydantic because numpydantic does not define its version in __init__
-    NDArray = Any  # type: ignore
-    Shape = Any  # type: ignore
+    _loc_type = np.ndarray  # type: ignore[misc]
 
 
 class ChsInfoType(TypedDict, total=False, closed=True):  # type: ignore[call-arg]
@@ -29,7 +30,7 @@ class ChsInfoType(TypedDict, total=False, closed=True):  # type: ignore[call-arg
     coil_type: int
     coord_frame: int
     kind: str
-    loc: NDArray[Shape["12"], np.float64]  # type: ignore[misc]
+    loc: _loc_type  # type: ignore[misc]
     logno: int
     range: float
     scanno: int
--- a/braindecode/models/config.py
+++ b/braindecode/models/config.py
@@ -30,7 +30,7 @@ class ChsInfoType(TypedDict, total=False, closed=True):  # type: ignore[call-arg
     coil_type: int
     coord_frame: int
     kind: str
-    loc: _loc_type  # type: ignore[misc]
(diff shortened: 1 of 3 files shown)


bruAristimunha:
amazing catch!

tayal-sarthak:
thank you! :)

PR #990, opened 2026-04-16
default window_stride_samples to window_size_samples when None

### LINKED ISSUE #509: Inconsistency between doc and code in create_fixed_length_windows
The doc says :
"window_stride_samples ([int](https://docs.python.org/3/library/functions.html#int) | None) – Stride between windows in samples. If None, set to be equal to winddow_size_samples, so windows will not overlap."
but when `window_stride_samples` is None and `window_size_samples` is not, we get an assertion error. 
`window_stride_samples` is not set to be equal to  `window_size_samples` :
https://github.com/braindecode/braindecode/blob/master/braindecode/preprocessing/windowers.py#L583

tayal-sarthak:
create_fixed_length_windows docstring says window_stride_samples defaults to window_size_samples when None, creating non-overlapping windows. the code crashes with an AssertionError instead.

now _check_and_set_fixed_length_window_arguments handles the default before validation runs. stride gets set to window_size_samples, drop_last_window gets set to True.

closes #509


bruAristimunha:
thank you again @tayal-sarthak 🙏🏽

PR #992, opened 2026-04-19
Remove BENDR channel_projection, hardcode pre-training channel list

PierreGtch:
## Summary

- Removes the `channel_projection` layer and the `n_chans_pretrained` /
  `chan_proj_max_norm` parameters from `BENDR`. These were added as a
  shim for users whose datasets did not have the 20 pre-training channels,
  but they were masking a silent-load bug in `BENDR.from_pretrained`:
  `load_state_dict(..., strict=False)` was matching **0 of 99 weights**
  because the published checkpoint stored its state dict under
  `encoder_state_dict` / `contextualizer_state_dict` top-level keys that
  braindecode's loader did not unwrap. Every call to
  `BENDR.from_pretrained("braindecode/braindecode-bendr", ...)` up to now
  was silently returning a fully random model.
- Hard-codes the 20 pre-training channels (19 EEG + 1 `SCALE` placeholder)
  as `_BENDR_TARGET_CHS_TUPLES` in `braindecode/models/bendr.py`. The 19
  EEG names and order come verbatim from
  [`dn3.transforms.instance.To1020.EEG_20_div`](https://github.com/SPOClab-ca/dn3/blob/master/dn3/transforms/instance.py)
  (the montage transform applied during BENDR's TUEG pre-training).
  Positions are from MNE's `standard_1005`; the `SCALE` channel has no
  physical position (it is a relative-amplitude statistic appended by
  `To1020(include_scale_ch=True)`) so its `loc` is the centroid of the
  19 EEG positions — a placeholder used only by the
  `ChannelInterpolationLayer` in `InterpolatedBENDR`.
- Rejects non-canonical `chs_info` in `BENDR.__init__` (mirrors the
  `Labram` / `InterpolatedLaBraM` pattern): if the user supplies
  `chs_info`, its names must match `BENDR_CHANNEL_ORDER` exactly
  (case-insensitive); otherwise they should use `InterpolatedBENDR`. The
  common `n_chans=20, chs_info=None` path (incl. `from_pretrained`) is
  untouched.
- Adds `InterpolatedBENDR`, the `InterpolatedModel` wrapper that accepts
  arbitrary user `chs_info` and projects to the canonical 20 BENDR
  channels. Same API as `InterpolatedLaBraM` / `InterpolatedBIOT` /
  `InterpolatedSignalJEPA` introduced in #993.
- **Companion change (already done):** the official HF checkpoint at
  [`braindecode/braindecode-bendr`](https://huggingface.co/braindecode/braindecode-bendr)
  has been re-uploaded in a flat state-dict layout (`model.safetensors`,
  legacy `pytorch_model.bin` removed). `config.json` now includes
  `n_outputs=2` so the smart head-swap path in
  `EEGModuleMixin._from_pretrained` keeps working across arbitrary
  user-requested `n_outputs` under `strict=True`.

## Motivation / context

The silent load was surfaced by OpenEEG-Bench's backbone validation,
which checks that every missing key returned by `load_state_dict` is
covered by the head or explicitly training-required modules. Under that
check, BENDR reports 99 missing learnable parameters (the full encoder
+ contextualizer). A pure key-prefix remap
(`encoder.*` → `encoder.encoder.*`,
`transformer_layers.*` → `contextualizer.transformer_layers.*`) loads
99/99 weights with 0 shape mismatches and 0 unexpected keys, so the
checkpoint itself was fine — only its wire format was wrong.

Rather than teach the loader to accept the legacy nested layout, we
re-uploaded the checkpoint flat and dropped the ad-hoc
channel-projection shim so the API is honest about what the pretrained
weights actually expect: exactly 20 channels in the canonical order,
with arbitrary user channel sets handled explicitly through
`InterpolatedBENDR`.

## Test plan

- [x] `pytest -k bendr` on `test_models.py`, `test_integration.py`,
  `test_interpolated.py`, `test_interpolation.py`,
  `test_return_features.py`, `test_model_categorization.py`,
  `test_huggingface.py` — 207 passed, 22 skipped.
- [x] Verified each hard-coded position in `_BENDR_TARGET_CHS_TUPLES`
  matches `mne.channels.make_standard_montage("standard_1005")` within
  floating-point precision (~1e-17 m), including the T5==P7 and T6==P8
  legacy-name equivalence.
- [x] Verified `BENDR(n_chans=20, ...)` instantiates and forward-passes;
  `model.channel_projection` attribute is gone.
- [x] Verified `BENDR(chs_info=non_canonical)` raises `ValueError`
  pointing to `InterpolatedBENDR`.
- [x] Verified `InterpolatedBENDR(chs_info=<user>)` instantiates,
  forward-passes, and that the interpolation matrix has shape
  `(20, len(user))`.
- [x] Verified end-to-end against the re-uploaded HF checkpoint:
  `BENDR.from_pretrained("braindecode/braindecode-bendr", strict=True,
  n_outputs=2)` and `n_outputs=5` (smart head-swap path) both load all
  99 pretrained weights cleanly; the loaded encoder weights differ from
  a fresh random model (mean-abs diff ≈ 0.067 on
  `encoder.encoder.Encoder_0.0.weight`).


PierreGtch:
@bruAristimunha are you ok with removing the custom channel adaptation layer (and the `n_chans_pretrained` param) of BENDR? in favour of the `InterpolatedBENDR` (#993)

If it's ok, you can merge

bruAristimunha:
hey Pierre, i think this will destroy the paper that kuntal is working, let's talk quick?

bruAristimunha (approved):

PierreGtch:
@Kkuntal990, with PR #993, the `channel_projection` is not necessary anymore. You can use the class `InterpolatedBENDR(..., trainable=true)` instead which does the same as `channel_projection` did but with a non-random initialisation.

Would that work for you?

PR #996, opened 2026-04-23
[EHN] Add  MetaNeuromotorHand model

bruAristimunha:
## Summary
- add `GenericNeuromotorInterface`, a Meta handwriting sEMG conformer model for CTC sequence emissions
- register the model in exports, mandatory-parameter metadata, non-classification handling, and summary categorization
- add compact unit coverage plus integration skips for compile/export paths that do not handle the MPF STFT/eigendecomposition stack

## Notes
This implementation is marked as derived from Meta's CC BY-NC 4.0 reference implementation, not Braindecode's BSD-3 license. Please confirm license compatibility before moving the PR out of draft.

## Validation
- `python -m pytest test/unit_tests/models/test_models.py -k gni -q`
- `python -m pytest test/unit_tests/models/test_integration.py -k GenericNeuromotorInterface -q`
- `python -m pytest test/unit_tests/models/test_config.py -k GenericNeuromotorInterface -q`
- `python -m pytest test/unit_tests/models/test_model_categorization.py -q`
- `pre-commit run`

### DIFF
--- a/braindecode/models/generic_neuromotor_interface.py
+++ b/braindecode/models/generic_neuromotor_interface.py
@@ -1042,12 +1042,11 @@ class GenericNeuromotorInterface(EEGModuleMixin, nn.Module):
     Conformer-based surface-EMG-to-character decoder for the handwriting task of
     Meta's generic neuromotor interface (CTRL-labs at Reality Labs, Nature 2025).
     It takes raw 16-channel surface EMG recorded at the wrist and emits a
-    per-token score sequence that feeds directly into a CTC loss or decoder
-    [graves2006ctc]_. The
-    upstream repository (``facebookresearch/generic-neuromotor-interface``)
-    ships three sibling architectures for three tasks (1-DOF wrist control,
-    discrete gestures, handwriting); only the handwriting / conformer head is
-    ported here.
+    per-token score sequence that feeds a CTC loss or decoder
+    [graves2006ctc]_. The upstream repository
+    (``facebookresearch/generic-neuromotor-interface``) has one architecture
+    per task for 1-DOF wrist control, discrete gestures, and handwriting;
+    only the handwriting / conformer head is ported here.
 
     The pipeline mirrors the paper's Methods (Fig. 1 / Extended Data Fig. 6):
 
@@ -1069,7 +1068,7 @@ class GenericNeuromotorInterface(EEGModuleMixin, nn.Module):
        Per-block stride ``2`` at blocks 5 and 10, attention window ``16`` for
        blocks 1–10 and ``8`` for blocks 11–15 (the paper's schedule).
        Attention is restricted to a fixed local window ending at the current
-       frame, giving a streaming / causal model.
+       frame, so the model runs as a streaming causal decoder.
     5. Linear readout to ``n_outputs`` classes, optionally followed by
        :class:`~torch.nn.LogSoftmax`.
 
@@ -1088,8 +1087,9 @@ class GenericNeuromotorInterface(EEGModuleMixin, nn.Module):
     dataset and Reddit, in three postures (seated on surface, seated on leg,
     standing on leg). Participants wrote letters, digits, words and phrases;
     spaces were either implicit or prompted by a right-dash token that the
-    participant produced with a right index swipe. Train/val/test split was
-    geometric scaling from 25 to 6,527 train participants, 50 val, 50 test.
+    participant produced with a right index swipe. Training sizes scale
+    geometrically from 25 to 6,527 participants; validation and test sets
+    hold 50 participants each.
 
     .. rubric:: MPF featurizer (paper defaults)
 
--- a/braindecode/models/generic_neuromotor_interface.py
+++ b/braindecode/models/generic_neuromotor_interface.py
@@ -16,7 +16,7 @@ from typing import Literal, cast
 
 import torch
 import torch.nn.functional as F
-from einops import rearrange
+from einops import pack, rearrange, reduce, repeat, unpack
 from einops.layers.torch import Rearrange
 from torch import nn
 
@@ -113,25 +113,29 @@ class _FrequencyBandAverager(nn.Module):
             return x
 
         freq_masks = cast(torch.Tensor, self.freq_masks)
-        return torch.stack(
-            [
-                (x * freq_mask).sum(2) / freq_mask.sum(2)
-                for freq_mask in freq_masks.unbind(2)
-            ],
-            dim=2,
+        weighted = rearrange(
+            x,
+            "batch time freq channels1 channels2 -> "
+            "batch time 1 freq channels1 channels2",
         )
+        weighted = weighted * freq_masks
+        return weighted.sum(dim=3) / freq_masks.sum(dim=3)
 
 
 class _CrossSpectralDensity(nn.Module):
     """Estimate channel-pair cross-spectral density inside each MPF window."""
 
     def forward(self, inputs: torch.Tensor) -> torch.Tensor:
-        input_dims = inputs.shape
-        num_channels, window_size = input_dims[-2:]
-        outputs = inputs.reshape(-1, num_channels, window_size)
-        outputs = (outputs @ outputs.transpose(-2, -1).conj()) / window_size
+        window_size = inputs.shape[-1]
+        outputs, packed_shape = pack([inputs], "* channels window")
+        outputs = (
+            outputs
+            @ rearrange(
+                outputs.conj(), "items channels window -> items window channels"
+            )
+        ) / window_size
         outputs = outputs.abs().pow(2)
-        return outputs.reshape(*input_dims[:-2], num_channels, num_channels)
+        return unpack(outputs, packed_shape, "* channels1 channels2")[0]
 
 
 class _SPDMatrixLog(nn.Module):
@@ -140,7 +144,9 @@ class _SPDMatrixLog(nn.Module):
     def forward(self, inputs: torch.Tensor) -> torch.Tensor:
         eigvals, eigvecs = torch.linalg.eigh(inputs)
         eigvals = eigvals.log().nan_to_num(nan=0.0, neginf=0.0)
-        return (eigvecs * eigvals.unsqueeze(dim=-2)) @ eigvecs.transpose(-1, -2)
+        return (eigvecs * rearrange(eigvals, "... channels -> ... 1 channels")) @ (
+            rearrange(eigvecs, "... row col -> ... col row")
+        )
 
 
 class _MultivariatePowerFrequencyFeatures(nn.Module):
@@ -351,7 +357,7 @@ class _RotationInvariantMPFMLP(nn.Module):
         x = self.flatten_features(x)
         for layer in self.fully_connected_layers:
             x = self.activation(layer(x))
-        x = torch.mean(x, dim=2, keepdim=False)  # average over rotations
+        x = reduce(x, "batch time rotation features -> batch time features", "mean")
         return rearrange(x, "batch time features -> batch features time")
 
 
@@ -641,18 +647,31 @@ class _MultiHeadAttention(nn.Module):
         return cast(torch.Tensor, self.attn_mask)
 
     def _windows(self, inputs: torch.Tensor) -> torch.Tensor:
-        batch_size, input_time = inputs.shape[:2]
-        inputs = inputs.reshape((batch_size, input_time, self.input_dim))
+        inputs = rearrange(inputs, "batch time ... -> batch time (...)")
+        if inputs.shape[-1] != self.input_dim:
+            raise ValueError(
+                f"Expected flattened input dim {self.input_dim}, got {inputs.shape[-1]}"
+            )
         windows = self.attention_window(inputs)
-        windows = windows.transpose(-1, -2).reshape(
-            (batch_size, -1, self.window_size, self.input_dim)
+        return rearrange(
+            windows, "batch time channels window -> batch time window channels"
         )
-        return windows
 
     def _attn_params(self, windows: torch.Tensor):
-        windows = windows.reshape(-1, self.window_size, self.input_dim)
+        windows = rearrange(
+            windows, "batch time window channels -> (batch time) window channels"
+        )
         query = windows[:, -1]
-        kv = windows.transpose(1, 2).reshape(-1, self.window_size, self.input_dim)
+        # Match the upstream key/value ordering from transpose(...).reshape(...).
+        kv = rearrange(
+            windows, "batch_time window channels -> batch_time (channels window)"
+        )
+        kv = rearrange(
+            kv,
+            "batch_time (window channels) -> batch_time window channels",
+            window=self.window_size,
+            channels=self.input_dim,
+        )
         return query, kv, kv
 
     def _attn_mask_op(
@@ -684,10 +703,11 @@ class _MultiHeadAttention(nn.Module):
     ) -> torch.Tensor:
         batch_size, output_time = windows.shape[:2]
         attn_mask = self._attn_mask_op(windows, t_start, t_end)
-        return (
-            attn_mask.reshape(1, output_time, 1, -1)
-            .expand(batch_size, output_time, self.op.num_heads, -1)
-            .reshape(batch_size * output_time * self.op.num_heads, -1)
+        return repeat(
+            attn_mask,
+            "time window -> (batch time heads) window",
+            batch=batch_size,
+            heads=self.op.num_heads,
         )
 
     def forward(self, inputs: torch.Tensor) -> torch.Tensor:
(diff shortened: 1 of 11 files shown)


braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+            raise ValueError("Only 3, 4, 5, and 6D tensors are supported")
+        self.from_dims = from_dims
+        self.to_dims = to_dims
+        self._permute_idx: list[int] = [from_dims.index(d) for d in to_dims]
+
+    def get_inverse_permute(self) -> "_Permute":
+        return _Permute(from_dims=self.to_dims, to_dims=self.from_dims)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        return x.permute(self._permute_idx)
```
bruAristimunha: Replace for einops

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+            n_fft=self.n_fft,
+            hop_length=self.fft_stride,
+            window=cast(torch.Tensor, self.window),
+            center=False,
+            normalized=False,
+            onesided=True,
+            return_complex=True,
+        ) / cast(
+            torch.Tensor, self.window_normalization_factor
+        )  # (batch * channels, freqs, time)
```
bruAristimunha: Move this for an layers

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+
+        outputs = outputs.unfold(
+            dimension=-1,
+            size=self.window_length // self.fft_stride,
+            step=self.stride // self.fft_stride,
+        )  # (batch * channels, freqs, windows, window_size)
```
bruAristimunha: Should be layer

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+                    for freq_mask in freq_masks.unbind(2)
+                ],
+                dim=2,
+            )
+
+        # Matrix logarithm (Barachant et al. 2012) via eigen-decomposition.
+        eigvals, eigvecs = torch.linalg.eigh(outputs)
+        eigvals = eigvals.log().nan_to_num(nan=0.0, neginf=0.0)
+        outputs = (eigvecs * eigvals.unsqueeze(dim=-2)) @ eigvecs.transpose(-1, -2)
+
```
bruAristimunha: Use SPD Learn

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+        # (batch, time, freq, chan, chan) -> (batch, freq, chan, chan, time)
+        outputs = outputs.permute(0, 2, 3, 4, 1)
```
bruAristimunha: Einops layers

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+        for i in range(num_channels):
+            for j in range(num_adjacent_diagonals):
+                mask[i, (i + j + 1) % num_channels] = 1
+                mask[i, (i - j - 1) % num_channels] = 1
+        return mask
+
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+        inputs = inputs.flatten(4)
+        indices = cast(torch.Tensor, self.flattened_matrix_indices)
+        return inputs[:, :, :, :, indices]
```
bruAristimunha: Check spd learn for replace

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+        if not self.training:
+            return inputs
+        x = inputs
+        for i, mask in enumerate(self.masks):
+            n_masks = int(torch.randint(self.max_num_masks[i] + 1, size=()).item())
+            for _ in range(n_masks):
+                x = mask(x)
+        return x
```
bruAristimunha: Optimize for module list

braindecode/models/meta_neuromotor.py
```diff
@@ -0,0 +1,1299 @@
...
+                x = mask(x)
+        return x
+
+
+class _Window(nn.Module):
+    """Extract sliding windows from a sequence.
+
+    Input ``(N, T, ...)`` -> output ``(N, T', ..., kernel_size)`` where
+    ``T' = ceildiv(T - dilation*(kernel_size-1), stride)``.
+    """
```
bruAristimunha, line 1274: Doc string should be much better

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+
+    def forward(self, inputs: torch.Tensor) -> torch.Tensor:
+        if self.state_size > 0:
+            buffered = F.pad(inputs, (0, 0, self.state_size, 0), "constant", 0)
+        else:
+            buffered = inputs
+        windows = buffered.unfold(
+            dimension=1, size=self.receptive_field, step=self.stride
+        )
+        return windows[..., :: self.dilation]
```
bruAristimunha: Should be a layer

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+        self.input_dim = input_dim
+        self.num_heads = num_heads
+        self.window_size = window_size
+        self.op = nn.MultiheadAttention(
+            embed_dim=input_dim,
+            num_heads=num_heads,
+            dropout=dropout,
+            batch_first=True,
+        )
+        self.window = _Window(
```
bruAristimunha: More descriptive variable name

braindecode/models/meta_neuromotor.py
```diff
@@ -0,0 +1,1299 @@
...
+            attn_mask=attn_mask.unsqueeze(1),
+        )
+        output = output.reshape(N, -1, self.input_dim).unsqueeze(2)
+        return output
+
+
+class _Conv1d(nn.Module):
+    """Conv1d in ``NTC`` format that exposes ``extra_left_context``/``stride``."""
+
+    def __init__(
```
bruAristimunha, line 1546: Real doc string

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+
+def _broadcast(val, num_layers: int, name: str):
+    if isinstance(val, (list, tuple)):
+        out = list(val)
+        if len(out) != num_layers:
+            raise ValueError(f"{name} length {len(out)} != num_layers {num_layers}")
+        return out
+    return [val] * num_layers
+
+
```
bruAristimunha: Do we really need this?

braindecode/models/generic_neuromotor_interface.py
```diff
@@ -0,0 +1,1299 @@
...
+    16,
+    16,
+    16,
+    8,
+    8,
+    8,
+    8,
+    8,
+)
+_PAPER_NUM_LAYERS: int = 15
```
bruAristimunha: It is not super nice this, should be much more compact

braindecode/models/meta_neuromotor.py
```diff
@@ -0,0 +1,1299 @@
...
+        return attn_window
+    if num_layers == _PAPER_NUM_LAYERS:
+        return _PAPER_CONFORMER_ATTN_WINDOW_15
+    return 16
+
+
+class GenericNeuromotorInterface(EEGModuleMixin, nn.Module):
+    r"""Generic neuromotor interface for handwriting from Meta (2025) [gni2025]_.
+
+    :bdg-info:`Attention/Transformer` :bdg-success:`Convolution`
```
bruAristimunha: Recurrent (?)

braindecode/models/meta_neuromotor.py
```diff
@@ -0,0 +1,1299 @@
...
+        from braindecode.models import GenericNeuromotorInterface
+
+        ckpt = torch.load("model_checkpoint.ckpt", weights_only=False)
+        sd = {k[len("network."):]: v for k, v in ckpt["state_dict"].items()}
+        # The final linear moved out of the conformer into `final_layer`.
+        sd["final_layer.weight"] = sd.pop("conformer.3.weight")
+        sd["final_layer.bias"] = sd.pop("conformer.3.bias")
+
+        model = GenericNeuromotorInterface(n_times=32000, log_softmax=True)
+        model.load_state_dict(sd, strict=True)
```
bruAristimunha: Put the mapping variables

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Resolved in bc3c698c: `_Permute` removed; all static-pattern rearranges are now `einops.layers.torch.Rearrange` layers declared in `__init__`.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Resolved in bc3c698c: window-normalization divide is now inside `_ChannelwiseSTFT.forward`; the `(batch, channels, time) -> (batch*channels, time)` flatten is a `Rearrange` layer (`flatten_batch_channels`); unflatten is `tensor.unflatten(0, (batch, channels))`.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Left as a direct `tensor.unfold(...)` call after e6461395 where the one-use `_SlidingWindowLast` wrapper was removed. If you still want a named layer at this site I can restore it - but wanted to flag the tension with your earlier over-engineering request.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Added a `See Also` pointer to `spd_learn.functional.matrix_log` in 5311e0b5. I kept the local implementation because spd_learn's matrix log applies an eigenvalue safe-clamp, which diverges from Meta's pretrained-checkpoint numerics (the bit-exact parity I verify in the eval harness). Happy to switch if you decide bit-exact upstream parity is not required.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Resolved in bc3c698c: the final `permute(0, 2, 3, 4, 1)` is now `self.time_last_layout = Rearrange('batch time freqs channels1 channels2 -> batch freqs channels1 channels2 time')`.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Added a `See Also` pointer to `spd_learn.functional.sym_to_upper` in 5311e0b5. I didn't replace the implementation because `_VectorizeSymmetricMatrix` carries the armband-specific circular-adjacency masking (keep only `|i - j| mod n_channels <= num_adjacent_cov`), which the generic vectorizer doesn't have.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Resolved in e6461395: `_MaskAug` is now a single class whose `forward` is two nested loops over a pre-resolved plan list (no `ModuleList` of `_RepeatedRandomMask` wrapping `_AxesMask`). ~-35 LOC.

braindecode/models/meta_neuromotor.py
bruAristimunha: Resolved in 5311e0b5: `_Window` now has a full Parameters / Attributes / Notes docstring covering `kernel_size`, `stride`, `dilation`, `lpad` (with the three modes `none`/`steady`/`full` spelled out), and the I/O shape formula.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Left as inline `windows[..., :: self.dilation]` after ff43310d where the one-use `_TakeEvery` helper was removed. Happy to restore it as an explicit layer if you prefer module indirection over an inline slice.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Already renamed in an earlier commit: the `_Window` attribute inside `_MultiHeadAttention` is `self.attention_window` (previously collided with `_ChannelwiseSTFT.window`). The cascade of internal reshape helpers was also renamed to say what they do (`flatten_trailing`, `merge_batch_time`, `add_query_axis`, ...).

braindecode/models/meta_neuromotor.py
bruAristimunha: Resolved in 5311e0b5: `_Conv1d` now has a full docstring explaining why the NTC<->NCT sandwich exists (rest of the conformer block is NTC), which attributes feed into `_SlicedSequential`'s residual alignment, and which conformer-block call site uses it (depthwise temporal conv, kernel_size=8, groups=input_dim).

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Kept the helper but tightened its docstring in 5311e0b5 to say *why*: three identical broadcast+validate schedules in `_conformer_encoder` (`kernel_size`, `attn_window_size`, `attn_lpad`). `stride` stays inline because its broadcast rule is different (scalar applies only to the last block).

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Resolved in an earlier commit: the paper's 15-layer schedule is now a one-liner per config: `_PAPER_CONFORMER_STRIDE_15 = (1, 1, 1, 1, 2) * 2 + (1,) * 5` and `_PAPER_CONFORMER_ATTN_WINDOW_15 = (16,) * 10 + (8,) * 5`.

braindecode/models/meta_neuromotor.py
bruAristimunha: Good catch - the model is pure feedforward conformer so no `Recurrent` badge. Also dropped the `:bdg-primary:\`CTC\`` I'd added in 5311e0b5 (`bdg-primary` is the Filterbank colour in `test_model_categorization.CATEGORIZATION_BADGES`). Only the two valid categories now: `Convolution` and `Attention/Transformer`.

braindecode/models/meta_neuromotor.py
bruAristimunha: Resolved in an earlier commit: the docstring's loading example now relies on the class-level `mapping` dict that auto-applies on `load_state_dict`. It strips the `network.` prefix and calls `load_state_dict(sd, strict=True)`; the `conformer.3 -> final_layer` remap is automatic via `EEGModuleMixin.load_state_dict`.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Correction to my earlier reply: actually addressed in bfb83e4b. `.unfold(...)` is now a dedicated `_SlideWindow` nn.Module, used as `self.frame` in `_MultivariatePowerFrequencyFeatures`. Named `_SlideWindow` rather than `_Unfold` to avoid shadowing `torch.nn.Unfold`, which is the 2D im2col-style op and doesn't fit our single-axis / complex-dtype / arbitrary-leading-dim case (I verified this - docstring on `_SlideWindow` explains).

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Correction to my earlier reply: actually addressed in bfb83e4b. `windows[..., :: self.dilation]` is now a dedicated `_TakeEvery` nn.Module, used as `self.dilation_selector` in `_Window.forward`. Forward is now a flat sequence of layer calls.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Correction to my earlier reply: actually removed in bfb83e4b. `_as_layer_list` is gone as a top-level helper; the broadcast+validate logic is now a local closure inside `_conformer_encoder` (single call site). `stride` keeps its own inline branch because its broadcast rule is different (scalar applies only to the last block).

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Second pass in 40583365: you were right, `_SlideWindow` was just `torch.nn.Unfold` wearing a custom name. Replaced with the real builtin: `self.frame = torch.nn.Unfold(kernel_size=(1, window_size), stride=(1, window_step))` in the MPF featurizer, with an einops `Rearrange` folding the un-pack into the same step (replacing the separate `csd_input_layout`). `_Window` keeps `tensor.unfold` inline because `nn.Unfold` there would need flatten+reshape+permute for variable trailing dims, which is a net loss.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Second pass in 40583365: `_TakeEvery` removed. `_Window` already owns the custom causal-padding logic and the `lpad`-mode branching, so the `windows[..., :: self.dilation]` stays inline inside its forward. One class, not three.

braindecode/models/generic_neuromotor_interface.py
bruAristimunha: Second pass in 40583365: `_as_layer_list` is removed as a top-level helper; the broadcast+validate logic is now a local closure inside `_conformer_encoder` (single call site).

braindecode/models/meta_neuromotor.py
```diff
@@ -0,0 +1,1803 @@
...
+from collections import OrderedDict
+from collections.abc import Sequence
+from typing import Literal, cast
+
+import torch
+import torch.nn.functional as F
+from einops import pack, rearrange, reduce, repeat, unpack
+from einops.layers.torch import Rearrange
+from torch import nn
+
```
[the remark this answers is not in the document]
bruAristimunha: no....

PR #998, opened 2026-04-24
pick channels before average reference in tuh example

### LINKED ISSUE #329: Order of preprocessing steps in TUH example is incorrect
In the TUH example, recordings are first re-referenced to CAR and afterwards channels are picked. 
Like this, artifacts in channels not part of the selection will be added to all other channels.
The order should be flipped, so first picking desired channels and re-referencing afterwards.

First reported by @Ann-KathrinKiessner

tayal-sarthak:
closes #329 

fix for the tuh example preprocessing order. right now set_eeg_reference with average runs before pick_channels, so the car reference is computed across every channel in the recording, including ones that get dropped a step later. any artifact on a dropped chan still leaks into the chans you keep through the subtracted mean.

swapped the order so pick_channels runs first, then the car reference is computed only on the chans we actually want. also flipped the matching bullet list in the markdown cell description to match the new order.

no api change, no behavior change anywhere else, just the example.

### DIFF
--- a/examples/applied_examples/plot_tuh_eeg_corpus.py
+++ b/examples/applied_examples/plot_tuh_eeg_corpus.py
@@ -282,7 +282,7 @@ preprocessors = [
     Preprocessor(
         custom_crop, tmin=tmin, tmax=tmax, include_tmax=False, apply_on_array=False
     ),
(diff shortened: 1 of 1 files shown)


bruAristimunha:
thanks @tayal-sarthak 🙏🏽

PR #1015, opened 2026-05-09
[ENH] EMG2QwertyNet: built-in SpecAugment + feature-extraction flags

bruAristimunha:
## Summary

Move emg2qwerty's SpecAugment from a Lightning callback into `EMG2QwertyNet` itself, and expose the encoder representation through two BIOT-style feature-extraction flags so downstream wrappers (e.g. neuroai's `DownstreamWrapperModel`) can drop in without a custom hook.

- **`spec_augment` (init kwarg, default `False`)** — installs a parameter-free `_SpecAugment` submodule between the log-spectrogram and BatchNorm. Active only in `train()`; default builds an `nn.Identity`, so state-dict layout and existing checkpoints are bit-identical. Knobs match the paper (Sivakumar et al. NeurIPS 2024, Sec 5.2): `n_time_masks=3`, `time_mask_param=25`, `n_freq_masks=2`, `freq_mask_param=4`, `spec_augment_prob=1.0`. Per-`(sample × band × electrode)` IID masking via `torchaudio.transforms.{Time,Frequency}Masking(iid_masks=True)` on the 5-D `(B, num_bands, electrodes, freq, T_spec)` tensor — same recipe as the upstream `emg2qwerty/transforms.py:SpecAugment` dataset transform. `mask_value` stays a 0-D on-device tensor (no host sync on GPU).
- **`return_features=True` (runtime kwarg)** → `{"features": (B, T_out, num_features), "cls_token": None}` dict, BIOT / signal-JEPA convention.
- **`return_feature=True` (init kwarg, BIOT legacy)** → default forward returns `(emissions, features)` tuple, so YAML-driven wrappers can pick up features via `model_output_key=1` without runtime kwargs.

## Test plan

`pytest test/unit_tests/models/test_models.py -k emg2qwerty` — 6 passing (4 pre-existing + 2 new):

- `test_emg2qwerty_spec_augment_contract` — eval determinism, train mutation + stochasticity, per-`(B, band, electrode)` IID (NOT band-shared, regression guard), zero state-dict footprint, constructor validation, `from_config` round-trip.
- `test_emg2qwerty_feature_flags` — all three forward paths (default tensor / runtime dict / init tuple), runtime-wins-over-init, `get_config` round-trip, `get_output_shape` invariant.

## Notes for reviewers

- All review comments resolved (5 threads). Codex's "band-shared masking" P2 was reverted after re-auditing upstream `emg2qwerty/transforms.py` — the paper recipe is per-electrode IID; band-shared was a 16× under-augmentation regression.
- `_SpecAugment` is `_`-prefixed and intentionally not re-exported — only the constructor flag is public, mirroring `_LogSpectrogram` / `_SpectrogramNorm`.
- No new dependencies; reuses `torchaudio.transforms` already imported for the STFT.

### DIFF
--- a/braindecode/models/emg2qwerty.py
+++ b/braindecode/models/emg2qwerty.py
@@ -568,11 +568,18 @@ class _SpecAugment(nn.Module):
     r"""SpecAugment masking on the log-spectrogram during training.
 
     Applies up to ``n_time_masks`` × ``time_mask_param``-frame time bands
-    and ``n_freq_masks`` × ``freq_mask_param``-bin frequency bands, IID
-    per ``(sample × band)`` and shared across the electrodes within a band
-    — per-electrode masking is :math:`\\sim 32\\times` more aggressive and
-    collapses CTC. No-op outside training. Defaults match Sivakumar et al.
-    Sec 5.2 (3×25 / 2×4 frames-bins, ``prob=1.0``).
+    and ``n_freq_masks`` × ``freq_mask_param``-bin frequency bands, with
+    one mask drawn per ``(sample × band)`` row and broadcast across all
+    electrodes within that band — so a wristband's 16 electrodes share
+    the same masked window. No-op outside ``training``. Defaults match
+    Sivakumar et al. Sec 5.2 (3×25 / 2×4 frames-bins, ``prob=1.0``).
+
+    Implemented as pure tensor ops (per-row vectorised start / width
+    sampling, on-device mean fill, ``masked_fill`` broadcast) so the
+    forward pass keeps no host round-trips on GPU. ``torchaudio``'s
+    ``TimeMasking(iid_masks=True)`` cannot be reused here because, on a
+    ``(B*num_bands, electrodes, freq, T)`` tensor, it samples one mask
+    per ``(batch, electrode)`` pair instead of per ``(sample × band)``.
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


bruAristimunha:
Thanks for the reviews — pushed `b355eb2b` addressing all three:

**@chatgpt-codex (P2: per-electrode masks)** — Confirmed by direct check: `torchaudio.transforms.TimeMasking(iid_masks=True)` on a `(B*num_bands, electrodes, freq, T)` 4-D tensor samples one mask per `(batch, electrode)` pair, so each of the 16 electrodes in a band drew its own independent mask. The comment and the original neuralbench callback were both wrong about the semantics. Replaced the torchaudio path with manual band-shared masking (`_band_shared_axis_mask` static method): per-row vectorised width / start sampling, `masked_fill` broadcasts the mask across the electrode axis. Pinned by a new contract assertion in `test_emg2qwerty_spec_augment_train_only` that builds a uniquely-valued tensor and verifies every electrode within each `(B, band)` row sees the *exact same* mask pattern.

**@Copilot (`.item()` GPU sync)** — Mask value is now `flat.mean()` (0-D on-device tensor) and `masked_fill` accepts it directly. No more host round-trip per training forward.

**@Copilot (seed-dependent test)** — The train-mode regression now monkeypatches `torch.randint` for the size-`()` calls so both `n_t` and `n_f` come back as 2 deterministically, regardless of global RNG / torchaudio versions.

Also added a what's-new entry covering both built-in SpecAugment and the `return_features` flag from the previous commit.

bruAristimunha:
Self-correction on Codex's earlier P2 — pushed \`55b78ec1\`.

Re-audited the upstream \`emg2qwerty/transforms.py\` to confirm the actual paper recipe before locking in the band-shared rewrite. Findings:

\`\`\`python
# emg2qwerty/transforms.py — applied at the dataset level on a SINGLE sample
def __call__(self, specgram: torch.Tensor) -> torch.Tensor:
    # (T, ..., C, freq) -> (..., C, freq, T)
    x = specgram.movedim(0, -1)
    ...
    x = self.time_mask(x, mask_value=self.mask_value)  # iid_masks=True
\`\`\`

The single-sample input is \`(T, bands, electrodes, freq)\`. After \`movedim(0, -1)\` it's 4-D \`(bands, electrodes, freq, T)\`, and \`torchaudio\`'s \`iid_masks=True\` samples one mask per leading axis pair → per \`(band, electrode)\`. **Each electrode in a band gets its own mask.** Verified empirically too. The previous neuralbench callback's \`(B*num_bands, electrodes, freq, T)\` reshape produced the same per-\`(sample × band × electrode)\` independence; only its comment was wrong.

So the band-shared "fix" was a regression — it made the augmentation 16× *less* aggressive than the paper. Reverted to per-\`(sample × band × electrode)\` iid masking via a direct \`torchaudio\` call on the 5-D tensor (\`mask_along_axis_iid\` draws one mask per leading axis triple). Kept the on-device \`mask_value\` (no host sync) and the deterministic \`torch.randint\` monkeypatch in the test. Test contract is now an assertion that masking is **NOT** band-shared.

Also fixed a separate issue surfaced while double-checking the \`return_features\` integration with neuroai. \`DownstreamWrapper.build\` calls \`model(**dummy_batch)\` where \`dummy_batch\` only contains the first forward parameter — it does **not** pass \`return_features=True\` at runtime. So the runtime dict flag was inert under neuroai's YAML-driven flow. Added a \`return_feature\` (singular, init kwarg, BIOT-style legacy) that makes \`forward\` return a \`(emissions, features)\` tuple by default, so a config such as \`kwargs: {return_feature: true}\` + \`model_output_key: 1\` works out of the box. Runtime \`return_features=True\` still wins for programmatic users.

Apologies @chatgpt-codex-connector — your concern that masking was per-\`(B × electrode)\` was *factually correct*, but the conclusion that the paper is band-shared isn't supported by the upstream code. The paper recipe is per-electrode independent, and that's what's in this commit.

bruAristimunha:
Pushed \`d02ac8af\` addressing all three @copilot-pull-request-reviewer comments:

1. **Lazy \`features\` materialisation** — \`forward()\` no longer transposes \`encoded\` on the default emissions-only path; the transpose + contiguous copy now happens only inside the two branches that actually return the features tensor. Default path is back to pre-flag behaviour.
2. **Device-aware RNG** — both the prob gate (\`torch.rand((), device=x.device)\`) and the mask-count draws (\`torch.randint(n+1, (), device=x.device)\`) now use \`x.device\`. Single seed (\`torch.manual_seed\` or \`torch.cuda.manual_seed\`) reproduces the augmentation, and torchaudio's internal device-side RNG stays in the same stream as our Python-level gate. \`.item()\` is still required to bind the Python \`if\`/\`range\`, so a host sync remains for control flow only.
3. **Changelog ↔ PR description alignment** — the PR description was rewritten in a previous step (the old "shared across electrodes within a band" wording was a stale draft from the band-shared revert). Both the description and \`whats_new.rst:45\` now say per-\`(sample × band × electrode)\` IID.

PR #1020, opened 2026-05-14
Adding ZUNA to Braindecode

jonathanhuml:
**SUMMARY**

We propose to add ZUNA, a masked diffusion autoencoder trained to perform masked channel infilling and superresolution for arbitrary electrode numbers and positions in EEG signals. 

While the original encoder-decoder model has 380M parameters, this port is a feature extractor that only exposes the latents and does not perform reconstruction. The implementation also includes basic support for channel masking and dropped-channel inference, either by channel index or by channel name. This is intended to preserve some of the practical behavior expected from a model trained around masked channel infilling.
By discarding the decoder, the total model size is about 170M parameters. We are totally happy to add the decoder back in later: let us know what BrainDecode could most benefit from! We did this to keep the file as lightweight and readable as possible. We are currently training a new version and will likely submit another PR soon, so we can definitely integrate this into the next version if so desired. 

We have tried to keep as close as possible to the requirements and [base.py](http://base.py/) structure in Braindecode. We have two main design decisions that would probably be helpful to highlight for the Braindecode team for any potential feedback:
ZUNA currently depends on PyTorch `flex_attention`. This is only available in PyTorch `>=2.5`, while Braindecode currently supports PyTorch `>=2.0`. To avoid making ZUNA break imports for users on older PyTorch versions, this PR uses a soft import pattern similar to the Hugging Face integration. As a result, Braindecode can still be imported normally, but instantiating or running ZUNA requires a compatible PyTorch version. ZUNA could theoretically support other attention variants at the cost of less efficient GPU usage, but we have implemented the flex-only variant for now
ZUNA is montage-agnostic but requires 3D channel positions during the forward pass. These positions are not inherent to the model weights, so this PR allows users to provide them dynamically rather than binding them at model  initialization. When `chs_info` contains the necessary position metadata, we also support extracting the positions from there.


bruAristimunha:
I am mostly struggling to get some results a little above the change level with this model @jonathanhuml, any input is welcome. I don't know what to test more.

jonathanhuml:
> I am mostly struggling to get some results a little above the change level with this model @jonathanhuml, any input is welcome. I don't know what to test more.

Hey @bruAristimunha! I appreciate you testing this out, sorry it's been a pain. The 5 second trial limitation is definitely a major constraint here, we are in the midst of a new training run that relaxes this assumption to arbitrary trial lengths (up to 30 seconds) and the preprocessing in general, as well as diversifying the tasks. Could you give me some time (at least a week) to chop the new version up into braindecode's API? We are seeing much better results in Meta's new NeuralBench repo with this updated version. This seems similar to Open EEG Bench so hopefully that should generalize.  I will send you a message on Discord next week to update you on my progress if that works

bruAristimunha:
Works for me @jonathanhuml!

jonathanhuml:
> Works for me @jonathanhuml!

Thank you @bruAristimunha!

jonathanhuml:
I'm coming back to this @bruAristimunha! Sorry for the delay here, we had a lot of things to catch up on with our model release.

bruAristimunha:
Hey @jonathanhuml, once everything is green! I am happy to merge, many thanks for all the effort here!

PR #1026, opened 2026-05-21
[ENH] EEGModuleMixin.load_state_dict: drop & report shape mismatches on strict=False

ArseniiB-o:
## Summary

When a user does `model.load_state_dict(checkpoint, strict=False)` and at
least one tensor in the checkpoint has a shape that no longer matches
the current model, PyTorch ≥ 2.1 raises `RuntimeError` instead of
silently re-initialising the offending tensor (pytorch/pytorch#92344).
The result is that `strict=False` is no longer a one-liner for "load
the pretrained backbone and let the head be re-initialised."

The most concrete case in the wild is `InterpolatedLaBraM`:
`temporal_embedding` is tied to the pretraining sequence length, so any
fine-tune with a different epoch length fails to load via `strict=False`
unless the user manually pre-filters the state-dict.

This PR keeps `strict=True` behaviour untouched (still raises) but for
`strict=False`:

1. drops shape-mismatched tensors *before* `super().load_state_dict()`
   sees them, so PyTorch does not raise,
2. emits a single `logging.warning` listing the dropped keys and the
   checkpoint-vs-model shapes, and
3. surfaces the dropped names via the standard
   `_IncompatibleKeys.missing_keys` namedtuple — the corresponding model
   parameters keep their freshly-initialised values.

The new override is also defensive about non-tensor entries in the
state-dict (some downstream modules stash metadata there): if an entry
has no `.shape` attribute, the shape-check pass leaves it for PyTorch
to handle, rather than raising `AttributeError`.

## Motivation

Downstream BCI Comp IV 2a fine-tune of `InterpolatedLaBraM` (full,
50 epochs CPU, single seed): mean accuracy **0.30** / κ=0.07 across 9
subjects — worse than every from-scratch deep model and every classical
baseline. The root cause was the silent re-init of
`temporal_embedding` under a pre-2.1 `strict=False`; on PyTorch 2.1+ the
same fine-tune now fails to load at all without a hand-written pre-
filter. Write-up:
https://github.com/ArseniiB-o/eeg-signal-classifier/blob/main/eeg-bci/docs/10_results.md

The hand-written workaround in that project
(`eegbci.models.foundation.LaBraM._load_pretrained_weights`) is exactly
the logic this PR moves into `EEGModuleMixin` so every braindecode
model benefits.

## Test plan

New tests in `test/unit_tests/models/test_base.py` exercise:

- [x] `strict=False` drops shape-mismatched tensors and emits the
      warning; the dropped key shows up in `missing_keys`.
- [x] `strict=True` still raises `RuntimeError("size mismatch")`.
- [x] When all shapes match, no warning is emitted and
      `missing_keys` / `unexpected_keys` are empty.
- [x] Positional `model.load_state_dict(sd, False)` behaves the same
      as the keyword form.
- [x] Existing `mapping` remapping still applies before the shape check.
- [x] Checkpoint keys absent from the model still go to
      `unexpected_keys` (PyTorch contract preserved).
- [x] Multiple shape mismatches are reported in **one** combined
      warning, not N separate warnings.
- [x] `register_buffer()` tensors follow the same drop semantics as
      parameters.
- [x] Omitting `strict` keeps the default `True` and still raises.

Verified locally on Windows 11, Python 3.13, PyTorch 2.12:

* `pytest test/unit_tests/models/test_base.py -v` → **49 passed**.
* `pytest test/unit_tests/models/` (full models suite) →
  **1111 passed, 209 skipped, 0 failed** (7m 21s).
* `ruff check` introduces **0 new errors** (the pre-existing 7 E402
  warnings on `base.py` are unchanged on this branch).

## Behavioural compatibility

- `strict=True` — unchanged.
- `strict=False` and *no* shape mismatches — unchanged (no warning is
  emitted in this case).
- `strict=False` *with* shape mismatches that used to raise on PyTorch
  ≥ 2.1 — now silently loads what fits, plus a warning. This is a
  behaviour *unblock*, not a behaviour break.

## Docs

- `docs/whats_new.rst`: a new bullet under *API and behavior changes*
  for 1.6.0dev points to pytorch/pytorch#92344 and credits the
  contributor.
- The `load_state_dict` docstring now has dedicated *Parameters* /
  *Returns* sections plus a `.. versionchanged:: 1.6` directive that
  Sphinx will surface in the rendered API reference.

## Notes

- The warning text includes a hint about
  `temporal_embedding` / LaBraM as the most common cause, since that
  is the only model in the current model zoo where this is expected
  to fire under common usage. Happy to soften the wording if
  maintainers prefer something more general.
- The new module-level `log = logging.getLogger(__name__)` follows
  the same pattern as `braindecode/preprocessing/*.py`.
- No existing test was modified; only new tests were added.

### DIFF
--- a/braindecode/models/base.py
+++ b/braindecode/models/base.py
@@ -6,10 +6,13 @@
 from __future__ import annotations
 
 import json
+import logging
 import warnings
 from collections import OrderedDict
 from pathlib import Path
 from typing import Dict, Iterable, Optional, Type, Union
+
+log = logging.getLogger(__name__)
 
 import numpy as np
 import torch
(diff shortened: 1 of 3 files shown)


bruAristimunha:
hey @ArseniiB-o!

Many thanks! The pytorch issue doesn't seem related. Can you please take a look? And can you please create an MVC (15 lines max) that reproduces the issue?

adammounir:
Hi @ArseniiB-o, thanks a lot for the detailed PR! After review, we don't think this should land, for two main reasons:
  
  - The premise doesn't hold. strict only governs missing/unexpected keys, never shape mismatches, those have always raised, with strict=True and False alike (it's by
  design in _load_from_state_dict). So this isn't a PyTorch ≥ 2.1 regression, and #92344 describes a different case.
  - braindecode already covers this use case: the Interpolated* family (InterpolatedLaBraM, InterpolatedEEGPT, …) interpolates sequence-length-tied weights
  instead of dropping them and re-initialising at random, which is what we'd recommend for the BCI IV 2a fine-tune.

  Closing for now, but please reopen if you have a minimal repro of a case Interpolated* doesn't cover. Thanks again for you contribution! 🙂

Issue #545, opened 2023-09-27
Making  Unit Tests Faster by Using Simulated Data


robintibor:
Right now in several unit tests we load the BCI Competition  IV 2a Dataset to then test windowing serialization etc. 
In my view, this is an overkill and uses a lot of memory and runtime unnecessarily. We can just create smaller entirely fake data in those tests and this will speed up the unit tests quite a lot. Also makes it more appropriate for unit tests, like will not break in case anything changes about the `MOABBDataset`.

Affected files under `test/unit_tests`:
* `datautil/test_serialization.py`
* `datasets/test_dataset.py`
* `preprocessing/test_mne_preprocessor.py
* `preprocessing/test_preprocess.py`
* `preprocessing/test_windowers.py`
* `samplers/test_samplers.py`
* `training/test_scoring.py`

bruAristimunha:
We have a util function called: https://github.com/braindecode/braindecode/blob/ab87bbba71719b122e8b764259854c80974d9d73/braindecode/util.py#L268

maybe a easy solution

adammounir:
Thanks @robintibor for raising this back in 2023.

After discussion we've decided it may preferable to keep real data here to guarantee a stronger contract within the ecosystem: integration-style tests on real recordings catch real-world edge cases that synthetic signals tend to miss

Also the related PRs (#1035, #1036) have been closed accordingly. Once again, many thanks to @YihengLi-1 for the effort on them.

PR #1035, opened 2026-05-30
test: use synthetic data instead of downloading a real dataset in sampler tests (#545)

YihengLi-1:
## Summary

Partially addresses #545. The sampler unit tests (`test/unit_tests/samplers/test_samplers.py`) downloaded the real `BNCI2014_001` MOABB dataset (subject 4) in their `windows_ds` and `target_windows_ds` fixtures. These tests only depend on the dataset *structure* — the number of recordings, the number of windows, and the `subject`/`session`/`run`/`target` metadata columns — not on real EEG signals, so the download is unnecessary.

This replaces the two fixtures' data source with small synthetic data built from the existing `braindecode.util.create_mne_dummy_raw` helper (the util pointed to in the issue). The windowing calls and all test logic are left unchanged.

## Why this is safe

The synthetic data reproduces the structure the samplers rely on:

- 3 recordings with distinct `subject`/`session`/`run`, so `RecordingSampler` groups them into the expected number of recordings;
- recordings long enough (52 s) that the fixed-length windows span the `tau_pos`/`tau_neg` range used by `RelativePositioningSampler`;
- 48 trials per recording across 2 alternating classes, so `BalancedSequenceSampler` has enough consecutive windows and at least two target classes (this covers the `n_windows=40` parametrization).

## Result

- All 19 tests in the file pass (12 non-distributed + 7 `mp.spawn`/gloo distributed).
- The data-loading cost is gone, and the tests no longer require network access or `moabb`. For example, `test_recording_sampler` went from **~11.5 s** (real download) to **~3.0 s** (synthetic; the remainder is just import overhead).

Scope: only `test/unit_tests/samplers/test_samplers.py`. The other files listed in #545 can follow in separate PRs.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -66,6 +66,11 @@ Code health
   pull request for the version bump, so it never needs to push to the
   protected ``master`` branch. (:gh:`1031` by `Bruno Aristimunha`_)
 
+- Speed up the sampler unit tests by generating small synthetic recordings
+  with :func:`braindecode.util.create_mne_dummy_raw` instead of downloading
+  the ``BNCI2014_001`` dataset, removing the network dependency from
+  ``test/unit_tests/samplers/test_samplers.py``. (:gh:`1035` by `YihengLi-1`_)
+
 
 Current 1.5.1 (stable)
 ===============================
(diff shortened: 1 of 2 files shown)


adammounir:
hi @YihengLi-1! Thank you again for this suggestion, as for the PR #1036 and its associated issue, we prefer keeping real data for integration tests to guarantee a stronger contract within the ecosystem, as simulated data misses real-world edge cases.

PR #1036, opened 2026-05-30
test: use synthetic data instead of downloading a real dataset in windower tests (#545)

YihengLi-1:
## Summary

Partially addresses #545 (follow-up to #1035). The `concat_ds_targets` fixture in `test/unit_tests/preprocessing/test_windowers.py` was the last fixture in that file still downloading the real `BNCI2014_001` MOABB dataset — `_get_raw`, `concat_ds_targets_overlap`, and `lazy_loadable_dataset` already use synthetic data.

The ~11 tests that use this fixture only depend on the trial *geometry* (250 Hz sampling, four motor-imagery classes, uniform 4 s / 1000-sample trials) and compare each window's label against `targets` derived from the same annotations (self-consistent), so they do not need real EEG signals.

This replaces the fixture with synthetic data built from `mne.io.RawArray` + `mne.Annotations`, reproducing the structure the tests rely on. The class names match BNCI so `mne.events_from_annotations` assigns the same event ids (1..4) and `targets = events - 1` exactly as before, and the first-trial onset leaves room for the negative `trial_start_offset_samples` used by some tests. The windowing calls and all test logic are unchanged.

## Result

- All 87 tests in the file pass (1 pre-existing skip), with no changes to any test body.
- `test_one_window_per_original_trial` went from ~4.4 s (real download, warm local cache — a cold CI cache is much slower) to ~2.4 s, and the file no longer needs network access or `moabb`.

Scope: only `test/unit_tests/preprocessing/test_windowers.py`. The remaining files in #545 can follow in separate PRs.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -66,6 +66,12 @@ Code health
   pull request for the version bump, so it never needs to push to the
   protected ``master`` branch. (:gh:`1031` by `Bruno Aristimunha`_)
 
+- Speed up the windower unit tests by generating small synthetic recordings
+  with :func:`braindecode.util.create_mne_dummy_raw` instead of downloading
+  the ``BNCI2014_001`` dataset, removing the network dependency from
+  ``test/unit_tests/preprocessing/test_windowers.py``. (:gh:`1036` by
+  `YihengLi-1`_)
+
 
 Current 1.5.1 (stable)
 ===============================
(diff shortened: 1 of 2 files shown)


adammounir:
hi @YihengLi-1! Thank you for this suggestion, however we prefer keeping real data for integration tests to guarantee a stronger contract within the ecosystem, as simulated data misses real-world edge cases.

PR #1037, opened 2026-05-30
fix: apply safe standalone MNE preprocessors via callable (#885)

YihengLi-1:
## Summary

Addresses #885 for the subset of standalone-function preprocessors that can be fixed safely.

The auto-generated standalone-function preprocessors in `mne_preprocess.py` passed the function **name (a string)** to `Preprocessor` instead of the **callable**, so they never actually applied their wrapped MNE function. #885 lists 19 affected classes.

However, a blanket "pass the callable" fix is unsafe. When `fn` is a callable and `apply_on_array=False`, `Preprocessor.apply` reassigns the dataset's recording to whatever `fn` returns. Several of these MNE functions return **auxiliary data** rather than the modified recording — e.g. `annotate_amplitude` returns a `(annotations, bads)` tuple, `find_bad_channels_lof` returns a list, `set_eeg_reference` (standalone) returns a `(raw, ref_data)` tuple. Wrapping those as callables would silently replace the recording with that auxiliary object — worse than the current failure.

## What this PR does

- Introduces a small, explicit allow-list `_SAFE_STANDALONE_FUNCTIONS` of standalone functions **verified to return the modified instance**, and routes only those through the callable path. Currently three functions:
  - `ComputeCurrentSourceDensity` (`compute_current_source_density`)
  - `SetBipolarReference` (`set_bipolar_reference`)
  - `OversampledTemporalProjection` (`oversampled_temporal_projection`)
- Every other standalone function keeps its existing behaviour unchanged.

I verified the classification empirically (constructing each class, applying it to a Raw, and checking the return type). The remaining standalone functions fall into:
- **Return auxiliary data** (tuple/list/Annotations) — unsafe to pass as a callable as-is: `annotate_amplitude`, `annotate_nan`, `compute_bridged_electrodes`, `find_bad_channels_lof`, standalone `set_eeg_reference`, …
- **Need a second instance / list input** (don't fit the single-recording preprocessor model): `equalize_bads`, `equalize_channels`, `realign_raw`.
- **Operate on arrays** (`filter_data`) — already documented to use `Filter` instead.

These need their own handling/design and are intentionally out of scope here.

## Tests

- `test_safe_standalone_preprocessor_uses_callable` — parametrized over the three safe classes: asserts `.fn` is the callable (not a string) and that `apply` returns a `Raw`.
- `test_safe_standalone_preprocessor_runs_in_pipeline` — runs `ComputeCurrentSourceDensity` end-to-end through `preprocess`.
- `test_unsafe_standalone_preprocessor_keeps_string_fn` — guard: `AnnotateAmplitude.fn` stays the string `"annotate_amplitude"`, so a future change can't silently turn an auxiliary-data function into a data-corrupting callable.
- Un-skipped the existing `test_compute_csd`, which now passes.

Full file: 109 passed, 14 skipped, 0 failed.

> **Scope note:** this intentionally fixes only the 3 safe functions; the remaining 16 are tracked in #1055. Using `Addresses` (not `Fixes`) so #885 is not auto-closed on merge.


bruAristimunha:
@adammounir, did you have the chance to replicate the error and review this PR?

bkowshik:
Hi @YihengLi-1 @bruAristimunha — I took an independent look while reading around #885. A few data points that might help move this forward:

**Reproduced on current `master`** — minimal repro:
```python
import mne, numpy as np
from braindecode.preprocessing import FilterData
info = mne.create_info(["Cz", "Pz"], sfreq=250, ch_types="eeg")
raw = mne.io.RawArray(np.random.randn(2, 2500), info, verbose=False)

fd = FilterData(sfreq=250, l_freq=1, h_freq=40)
print(FilterData.fn)   # <function filter_data at 0x...>  -> class attr is the callable (correct)
print(fd.fn)           # 'filter_data'                    -> instance attr shadows it (the bug)
fd.apply(raw)          # AttributeError: MNE object does not have a filter_data method.
```
So the root cause is exactly as described: the generated `__init__` passes `fn=func_name` (the string), which shadows the correct class-level callable, and standalone functions then get routed down the `_apply_str` "string = Raw method name" path.

**One small thing:** since this intentionally fixes 3 of the 19 (the rest split into separate categories — auxiliary-data returns, second-instance inputs, array funcs — each needing its own design), the `Fixes #885` will auto-close the issue on merge. Might be worth "Addresses #885" + a tracking issue so the remaining buckets aren't lost.

Happy to pick up a follow-up for one of the remaining categories (the auxiliary-data group looks like the meatiest) if useful.

adammounir (approved):
Thanks @YihengLi-1,

I reproduced the bug on master and validated this branch locally:  

The 3 safe classes now expose a callable `fn`, while the auxiliary-data ones (`FilterData`, `AnnotateAmplitude`, `FindBadChannelsLof`) correctly stay on the string path, safely guarded by `test_unsafe_standalone_preprocessor_keeps_string_fn`. 

This conservative allow-list + guard-test approach is good, as it avoids the footgun of silently replacing a recording with a tuple or list.

Picking up @bkowshik 's point, since this fixes 3 of the 19, I changed the description from *Fixes #885* to *Addresses #885* and opened **#1055** to track the rest so #885 isn't auto-closed on merge. LGTM, thanks for the contribution !

PR #1039, opened 2026-06-08
fix pydantic serialize_as_any config dumps

### LINKED ISSUE #892: Pydantic v2 serialization failure in Braindecode model configs
Braindecode's Pydantic model configuration classes (e.g., `EEGNetConfig`, `EEGConformerConfig`) fail to serialize with Pydantic v2's JSON mode (`model_dump(mode="json", serialize_as_any=True)`), raising:

```
PydanticSerializationError: Unable to serialize unknown type: <class 'type'>
```

**Root cause**: The `activation` field in model configs stores a bare Python `type` object (e.g., `<class 'torch.nn.modules.activation.ELU'>`), which cannot be serialized to JSON.

This breaks any downstream tooling that relies on Pydantic's standard JSON serialization, such as:
- **Exca** (experiment caching framework)
- Any configuration management that needs JSON serialization
- Saving/loading configs to/from JSON files

## Minimal Reproduction

The issue can be reproduced without any external dependencies beyond Braindecode and Pydantic:

```python
from braindecode.models.config import EEGNetConfig

# Create a Braindecode model config
signal_kwargs = {"n_times": 1000, "n_chans": 26, "n_outputs": 4}
model_cfg = EEGNetConfig(**signal_kwargs)

# This fails with PydanticSerializationError
result = model_cfg.model_dump(mode="json", serialize_as_any=True)
```

**Error output:**
```
pydantic_core._pydantic_core.PydanticSerializationError: Unable to serialize unknown type: <class 'type'>
```

**Investigation:**
```python
# The problematic field:
cfg = EEGNetConfig(n_times=1000, n_chans=26, n_outputs=4)
d = cfg.model_dump()
print(d['activation'])  # <class 'torch.nn.modules.activation.ELU'>
print(type(d['activation']))  # <class 'type'>
```

The `activation` field contains a bare `type` object that cannot be serialized in JSON mode.

## Full Reproduction (with Exca)

When using Braindecode configs with Exca (as shown in the official `plot_exca_config.ipynb` example), the error occurs when Exca tries to serialize the configuration for caching:

```python
import exca
import pydantic
from braindecode.models.config import BraindecodeModelConfig, EEGNetConfig

class TrainingConfig(pydantic.BaseModel):
    model_config = pydantic.ConfigDict(extra="forbid")
    infra: exca.TaskInfra = exca.TaskInfra(folder=".cache/", cluster=None)
    model: BraindecodeModelConfig
    # ... other fields

    @infra.apply
    def train(self):
        # ... training logic
        pass

# Instantiate and call
signal_kwargs = {"n_times": 1000, "n_chans": 26, "n_outputs": 4}
model_cfg = EEGNetConfig(**signal_kwargs)
train_cfg = TrainingConfig(model=model_cfg, train_dataset=...)

train_cfg.train()  # Fails during Exca's config serialization step
```

**Stacktrace:**
```
File ".../site-packages/exca/utils.py", line 89, in ConfigExporter.apply
    out = model.model_dump(
        exclude_defaults=self.exclude_defaults,
        mode="json",
        serialize_as_any=True,
    )
File ".../site-packages/pydantic/main.py", line 464, in BaseModel.model_dump
    return self.__pydantic_serializer__.to_python(...)
PydanticSerializationError: Unable to serialize unknown type: <class 'type'>
```

## Expected Behavior

Braindecode's Pydantic configuration classes should serialize cleanly with:

```python
model_cfg.model_dump(mode="json", serialize_as_any=True)
```

This should work because:
1. Braindecode configs are advertised as serializable Pydantic models
2. The documentation shows examples of serializing configs (e.g., `model_dump(mode="json")`)
3. JSON serialization is a standard requirement for configuration management

## Actual Behavior

Serialization fails with `PydanticSerializationError: Unable to serialize unknown type: <class 'type'>` because the `activation` field (and potentially other fields) contains bare `type` objects that cannot be converted to JSON.

## Environment

- **OS**: macOS (darwin 24.6.0)
- **Python**: 3.12.11
- **Braindecode**: 1.3.0
- **Pydantic**: 2.12.5
- **Exca**: 0.5.8 (for full reproduction)
- **Installation method**: `uv` package manager

## Suggested Fix

The `activation` field (and any other fields storing `type` objects) should be serialized as a string representation that can be reconstructed. Options:

1. **Use string representation with custom serializer**:
   ```python
   from pydantic import field_serializer
   
   @field_serializer('activation', mode='wrap')
   def serialize_activation(self, value, handler):
       if isinstance(value, type):
           return f"{value.__module__}.{value.__name__}"
       return handler(value)
   ```

2. **Store as string by default** and use a validator to convert to type when needed:
   ```python
   activation: str = "torch.nn.modules.activation.ELU"
   
   @property
   def activation_class(self) -> type:
       module_path, class_name = self.activation.rsplit('.', 1)
       module = __import__(module_path, fromlist=[class_name])
       return getattr(module, class_name)
   ```

3. **Exclude from JSON serialization** and handle separately (less ideal, breaks serialization contract).

The first option is likely the cleanest, as it maintains backward compatibility while enabling JSON serialization.

## Additional Context

- This affects all model configs that use the `activation` field (likely `EEGNetConfig`, `EEGConformerConfig`, and others).
- The issue only manifests with `mode="json"`; `model_dump()` without `mode="json"` works but produces non-JSON-serializable output.
- The error message in the notebook output shows `'activation': 'torch.nn.modules.activation.ELU'` as a string, suggesting there may already be some serialization logic, but it's not working with `serialize_as_any=True`.

## Related

This issue prevents the official Braindecode + Exca example (`plot_exca_config.ipynb`) from working as documented, as Exca requires JSON-serializable configurations for its caching mechanism.

tayal-sarthak:
summary

fixes #892 

pydantic json dumps with serialize_as_any=true failed when model configs held torch activation classes. importstring handled normal json mode, serialize_as_any exposed the raw type value.

adds a json wrap serializer on base braindecode config. importable classes and functions dump as dotted paths. other fields still go through pydantic's default handler.

closes #892

### DIFF
--- a/braindecode/models/config.py
+++ b/braindecode/models/config.py
@@ -1,6 +1,6 @@
 # Authors: Sarthak Tayal <sarthaktayal2@gmail.com>
 from collections.abc import Callable
-from inspect import signature
+from inspect import isbuiltin, isfunction, signature
 from types import UnionType
 from typing import Annotated, Any, Literal, Union, get_args, get_origin
 
@@ -63,6 +63,14 @@ SIGNAL_ARGS_TYPES = {
 
 class BaseBraindecodeModelConfig(pydantic.BaseModel):
     """Base class for braindecode model pydantic configs."""
+
+    @pydantic.field_serializer("*", mode="wrap", when_used="json", check_fields=False)
+    def _serialize_import_strings(
+        self, value: Any, handler: Callable[[Any], Any]
+    ) -> Any:
+        if isinstance(value, type) or isbuiltin(value) or isfunction(value):
+            return f"{value.__module__}.{value.__qualname__}"
+        return handler(value)
 
     def create_instance(self) -> EEGModuleMixin:
         if self.model_name_ not in models_dict:
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -50,6 +50,11 @@ Requirements
 Bug fixes
 ==========
 
+- Fix JSON serialization of the braindecode model pydantic configs: type,
+  function, and builtin field values (such as ``activation``) are now dumped
+  as importable dotted paths (e.g. ``"torch.nn.modules.activation.ELU"``) so
+  that ``model_dump(mode="json", serialize_as_any=True)`` round-trips back
+  through ``model_validate`` (:gh:`1039` by `Sarthak Tayal`_)
 - Fix :class:`~braindecode.models.ContraWR`,
   :class:`~braindecode.models.DeepSleepNet`, and
   :class:`~braindecode.models.EEGMiner` raising a confusing
--- a/test/unit_tests/models/test_config.py
+++ b/test/unit_tests/models/test_config.py
@@ -62,6 +62,19 @@ def test_make_model_config_json_serialization(model_name, required, signal_param
         )
 
 
+def test_make_model_config_json_serialization_as_any():
+    from braindecode.models.config import EEGNetConfig
+
+    cfg = EEGNetConfig(n_times=1000, n_chans=26, n_outputs=4)
+    serialized = cfg.model_dump(mode="json", serialize_as_any=True)
+    cfg_from_serialized = EEGNetConfig.model_validate(serialized)
+
+    assert serialized["activation"] == "torch.nn.modules.activation.ELU"
+    np.testing.assert_equal(
+        cfg_from_serialized.model_dump(mode="python"), cfg.model_dump(mode="python")
+    )
+
+
 @pytest.mark.parametrize(
     "n_times, input_window_seconds, sfreq",
     [


PierreGtch:
Hi @adammounir and @tayal-sarthak 

why do you need to use serialize_as_any=true? 

The issue #892 was initially raised because exca was forcing serialize_as_any=true. But this issue has been solved on the exca side in  https://github.com/facebookresearch/exca/issues/165  and the related PR

If there is no concrete use case when serialize_as_any=true is necessary, I don't think we should hack pydantic

tayal-sarthak:
Sounds good, you are right - thank you for your review @PierreGtch ! :)

PR #1043, opened 2026-06-10
Add BrainOmni + BrainTokenizer (unified EEG/MEG foundation model)

bruAristimunha:
## Summary

Adds two models derived from **BrainOmni: A Brain Foundation Model for Unified
EEG and MEG Signals** (Xiao et al., NeurIPS 2025;
[arXiv:2505.18185](https://arxiv.org/abs/2505.18185); upstream
[OpenTSLab/BrainOmni](https://github.com/OpenTSLab/BrainOmni), MIT):

- **`BrainTokenizer`** — the SEANet/channel-compression/residual-VQ tokenizer,
  registered as a non-classification model.
- **`BrainOmni`** — the downstream classifier: frozen tokenizer inference,
  factored spatio-temporal transformer, released mean-pooling path, and
  classification head.

Key contracts:

- Strict public loaders accept the authors' raw tokenizer, tiny, and base
  checkpoints. Immutable artifact revisions and SHA-256 values are pinned in
  the integration tests; no checkpoint is copied or re-hosted by this PR.
- Pinned-source comparisons cover tokenizer initialization/quantization and
  released numerical signatures. The deterministic RoPE cache is regenerated
  from its authoritative frequency buffer because the DeepSpeed export lost
  its complex phase.
- Geometry is derived once from `chs_info` for EEG, MAG, planar GRAD, and axial
  GRAD. Real MNE types use public `mne.channel_type`; lightweight `ch_type`
  metadata and real-MNE metadata both round-trip through config and local Hub
  serialization with exact state/output parity.
- Fresh codebooks use the released first-batch K-means initialization.
  Distributed EMA initialization and sufficient statistics are synchronized
  across ranks.
- Token extraction temporarily enters evaluation mode and restores the
  tokenizer's prior mode; projection, transformer blocks, and head remain
  trainable during downstream fine-tuning.
- The historical positional `PatchTokenizer` API remains valid, short and odd
  SEANet cases follow the source padding policy, invalid public configurations
  fail early, and the implementation remains compatible with the declared
  PyTorch 2.4 floor, using native RMSNorm directly.

The released source preprocessing differs from the paper's prose at its final
normalization step. Model documentation records that discrepancy and tells
users to reproduce the released source pipeline when using released weights.

## Provenance and licensing

- Audited upstream source commit: `340d6b5aba886af76b217272cdb3251651e9bf16`.
- Audited OpenTSLab artifact revision:
  `9a4d3c70495370397ccfbfd6d2496f25647545a5`.
- `LICENSES/BrainOmni-MIT.txt` preserves the OpenTSLab and Meta/EnCodec MIT
  grants; `LICENSES/vector-quantize-pytorch-MIT.txt` preserves the additional
  derived quantizer grant; `NOTICE.txt` records both components.
- Wheel and sdist gates verify that the project BSD license, NOTICE, and both
  MIT files are packaged. No external weight is bundled.

The official OpenTSLab repositories contain raw `.pt` state dictionaries, not
Braindecode `config.json` plus safetensors repositories. The documentation
therefore shows `from_opentslab_config(...)` followed by strict
`load_state_dict(...)`; it does not claim the upstream repository works with
Braindecode `from_pretrained(...)`.

## Verification

- [x] All three immutable released artifacts strict-load through the public
  models; tokenizer/tiny numerical signatures match the pinned source.
- [x] PyTorch 2.4.0: 89 model, geometry and quantizer tests passed (3 skips).
- [x] Removed obsolete RMSNorm and compiler-namespace compatibility fallbacks.
- [x] Two-rank distributed EMA synchronization and bounded first training
  update.
- [x] Real-MNE and lightweight EEG/MAG/GRAD geometry, config, and strict local
  Hub round trips for both public models.
- [x] Foundation/modules, model integration, config/Hub, return-features,
  compile/export, reset-head, and invalid-boundary suites.
- [x] Ruff, mypy, isort, codespell, Sphinx lint, build, Twine, archive, and
  packaged-license gates.
- [x] Independent review at exact `c8b01ab478220a2ea8460bbc064c6f1bd61753c1`
  found no actionable issue.

The PR remains a draft while fresh exact-SHA CI runs. Publishing converted
Braindecode-format weights, if desired later, is a separate maintainer action
and is not part of this PR.


bruAristimunha:
I completed a maximum-scope and paper-gate audit of this draft. The exact live
head should remain provenance evidence only; it should not be rebased or merged
as one 63-commit, 21-file, +4,127-line candidate.

The smallest defensible replacement stack is:

1. one shared license/archive prerequisite, only if still needed;
2. one BrainTokenizer architecture/raw-checkpoint PR, keeping BrainOmni-only
   geometry, RoPE, VQ, and SEANet helpers private;
3. one classifier-only BrainOmni PR stacked on the exact tokenizer commit;
4. an external exact-final-SHA PhysioNet-MI gate;
5. tokenizer then unchanged-classifier merges only after that gate returns a
   schema-valid `PASS`.

The qualifying target is BrainOmni-tiny, PhysioNet-MI Table 2 balanced accuracy
`0.580 +/- 0.019`, with 30 fold/seed/LR cells, the official padded-sampler
metric, and a clean unique-example sensitivity. This is a prespecified target,
not a reproduced result. No Margaret/Iceberg data or checkpoint was downloaded:
the experiment must not start until the final replacement classifier SHA is
reviewed, published as a draft, and frozen.

Please drop the global base `ch_type` change, shared `PatchTokenizer` stride
change, public RoPE/VQ APIs, generic geometry utilities, broad README/docs, and
the 1,342-line shared-test expansion. Keep raw checkpoint loading beside each
architecture, use one focused `test_brainomni.py`, and keep conversion,
training, rehosting, and reproduction code outside Braindecode. Scope remains
frozen until the external result gate passes.

Audited head: `c8b01ab478220a2ea8460bbc064c6f1bd61753c1`.

PR #1044, opened 2026-06-12
clarify model summary type labels

### LINKED ISSUE #912: Model summary issue
In the model summary

https://braindecode.org/stable/models/models_table.html

The models are all "classification", yet most of them can do regression as well if you use a single output. Classification is not the right word. Also, it is unclear why some models cannot do embeddings (strip the last layer, and we have embeddings, no?).

tayal-sarthak:
summary
- swaps the model table type label from classification to prediction
- describes prediction as a supervised head that can cover classification, regression
- keeps embedding tied to documented public embedding output

closes #912


bruAristimunha:
heeeey @tayal-sarthak!

Super nice, and I am happy to merge once everything is green, but I would like to add a small provocation to the PR! 

Can we add a tag and this flag to each model, and stop depending on the table for this for this collumn? In the long run, taking inspiration from Hugging Face's learnable points, each model file should contain its own piece of information.  

https://huggingface.co/spaces/transformers-community/Transformers-tenets

bruAristimunha:
many thanks @tayal-sarthak 🙏🏽

PR #1047, opened 2026-06-15
ci: cap pytest below 9.1 to keep pytest_cases working

adammounir:
## Problem

CI is currently red on **basically every open PR**, and it's not the PRs' fault.

`pytest` **9.1.0** (released a few days ago) changed the `IdMaker` constructor signature, which breaks `pytest_cases` **3.10.1** (the latest release). The suite now crashes at **collection time**, before any test runs:

```
TypeError: IdMaker.__init__() takes 7 positional arguments but 8 were given
    (pytest_cases/common_pytest.py)
```

We don't cap `pytest` in the `tests` extra, so fresh CI jobs pull 9.1.0 and die instantly. Any green checks still visible are stale results from before 9.1.0 dropped.

## Confirmation (A/B, same code & pytest_cases 3.10.1)

| pytest | result |
|--------|--------|
| 9.0.3  | 123 collected → 109 passed / 14 skipped |
| 9.1.0  | crash at collection (`IdMaker` TypeError) |

So it's **9.1.0 specifically** — `pytest 9.0.x` is fine.

## Fix

Cap `pytest<9.1` in the `tests` extra until `pytest_cases` supports pytest >= 9.1. This PR's own CI validates the fix (green = cap works).

## Follow-up

Once merged, open PRs (incl. #1037) just need `master` merged in ("Update branch") to pick up the cap and go green.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -45,7 +45,11 @@ API and behavior changes
 Requirements
 ============
 
-- None yet
+- Cap the test dependency to ``pytest<9.1`` in the ``tests`` extra. ``pytest``
+  9.1.0 changed the ``IdMaker`` constructor signature, which breaks
+  ``pytest_cases`` 3.10.1 and makes the whole test suite crash at collection
+  time. The cap can be lifted once ``pytest_cases`` supports ``pytest>=9.1``.
+  By `Adam Mounir`_.
 
 Bug fixes
 ==========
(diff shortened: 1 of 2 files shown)


bruAristimunha:
Thanks @adammounir, the error is related to the internet connection, so it is okay... maybe we need to upgrade to become xfail if it is internet related.

PR #1048, opened 2026-06-15
docs: fix broken EEGNeX quickstart snippet on the landing page

bkowshik:
## Summary

The "Train a model in 15 lines" EEGNeX quickstart on the documentation landing page (`docs/index.rst`) crashes and could never have run as written. This rewrites it to the current idiomatic braindecode usage and adds a `whats_new` entry.

## The bug

Running the homepage snippet fails with:

```python
RuntimeError: mat1 and mat2 shapes cannot be multiplied (128x1240 and 280x4)
```

image

## Root cause

The snippet was adapted from the canonical, CI-tested tutorial (`examples/model_building/plot_bcic_iv_2a_moabb_trial.py`) but dropped the lines that make it work while keeping the magic numbers (`n_chans=22`, `n_times=1125`) that depend on them. Three independent issues:

1. **Channels never restricted.** `BNCI2014_001` loads 26 channels (22 EEG + 3 EOG + 1 STIM). With `n_chans=22`, EEGNeX's `(n_chans, 1)` spatial conv collapsed the channel axis to `26-22+1 = 5` instead of `1`. The STIM trigger channel is also onset-aligned with the class label → **label leakage** if forced to "work" by setting `n_chans=26`.

2. **Window length.** `create_windows_from_events(dataset)` (no offset) yields 1000-sample (4 s) windows, but the model hardcoded `n_times=1125`.

3. **Validation split.** The default stratified `ValidSplit` raises on `clf.fit(windows, y=None)`.

Net: the activation flattened to `8 × 5 × 31 = 1240`, vs the final layer's `8 × 1 × 35 = 280`.

## The fix

* `Preprocessor("pick_types", eeg=True)` — keep the 22 EEG channels, drop STIM+EOG (fixes the shape **and** the leakage).
* Pass the model **by name** (`EEGClassifier("EEGNeX", ...)`) so `n_chans`, `n_times`, and `n_outputs` are inferred from the data — no hardcoded dimensions to drift out of sync. (braindecode already warns against passing a pre-initialized module.)
* `train_split=None` so `y=None` doesn't hit the stratified split.

image


adammounir (approved):
The snippet now restricts to EEG channels, lets EEGClassifier infer the signal dimensions, and disables the validation split so it actually runs. Thanks!

bruAristimunha:
Very nice addition @bkowshik 🙏🏽 , and I think we need to address the network of hugging network issue that makes the test xfail.

Issue #570, opened 2024-02-14
Fix Issue with EEGClassifier Handling Small Training Sets and Unexpected Class Generation for Binary Classification


YuminosukeSato:
When using the `EEGClassifier` class from the `braindecode` library for training on small datasets, two significant issues arise:

1. **Loss Calculation Failure**: The classifier fails to compute the loss when the training dataset is too small. This problem occurs due to the inability of the loss function to process insufficient data samples, which is a common challenge in machine learning models but should be gracefully handled or documented for users of `EEGClassifier`.

2. **Unexpected Class Generation in Binary Classification**: Despite being configured for binary classification tasks, the classifier sometimes generates predictions for classes outside the binary scope. This behavior is unexpected and can lead to erroneous interpretation of the model's predictions. It suggests a possible issue in how the class labels are being inferred or managed within the classifier, especially under conditions where the model is expected to distinguish between two classes only.

### Expected Behavior

- The `EEGClassifier` should provide a clear error message or documentation indicating the minimum dataset size required for effective training and loss calculation. If possible, implementing a fallback mechanism or a warning that suggests best practices for training on small datasets could enhance usability.
- For binary classification tasks, the classifier should strictly limit its predictions to the two expected classes. Any deviation from this should be addressed by ensuring that the class inference mechanism correctly interprets the training data and the specified `classes` parameter, if provided.

bruAristimunha:
Hey @Buddies-as-you-know!

Thanks for pointing out those issues! 🙏 Could you share a bit of code to help us see what's happening? It'd really help us get to the bottom of this faster.

Also, would you like to try fixing them by creating a PR (Pull Request)? It’s a great way to dive into our open-source project and make a difference. Plus, we all get to learn and improve together. What do you think?

bkowshik:
Hi @bruAristimunha — I looked into this and was able to reproduce the small-training-set issue on a clean install (`braindecode==1.5.2`, and the relevant code path is unchanged on current `master`). Sharing a minimal repro and the root cause below. (The issue also mentions a binary-classification/label problem — that's a separate bug, I'll track it on its own so this stays focused.) I'd be happy to send a PR for this one — it's ready.

### Loss "fails" on small training sets → `drop_last` is silently dropping every batch

[...]

PR #1053, opened 2026-06-17
Warn when drop_last silently discards all training batches on small datasets

bkowshik:
## What & why

`EEGClassifier`/`EEGRegressor` set `iterator_train__drop_last=True`, and skorch's default `batch_size` is `128`. When the (post-`train_split`) training set is smaller than `batch_size`, every batch is dropped, the train loader yields **zero** batches, and the model silently "trains" on nothing — no forward pass, no error, `train_loss` never recorded. With the default `ValidSplit` it's worse: `valid_loss` is still reported but stays frozen across epochs.

Addresses the small-training-set half of #570. (Using `Refs #570`, not `Fixes` — the binary-classification issue in that report is a separate bug for its own PR.)

## Reproduction (before)

```python
import numpy as np, torch

X = np.random.randn(40, 22, 500).astype("float32")   # only 40 windows
y = np.random.randint(0, 2, 40).astype("int64")

clf = EEGClassifier(ShallowFBCSPNet, module__final_conv_length="auto",
                    max_epochs=3, train_split=None, device="cpu")  # default batch_size=128
clf.fit(X, y)
clf.history[:, "train_loss"]   # KeyError — 0 batches ran, nothing trained
```

## Fix

One-shot `UserWarning` in the shared `_EEGNeuralNet.get_iterator` (so both estimators are covered). It inspects the training `DataLoader` skorch already built — reading `dataset`, `batch_size`, `drop_last` — so it uses the true **post-`train_split`** size (catching the `ValidSplit` case a `len(X)` check would miss) and handles `batch_size=-1` for free. Fires only when `drop_last=True` and `n_train < batch_size`:

```
UserWarning: The training set has 32 example(s), which is smaller than batch_size=36. With
iterator_train__drop_last=True (the braindecode default), all training batches are dropped
and the model is not trained. Reduce batch_size to at most 32, or pass
iterator_train__drop_last=False.
```

A warning (not an error) keeps this non-breaking and matches the repo's convention for recoverable misconfig.

## Tests

`test_drop_last_small_trainset_warns`, parametrized over both estimators and several batch sizes, asserts the warning fires iff `batch_size > n_train` (incl. the `batch_size == n` no-warn boundary). Full `test_eegneuralnet.py` passes.

> Some pre-existing tiny-data tests now emit this warning — they fit 5–10 samples at `batch_size=32/128`, so they were silently no-op-training too. They still pass; left as-is to show how easy the footgun is to hit. Can silence with `iterator_train__drop_last=False` if preferred.


adammounir (approved):
Thanks @bkowshik, great catch! 🙌
 
Just validated locally: approach is clean, LGTM. Approving and merging.

This surfaces a few pre-existing tiny-data tests that were silently no-op training and now emit the warning. Will clean those up in a follow-up PR so the warning stays rare and useful.

Thanks again !

PR #1054, opened 2026-06-18
ci: install CPU-only torch in CI to fix disk-exhaustion crash

bkowshik:
## Problem

The `tests` job (`ubuntu-latest, 3.13`) on `master` crashes with **exit 134 (SIGABRT)** inside TorchScript `save`, right after the runner logs `Free space left: 74 MB`. The `docs` build hits the same disk pressure.

## Root cause

`uv pip install --system -e .[all]` resolves the **CUDA build of torch** on Linux, pulling ~1.8 GiB of `nvidia-*` CUDA 13 wheels (cublas, cudnn, cufft, nccl, ...) plus the 0.5 GiB torch wheel. GitHub runners have **no GPU**, so these are never used — pure disk and bandwidth cost, and a direct contributor to the crash.

## Fix

Set `UV_TORCH_BACKEND=cpu` on the install step in `tests.yml` and `docs.yml` so uv resolves the CPU-only torch index ([uv docs](https://docs.astral.sh/uv/guides/integration/pytorch/)):

```yaml
- name: Install Braindecode from Current Checkout
  env:
    UV_TORCH_BACKEND: cpu
  run: uv pip install --system -e .[all]
```

## Why it's safe

- **CI-only** — workflow env, not `pyproject.toml`. GPU users and downstream installs still get CUDA torch by default.
- No-op on macOS (CPU/MPS only); a win on Linux and Windows.
- Tests run on CPU regardless, so behavior is unchanged — just lighter.
- `UV_TORCH_BACKEND` needs uv ≥ 0.7; `setup-uv@v3` installs the latest uv.

## Impact

Drops ~1.8 GiB of CUDA downloads (~5 GB unpacked) per Linux/Windows job, speeds up install, and removes the disk pressure behind the crash.


adammounir (approved):
Thanks @bkowshik , 

The previously-crashing test (ubuntu-latest, 3.13) job now passes and the whole matrix + build_docs are green.

Merging, thanks for the fix! 🙂

PR #1056, opened 2026-06-18
[MAINT] Silence drop_last small-trainset warning in signal-arg tests

adammounir:
Follow-up to #1053. That PR added a UserWarning when `iterator_train__drop_last` drops every training batch on small datasets. It surfaced that several `test_eegneuralnet` signal-argument tests were silently no-op-training — they fit 5–10 mock samples at `batch_size=32/128` only to check that `n_chans`/`n_times`/`n_outputs` propagate, not to actually train.

This scopes a `@pytest.mark.filterwarnings("ignore:The training set has")` to those 9 tests (×2 estimators) so the warning no longer spams CI logs, while staying active everywhere else. The test that asserts the warning fires is untouched.

Not using `drop_last=False` here on purpose: forcing real training on the mock modules surfaces unrelated canned-prediction shape mismatches.

### DIFF
--- a/docs/whats_new.rst
+++ b/docs/whats_new.rst
@@ -104,7 +104,7 @@ Code health
   (:gh:`1053`) in the ``test_eegneuralnet`` signal-argument tests, which
   intentionally fit tiny mock data to check argument propagation rather than
   to train. Keeps the warning meaningful by not emitting it on every CI run.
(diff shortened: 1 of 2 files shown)


bruAristimunha (approved):

bruAristimunha:
thanks @adammounir

PR #1072, opened 2026-06-24
feat: opt-in channel positions in the batch for heterogeneous montages

### LINKED ISSUE #1066: The collection of the dataset should show the ch_info is present, or this should be easier to explore....

bruAristimunha:
Closes #1066.

## Motivation

Today a window yields `(X, y, crop_inds)` and drops channel identity. Channel-aware / variable-montage models (foundation models, REVE-style positional encoders) need to know *where each channel sits* per sample. This adds an opt-in path to carry electrode **positions `(n_ch, 3)` xyz** along in the batch, and makes collections that mix recordings with **different channel sets** trainable.

## What's in it

**Data layer** (`datasets/base.py`)
- Cached `ch_pos` property on `EEGWindowsDataset` / `WindowsDataset`, built from the canonical `info["chs"][i]["loc"][:3]` that braindecode models already consume (e.g. `signal_jepa`, `REVE`), aligned to `X`'s channel rows (misc channels handled).
- `BaseConcatDataset.set_return_ch_pos(True)` toggles the whole collection. **Opt-in / default off** — the `(X, y, crop_inds)` contract is unchanged.

**Heterogeneous montages** (`datasets/collate.py`)
- `pad_channels_collate`: pads each batch to the largest channel count and emits a boolean `ch_mask`. Channel identity is carried by the positions (positional, not index-based), so a batch can mix e.g. a 19- and a 22-channel recording.

**Training integration** (`util.py`)
- `ThrowAwayIndexLoader` routes the signal, positions and mask into the model's `forward` via a skorch dict input, so it flows through `EEGClassifier`. The existing 2-/3-tuple paths are untouched.

**Example + docs**
- `examples/model_building/plot_heterogeneous_montage_training.py` — trains a position-aware model on a mixed-montage collection end to end.
- `whats_new.rst` entry.

## Scope / follow-up

This makes heterogeneous collections **batchable** and feeds positions + a channel mask to the model. How a specific architecture *normalizes* and consumes variable channels (applying `ch_mask` inside attention, mapping positions to a canonical space) is model-specific and intended as a follow-up. The example uses a tiny inline position-aware net to demonstrate the `forward(x, pos=None, ch_mask=None)` contract.

## Tests

- New `test/unit_tests/datasets/test_ch_pos.py` (12 tests): accessor shape/values, default-off backward-compat, misc-channel alignment, no-montage warning, uniform default-collate, heterogeneous pad-collate, non-uniform-time guard, info JSON round-trip (guards the Zarr/Hub lazy path).
- `test/unit_tests/test_util.py`: loader dict-routing + 2-/3-tuple backward-compat.
- Existing dataset/neuralnet suites pass unchanged (default-off).

### DIFF
--- a/braindecode/datasets/__init__.py
+++ b/braindecode/datasets/__init__.py
@@ -9,6 +9,7 @@ from .base import (
 )
 from .bcicomp import BCICompetitionIVDataset4
 from .bids import BIDSDataset, BIDSEpochsDataset
+from .collate import pad_channels_collate
 from .chb_mit import CHBMIT
 from .mne import create_from_mne_epochs, create_from_mne_raw
 from .moabb import BNCI2014_001, HGD, MOABBDataset
@@ -42,4 +43,5 @@ __all__ = [
     "SleepPhysionetChallenge2018",
     "create_from_X_y",
     "BCICompetitionIVDataset4",
+    "pad_channels_collate",
 ]
--- a/braindecode/datasets/__init__.py
+++ b/braindecode/datasets/__init__.py
@@ -9,8 +9,8 @@ from .base import (
 )
 from .bcicomp import BCICompetitionIVDataset4
 from .bids import BIDSDataset, BIDSEpochsDataset
-from .collate import pad_channels_collate
 from .chb_mit import CHBMIT
+from .collate import pad_channels_collate
 from .mne import create_from_mne_epochs, create_from_mne_raw
 from .moabb import BNCI2014_001, HGD, MOABBDataset
 from .nmt import NMT
--- a/braindecode/datasets/base.py
+++ b/braindecode/datasets/base.py
@@ -192,6 +192,31 @@ def _channel_info(mne_obj):
     return n_ch, type_str, sfreq
 
 
+def _compute_ch_pos(info, targets_from):
+    """Return ``(n_ch, 3)`` float32 electrode positions aligned to ``X``'s rows.
+
+    Positions are the MNE head-frame coordinates ``info["chs"][i]["loc"][:3]``
+    (meters) -- the same representation braindecode models already consume (see
+    e.g. ``signal_jepa``/``REVE``). When ``targets_from != "metadata"`` the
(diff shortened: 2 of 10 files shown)


bruAristimunha:
Pushed `158412ae` addressing the failing `pre-commit` and the review feedback:

**CI (pre-commit)**
- Fixed import order in `datasets/__init__.py` (ruff/isort).
- Fixed mypy on `collate.py` — guard the loop/return on `pos_out is not None` so the optional narrows (no more "Unsupported target for indexed assignment").

**Copilot**
- `ch_pos` is now invalidated in the `raw`/`windows` setters, so the cache can't go stale after a montage update or reassignment.
- `_looks_like_channel_mask` keys on **bool dtype only** — a 2D `(n_ch, 3)` position array can no longer be misclassified as a mask.
- `BaseConcatDataset.ch_pos` raises a clear `AttributeError` on non-windowed (e.g. `RawDataset`) collections instead of a bare attribute error.

**Codex (P2 — cropped paths)**
- `predict_with_window_inds_and_ys` and `training.scoring.predict_trials` now unpack the first three batch elements, so they tolerate extended (positions) batches instead of raising "too many values to unpack". Full routing of positions into the *cropped* model forward remains the model-side follow-up noted in the PR.

Added tests for cache invalidation and the clear `RawDataset` error; `pre-commit run --all-files` passes locally and the dataset/util/eegneuralnet suites are green.

bruAristimunha:
Pushed `53c5b04f` addressing the two follow-up review comments:
- `_looks_like_channel_mask` now also accepts NumPy bool dtype (not only `torch.bool`), so a bool-array mask can't be misrouted as positions.
- The cropped prediction paths (`predict_trials`, `predict_with_window_inds_and_ys`) now raise a **clear error** when given position batches instead of silently dropping `ch_pos`/`ch_mask` — cropped decoding + `return_ch_pos` is unsupported, and failing loudly is better than silently mis-evaluating. Added a unit test for the dtype detection.

PR #1075, opened 2026-06-25
Add DANCE: detect-and-classify event detection model

bruAristimunha:
## Summary

Adds **DANCE** (Detect And classify eveNts in eeg) — Lévy, Banville, Rapin, King, Moreau & d'Ascoli, *DANCE: Detect and Classify Events in EEG*, arXiv:2605.10688 (Meta FAIR + Inria, MIT) — ported from [`facebookresearch/dance`](https://github.com/facebookresearch/dance).

Unlike braindecode's usual epoch-classification models, DANCE frames decoding as **set prediction**: it takes a long, unaligned window and **detects the set of events** it contains, each as `(start, end, class)` — DETR for 1-D EEG. A Perceiver bottleneck makes it **length-agnostic** (any `T` → a fixed 256-latent grid); a spatial-Fourier `ChannelMerger` makes it **montage-agnostic**.

Architecture: `(B,C,T) → ChannelMerger → SimpleConv (dilated convs) → Perceiver (256 latents) → {dense head (forward), DETR cross-attention decoder (detect)}`.

## braindecode-native adaptation

DANCE doesn't fit the classifier contract, so it was adapted (not vendored):

- **`forward(x) -> Tensor`** returns the dense per-token logits `(B, 256, n_outputs)` (a sequence output, like `EMG2QwertyNet`/`MetaNeuromotorHand`); `DANCE` is registered in `non_classification_models`.
- **`detect(x) -> dict`** returns the event set `{class, start, end, dense}` — `forward` stays monomorphic.
- **Loss lives outside the model**: `DanceLoss` (Hungarian matcher + DETR/dense/consistency terms) in `braindecode/training/losses.py`, like every other braindecode model.
- **Channel positions** are derived from `chs_info` at init (registered buffer), with auto-fallback to merger-off when locations are absent — `forward` keeps the `(B, C, T)` signature.
- All architecture internals (`FourierEmb`, `ChannelMerger`, `SimpleConv`, `Perceiver`) are **re-implemented in `braindecode/modules/`** — no new runtime dependencies.

## What's added

- `braindecode/models/dance.py` — `DANCE(EEGModuleMixin, nn.Module)` (`forward`, `detect`, `reset_head`, positions buffer)
- `braindecode/modules/dance_modules.py` — `FourierEmb`, `ChannelMerger`, `SimpleConv`, `Perceiver` (exported) + `DanceDetrDecoder` (internal)
- `braindecode/training/losses.py` — `HungarianMatcher`, `DanceLoss`
- `braindecode/training/scoring.py` — `f1_event` (COCO-style, IoU>0.5 + class match), `f1_sample`
- `braindecode/functional/functions.py` — `iou_1d`, `pairwise_iou_1d`, `detr_to_dense_probs`, `events_to_mask`, `extract_events_from_detr_batch`
- `examples/applied_examples/plot_dance_event_detection.py` — plain-torch tutorial with reusable target-builder/collate + an executed smoke test
- Registration: `models/__init__.py`, `util.py` (mandatory params + `non_classification_models`), `summary.csv`
- `scripts/dance_parity_check.py` — dev-only numerical-parity verification (not in CI)

## Numerical parity

Re-implemented modules were verified **weight-for-weight against upstream** `neuraltrain` + `perceiver_pytorch`: the encoder output and the dense `forward` logits are `torch.allclose` (183 tensors mapped). The DETR decoder is a **documented loose port** (inline pre-norm attention vs upstream's `x_transformers` block), so its `detect()` output is correctness-tested, not parity-gated; `DanceLoss` is likewise correctness-tested (finite + decreases on overfit), not parity-gated. These loosenings are documented at the code sites.

## Test plan

- `test/unit_tests/modules/test_dance_modules.py` — module shapes, dilation rule, positions, length-agnostic, + parity gates (run against upstream when importable)
- `test/unit_tests/models/test_dance.py` — init / `forward` / `detect` / `reset_head` / merger fallback / length-agnostic / batch-1 train mode
- `test/unit_tests/training/test_dance_loss.py` — matcher, `DanceLoss` (finite + overfit), functional helpers, target-builder, end-to-end tutorial smoke test
- `test/unit_tests/functional/test_dance_metrics.py` — `f1_event` / `f1_sample` hand-computed cases
- braindecode contract suite (`test_integration`/`test_config`/`test_huggingface`/`test_model_categorization`) green for `DANCE`; added to the `jit.script` skip-list (einops/Perceiver); `torch.compile` and `torch.export` verified to pass.

Full model-contract suite: **791 passed / 0 failed**. `make html-noplot` docs build succeeds (tutorial + autodoc render).


bruAristimunha:
### Automated-review disposition

Worked through every Copilot/Codex thread (several are duplicates across re-reviews). Fixes are in `4fce03ac`.

**Fixed**
- **`HungarianMatcher` `__call__` → `forward`** — proper `nn.Module` (hooks no longer bypassed).
- **Device mixing** — the matched-target buffers are now built on the predictions' device, so the matcher runs on CUDA.
- **Matcher cost → numpy** — `cost.detach().cpu().numpy()` (it was already under `@torch.no_grad()`; detach added defensively).
- **`DanceLoss` dense-shape guard** — `assert` → `ValueError` (survives `python -O`).
- **Weight init order** — `self.apply(self._init_weights)` now runs *after* `final_layer`/the decoder class head are created, so the dense head gets the custom init; `final_layer` stays the last child (contract A3).
- **`versionadded`** — `1.5` → `1.6.1` (matches the `_current` whatsnew target).
- **`activation` docstring** — corrected: it is accepted for interface symmetry but currently inert (the decoder feed-forward hardwires GEGLU).
- **`f1_event` empty/empty** — now returns `1.0` (nothing to detect, nothing predicted) instead of `0.0`; covered by new tests.

**Kept by design (with rationale)**
- *Decoder predicts `start`/`end` independently (can yield `start > end`)* — upstream-faithful; the paper predicts each boundary independently and the IoU/loss handles ordering.
- *`detr_to_dense_probs` Python loops / `float(tensor)`* — a training-side, parity-frozen rendering used only by the consistency loss (not on the inference path); a vectorized version is a separate follow-up.
- *Tutorial executes at import / the test re-imports it* — sphinx-gallery requires gallery scripts to run; the test intentionally smoke-runs the full pipeline once.

bruAristimunha:
Hey @jarod1212 and @tomMoral! 

I am trying to adapt dance to braindecode, but I still have some more questions in the sense of what we actually need to port to braindecode, and what we can cut.

I try to push everything at the moment, but I would like to trim as much as possible, and also replicate one of the results of the paper (some easy results pls).

Can you give me some input here?

I think the example is nice, but we can potentially drop many components, like the loss, the metrics, or replace for things that are already default in pytorch or Torch Audio or any of the dependencies. Everything that it is not core of the model.

Many thanks in advance for your input and revision @jarod1212 🙏🏽

jarod1212:
Hello @bruAristimunha ! So cool to have tried this — many thanks! 👯‍♂️ 👯‍♂️ 👯‍♂️ 👯‍♂️ 👯‍♂️ 

We can definitely remove f1_sample (it's directly torchmetrics.MultilabelF1Score(average="macro")), and along with it events_to_mask and extract_events_from_detr_batch. detr_to_dense_probs can be moved inside DanceLoss as a private helper. Other parts are all core.

One thing to fix before reproducing, otherwise the model will plateau low for the wrong reason. Your IoU term is averaged over all Q queries; losses.py #L87-92 (https://github.com/facebookresearch/dance/blob/main/dance/losses.py) averages only over the Hungarian-matched pairs. Your refactor dropped the matched-preds wrap, so every unmatched query contributes (1 - 0) = 1 to the mean and flattens the IoU signal. You should mask before averaging! 

Easiest example to reproduce: Korczowski et al. 2014 (BI2014a, P300) — auto-downloaded via MOABB, should give F1-event ~0.525 (per-subject, mean ± SEM over 5 folds). BNCI2014 is also easy/light and quite mainstream! 

Dataset config can be found here: https://github.com/facebookresearch/dance/blob/main/dance/configs/datasets/korczowski2014a.yaml
and hparams here: https://github.com/facebookresearch/dance/blob/main/dance/configs/defaults.yaml 

Let me know if any issue, I can reproduce on my end as soon as I have time but I think an agent can easily do it.

Jarod

bruAristimunha:
Thanks @jarod1212 🙏🏽, please let me know your feedback on the documentation dev once is builded at https://braindecode.org/dev/

PR #1076, opened 2026-06-25
BrainModule: add the (missing) optional spatial Fourier ChannelMerger + fix float-dilation crash

bruAristimunha:
## Summary

`BrainModule` (the brainmagick "SimpleConv") **documents** a "spatial attention layer that projects sensor locations onto a 2D plane using Fourier-parameterized functions" — but the code never implemented it (`__init__` used `SubjectLayers` + a plain `nn.Conv1d`). This PR makes that docstring true, fixes several latent bugs uncovered along the way, and introduces a **shared** spatial-Fourier merger reused by both `BrainModule` and the DANCE model (#1075).

## What's added

- **`braindecode/modules/merger.py`** (new, exported): `FourierEmb`, `ChannelMerger`, `positions_from_chs_info`. Re-implemented to match upstream `neuraltrain.models.common` and **numerically parity-verified** (`torch.allclose`, the same blocks DANCE validates). One canonical home for the spatial merger.
- **`BrainModule`** gains `use_merger: bool = False` (default off → existing behavior unchanged), `n_virtual_channels: int = 270`, `merger_drop_prob: float = 0.2`. When enabled, electrode positions are derived from `chs_info` at init (a `(n_chans, 2)` buffer) and the merger remaps `n_chans → n_virtual_channels` as the first step of `forward`. Auto-disables (with a `UserWarning`) when `chs_info` lacks usable locations.

## Bug fixes to existing code

- **Float `dilation_growth` crash:** `_ConvSequence` passed a float dilation straight into `nn.Conv1d` — any non-integer `dilation_growth` (e.g. `2.5`) raised `TypeError` on `forward`. Now cast to `int` at use (float accumulator), matching upstream; `dilation_growth` is re-typed `float`.
- **`subject_layers` + STFT crash (and a latent STFT + `subject_embedding` mis-sizing):** the `__init__` channel accounting was computed in a different order than `forward` runs, mis-shaping `SubjectLayers`/`input_projection` for combined configs. The accounting now follows the forward order (`merger → STFT → subject_layers → subject_embedding → input_projection`), so these combinations work. Verified against the full parametrized subject/STFT test matrix (no regression).

## Robustness / validation

- `n_virtual_channels <= 0` and out-of-range `merger_drop_prob` are rejected up front with clear errors (`merger_drop_prob` is documented correctly as a spatial-attention **radius**, not a Bernoulli probability — values `>= 1` would zero the merger output).
- `ChannelMerger` works under half precision (`model.half()`) and across devices (score offset built on `positions`' device, `x`'s dtype).
- `use_merger` + STFT emits a memory-cost `UserWarning` (valid but heavy: STFT over 270 virtual channels).

## Differences from upstream brainmagick / neuraltrain SimpleConv (documented in the class docstring)

| BrainModule (this lib) | Upstream (DANCE config) |
|---|---|
| `kernel_size=3`, `dilation_growth=2` (int default), `dilation_period=5` reset | `9`, `2.5`, no reset |
| `batch_norm`/`glu`/residual default **on** | **off** |
| `AdaptiveAvgPool1d → Linear` head | per-timestep sequence |
| optional `SubjectLayers`, STFT, channel-dropout | (different training stack) |
| spatial `ChannelMerger` now opt-in, **shared** via `braindecode.modules.merger` | merger nested in SimpleConv |

## Relationship to #1075 (DANCE)

`braindecode/modules/merger.py` is the **single shared home** for the spatial merger; #1075 imports `ChannelMerger`/`FourierEmb` from it (its `dance_modules.py` copies were dropped). Both PRs touch `braindecode/modules/__init__.py` — reconcile the single export block at merge.

## Scope note

A canonical per-batch `ch_pos` pipeline (#1066, raw 3D coords) now exists; this PR intentionally uses the simpler `chs_info`-at-init `[0, 1]` positions (matching DANCE + parity). Wiring the per-batch `ch_pos` path (3D→2D projection + heterogeneous-montage masking, REVE-style `forward(x, pos=None)`) is a clean follow-up.

## Test plan

`test/unit_tests/models/test_models.py` (`brain_module`/`brainmodule`): **72 passed**, including the 59 parametrized `subject_embeddings`/`subject_layers`/STFT regression cases plus new tests for merger shape, auto-disable (all-zero & absent `chs_info`), param validation, half/device, the now-supported `subject_layers`+STFT and STFT+`subject_embedding` combos, and the merger+STFT warning. `test_modules.py` 180 passed; model-contract (config/integration/HF) green for `BrainModule`.


bruAristimunha:
### Automated-review disposition

Went through every Copilot/Codex comment. Two were genuinely actionable on the current code and are fixed in `89a9b247`; the rest were either already addressed by later commits (and now reference the removed `merger.py` — the classes moved into `layers.py` next to `SubjectLayers`) or are deliberate, documented choices.

**Fixed now (`89a9b247`)**

- **Dropout `center` device mismatch** — `center` was built on `x.device` but subtracted from `positions`; now keyed off `positions.device` (matching `score_offset`). Parity-neutral.
- **`channel_dropout_type` docs** — the param docstring now states it is unsupported with `use_merger=True` (and why).

**Already addressed by earlier commits (these threads reference the now-deleted `merger.py`)**

- *`warnings` imported but unused* — removed; `merger.py` no longer exists (code is in `layers.py`, no `warnings` import).
- *`score_offset` device / low-precision dtype* — `score_offset` is built on `positions.device` with `dtype=x.dtype`; `ChannelMerger` runs under `model.half()`.
- *`dilation_growth` annotated `int`* — now annotated `float` (signature, docstring, validator).
- *header references `scripts/dance_parity_check.py`* — that reference is gone; the `layers.py` provenance comment is generic.
- *auto-disable warning too narrow* — broadened to "`'loc'` missing, non-finite, or all-zero".
- *MIT header in a BSD-3 repo* — the standalone MIT-headed `merger.py` is removed. `ChannelMerger`/`FourierEmb` are **re-implementations** (not verbatim copies) that now live in BSD-3 `layers.py` with an attribution comment crediting the MIT/Meta upstream — the same pattern used for `SubjectLayers` (brainmagick) and `DropPath` (VISSL) already in that file.

**Kept by design (with rationale)**

- *Rename `merger_drop_prob` → `merger_drop_radius`* — kept for consistency with the DANCE model, which exposes the identically-named param; the docstring explicitly documents it as a *radius* (not a Bernoulli probability) and the value is range-validated (`0 <= x < 1`).
- *Validate `ChannelMerger` ctor args / `forward` shapes* — `pos_dim` is already validated via `FourierEmb`; the merger is a parity-frozen re-implementation used primarily through `BrainModule`/`DANCE` (which validate their public knobs), and a shape mismatch surfaces as a clear torch size error. Not adding redundant guards to the parity-gated path.
- *`.data` rescale in `__init__`* — a deterministic init-time scaling with no autograd interaction at construction, matching the neighboring `SubjectLayers` (which rescales via `.data` the same way) and the upstream.
- *`positions_from_chs_info` input validation* — paired with the companion `has_valid_locations`; the internal callers guard with it before calling.

The shared `layers.py` change (the `center` device fix) will be mirrored to the DANCE branch (#1075) to keep that file byte-identical.

Issue #732, opened 2025-04-23
Torch script compatibility for all the models


bruAristimunha:
All models since #730 are torch.compile and torch.export compatible. Almost all are torch.script compatible, except for the following:

```python
not_working_models = [
    "Deep4Net",
    "EEGResNet",
    "ShallowFBCSPNet",
    "Labram",
    "ContraWR",
    "BIOT",
    "EEGMiner",
    "SignalJEPA",
    "SignalJEPA_Contextual",
    "SignalJEPA_PostLocal",
    "SignalJEPA_PreLocal",
]
```

Each for different reasons, we have four groups of models:

1) Models that need better reconstruction from sequential data:
   - "Deep4Net"
   - "EEGResNet"
   - "ShallowFBCSPNet"

2) Models that need redefinition of some layers:
   - "Labram"
   - "EEGMiner"
   - "ContraWR"

3) Models with complex dependencies:
   - "BIOT"

4) Models that require a better parser for channel information:
   - "SignalJEPA"

From the benchmark results, only the first group matters here.

bruAristimunha:
now it only:

```python
not_working_models = [
    "Labram",
    "EEGMiner",
    "SignalJEPA",
    "SignalJEPA_Contextual",
    "SignalJEPA_PostLocal",
    "SignalJEPA_PreLocal",
]
```

PierreGtch:
now only:

```python
not_working_models = [
    "Labram",
    "BIOT",
    "EEGMiner",
]
```

PR #1101, opened 2026-07-13
make eegminer torchscript compatible

tayal-sarthak:
- replaces eegminer callable dispatch with a scriptable feature layer
- preserves magnitude, correlation, phase locking behavior
- keeps generalized gaussian parameter clamping during eager execution
- adds torchscript typing to hilbert plus plv helpers

partially addresses #732 

Thank you :)

### DIFF
--- a/braindecode/functional/functions.py
+++ b/braindecode/functional/functions.py
@@ -1,5 +1,6 @@
 # Authors: Robin Schirrmeister <robintibor@gmail.com>
 #          Bruno Aristimunha <b.aristimunha@gmail.com>
+#          Sarthak Tayal <sarthaktayal2@gmail.com>
 #
 # License: BSD (3-clause)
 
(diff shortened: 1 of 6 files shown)


bruAristimunha (commented):
Here, the only thing that needs to be adjusted is to make sure that we are not breaking the compatibility of the weights. 

I feel that we are.

tayal-sarthak:
Thank you for raising this concern. I verified compatibility using checkpoints created from the commit preceding this pull request across the `mag`, `corr`, and `plv` modes.

Each checkpoint loads on this branch with `strict=True`; the state dictionary keys are unchanged, and the outputs are bitwise identical with `rtol=0` and `atol=0`. The introduced feature layer and `Unflatten` layer do not have parameters or buffers, therefore they introduce no checkpoint keys.

I also added a regression test that asserts the legacy state dictionary key set and strict loading behavior.

bruAristimunha:
many thanks @tayal-sarthak, the error in the CI is not related

Issue #1097, opened 2026-07-03
Include more ieeg/seeg models:


bruAristimunha:
In sEEG cognitive-state decoding, we compare against the supervised baseline Seegnificant (Mentzelopoulos et al., [2024](https://arxiv.org/html/2606.15278v1#bib.bib28)), and leading self-supervised methods: Brant (Zhang et al., [2023](https://arxiv.org/html/2606.15278v1#bib.bib54)), Du-IN (Zheng et al., [2024](https://arxiv.org/html/2606.15278v1#bib.bib56)), BaRISTA(Oganesian et al., [2025](https://arxiv.org/html/2606.15278v1#bib.bib29)

julien-gadonneix:
A few other suggestions if they are of interest to you :)
- BrainBERT (Wang et al., [2023](https://arxiv.org/abs/2302.14367))
- PopT (Chau et al., [2024](https://arxiv.org/abs/2406.03044))

adammounir:
Thanks for the suggestion @julien-gadonneix ! I'll take a look at them. 🙌

julien-gadonneix:
Thanks and congratulations on the implementations, I'll soon start to work with them. I also found this model that may be relevant if you are interested in reimplementing it:
https://arxiv.org/pdf/2506.17068

julien-gadonneix:
Suggested by @bruAristimunha : https://arxiv.org/html/2512.19097v2

PR #1104, opened 2026-07-20
Add BrainBERT: foundation model for intracranial (sEEG/iEEG) signals

adammounir:
## Summary

Towards #1097 — adds a braindecode-native port of **BrainBERT** (Wang et al.,
*BrainBERT: Self-supervised representation learning for intracranial recordings*,
ICLR 2023), a self-supervised foundation model for intracranial (sEEG/iEEG)
signals. Suggested on #1097.

Upstream code and released weights:
* code — https://github.com/czlwang/BrainBERT
* paper — https://arxiv.org/abs/2302.14367

## Architecture

`(B, C, T) → STFT spectrogram → linear projection + sinusoidal positional
encoding → Transformer encoder → pool over channels & frames → head`.

Following the convention established for **Brant** (#1100), the short-time
Fourier transform front-end is computed **inside** `forward` (module
`_STFTSpectrogram`) so the model keeps the standard `(batch, n_chans, n_times)`
input signature, whereas the upstream reference consumes a pre-computed
spectrogram.

## What's in this PR

* `braindecode/models/brainbert.py` — `BrainBERT(EEGModuleMixin, nn.Module)`
  with modest ready-to-run defaults (~0.65M params) and the released *large*
  config documented (`hidden_dim=768, ffn_dim=3072, n_heads=12, n_layers=6`,
  ~43M), plus `from_pretrained`, `reset_head` and `return_features`.
* `braindecode/modules/brainbert_modules.py` — STFT front-end, input embedding,
  sinusoidal positional encoding, spectrogram-prediction head (weight parity)
  and classification head.
* Registration: `models/__init__.py`, `models/util.py`, `summary.csv`,
  `docs/whats_new.rst`.
* `test/unit_tests/models/test_brainbert.py` — contract tests + a **parity
  gate**.

## Numerical fidelity

* The in-model STFT reproduces the upstream scipy magnitude spectrogram to
  **1.5e-6** (bit-exact in float32).
* **Parity gate**: with the upstream reference available (`BRAINBERT_SRC`), the
  input encoding + Transformer are checked **bit-exact** (`atol=1e-5`) against
  the upstream `MaskedTFModel`; the gate is skipped otherwise so CI stays green
  without the external dependency.

## Checklist

- [x] Faithful architecture (STFT front-end, input encoding, Transformer, head).
- [x] Numerical-parity check against the upstream reference.
- [x] Register the model: `models/__init__.py`, `models/util.py`, `summary.csv`.
- [x] Contract tests (`test/unit_tests/models/test_brainbert.py`).
- [ ] Re-host the official pretrained weights on the braindecode HF Hub
      (`braindecode/brainbert-pretrained`).
- [ ] Tutorial in `examples/`.

### DIFF
--- a/braindecode/models/brainbert.py
+++ b/braindecode/models/brainbert.py
@@ -1,0 +1,220 @@
+"""BrainBERT: a self-supervised foundation model for intracranial signals.
+
+Port of BrainBERT (Wang et al., ICLR 2023) into a braindecode-native model.
+Upstream code and pretrained weights are released by the authors:
+
+* paper: https://arxiv.org/abs/2302.14367
+* code: https://github.com/czlwang/BrainBERT
+* weights: released by the authors (Google Drive, see the upstream README)
+
+BrainBERT learns representations of intracranial electrode data by masked
+modelling of its **spectrogram**. The Transformer encoder and input encoding are
+ported weight-for-weight from the upstream reference (bit-exact, see
+``scripts/brainbert_parity_check``); the short-time Fourier transform front-end
+is moved *inside* the model (a braindecode-native adaptation) and the pooling /
+classification head is a braindecode-native addition. The official pretrained
+checkpoint loads directly via
+``BrainBERT.from_pretrained("braindecode/brainbert-pretrained")``.
+"""
+
+from __future__ import annotations
+
+import torch
+import torch.nn as nn
+
+from braindecode.models.base import EEGModuleMixin
+from braindecode.modules.brainbert_modules import (
+    _BrainBERTHead,
+    _BrainBERTInputEmbedding,
+    _SpecPredictionHead,
+    _STFTSpectrogram,
+)
+
+
+class BrainBERT(EEGModuleMixin, nn.Module):
+    r"""BrainBERT from Wang et al. (2023) [BrainBERT2023]_.
+
+    :bdg-danger:`Foundation Model` :bdg-info:`Attention/Transformer`
+
+    BrainBERT is a self-supervised foundation model for intracranial neural
+    signals (sEEG/iEEG). The raw signal is turned into a **spectrogram** by a
+    short-time Fourier transform; each time frame is a token. A linear projection
+    and a fixed sinusoidal positional encoding feed a stack of standard
+    Transformer encoder layers. The model is pre-trained by masked-spectrogram
+    modelling — reconstructing masked time/frequency patches — and the resulting
+    per-frame representations are used for downstream decoding.
+
+    Following the braindecode convention, the STFT is computed **inside**
+    ``forward`` (via :class:`~braindecode.modules.brainbert_modules._STFTSpectrogram`)
+    so the model keeps the standard ``(batch, n_chans, n_times)`` input
+    signature, whereas the upstream reference consumes a pre-computed spectrogram.
+
+    The released checkpoint expects signals sampled at **2048 Hz**,
+    Laplacian-re-referenced, with ``nperseg=400``, ``noverlap=350`` and the first
+    ``freq_cutoff=40`` frequency bins. The defaults below are a modest,
+    ready-to-run configuration; the **released ("large") model** uses
+    ``hidden_dim=768``, ``ffn_dim=3072``, ``n_heads=12`` and ``n_layers=6``
+    (~43M parameters) — pass these to reproduce it.
+
+    .. important::
+       **Pre-trained weights available.** The official checkpoint is released by
+       the authors and loads directly::
+
+           model = BrainBERT.from_pretrained(
+               "braindecode/brainbert-pretrained", n_outputs=2
+           )
+
+       It uses the "large" configuration above; ``n_chans`` and ``n_outputs`` may
+       be changed freely, as frames are pooled and the classification head is
+       task-specific.
+
+    .. versionadded:: 1.7
+
+    Parameters
+    ----------
+    hidden_dim : int, optional
+        Transformer model width ``D``. Default 192. The released model uses 768.
+    ffn_dim : int, optional
+        Inner dimension of the Transformer feed-forward blocks. Default 384.
+        The released model uses 3072.
+    n_layers : int, optional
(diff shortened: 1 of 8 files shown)


adammounir:
I fed identical inputs through the upstream BrainBERT stack (`BrainBERT/models/masked_tf_model.py`, from the official repo) and the braindecode port with the released `stft_large` weights copied over.

**Port parity (gated in CI):**

| Check | Upstream ref | braindecode port | max\|diff\| |
|---|---|---|---|
| STFT front-end | `scipy.signal.stft` | `_STFTSpectrogram` | **~1.5e-6** |
| Transformer encoder | `MaskedTFModel` | `BrainBERT` | **bit-exact** (atol 1e-5) |
| `from_pretrained` | official `stft_large` (43.19M) | `braindecode/brainbert-pretrained` (43.19M) | **6.4e-5** |

The `from_pretrained` residual (6.4e-5, vs 0.0 for the freshly-built encoder) comes solely from the sinusoidal positional encoding being regenerated on CPU vs baked on GPU at pretraining time; it grows with position and stays negligible at real sequence lengths. Reproduce with `BRAINBERT_SRC=<upstream> pytest test/unit_tests/models/test_brainbert.py`; the parity gate skips cleanly without `BRAINBERT_SRC`. Weights re-hosted at `braindecode/brainbert-pretrained`: `BrainBERT.from_pretrained("braindecode/brainbert-pretrained", n_outputs=2)`.

**Downstream replication (Brain Treebank, CC-BY):** beyond activation parity, I ran the paper's linear-probing protocol end-to-end (frozen features → `Linear(768→1)`, ROC-AUC per electrode) on sub_3/sub_4, all valid Laplacian electrodes, on the 4 paper tasks. Swapping the upstream extractor for the braindecode port is a no-op on the metric — over 76 shared electrode×task pairs, **max\|ΔAUC\| = 0.0023, mean\|ΔAUC\| = 0.0004**. Absolute levels match the paper on good electrodes (best-electrode **onset 0.84**, **speech 0.80**).

bruAristimunha (changes requested):
I reviewed this exact candidate against the BrainBERT paper, released source,
dataset records, and checkpoint metadata. I am requesting changes because the
mandatory paper-result gate is not yet legally or scientifically runnable.

The easiest faithful gate is Table 2: frozen-STFT BrainBERT sentence-onset
ROC-AUC `0.66 +/- 0.03`, aggregated as the mean and population standard
deviation across the ten electrodes selected by the separate time-domain
baseline. Current blockers are material:

- the upstream source/checkpoint has no declared reuse terms, and the current
  Braindecode checkpoint is an unverified rehost with no proved immutable
  lineage to the official mutable archive;
- the PR pools all channels and frames with LayerNorm, whereas the paper's
  downstream path uses one electrode, exactly ten center frames, and a plain
  linear head;
- the released sources disagree about the preprocessing used for the
  published row, so choosing one silently would invent a protocol.

Please freeze the exact selected electrodes, recordings, preprocessing,
checkpoint bytes/hash/license, and aggregation contract, then reproduce this
row on one exact candidate SHA within a preregistered tolerance before merge.
For scope, reduce the library contribution to the faithful single-electrode
model/checkpoint boundary with focused tests; keep dataset acquisition and the
full reproduction experiment external and opt-in. No signal or checkpoint
bytes were downloaded because the permission/protocol gate is incomplete.

Reviewed head: `7b1cff66fb20cef78cce2ef991f9d6f77f48a55b`.

bruAristimunha:
Paper-replication gate at exact head 7b1cff66fb20cef78cce2ef991f9d6f77f48a55b: hold.

The exact released checkpoint now runs on the NEMAR Brain Treebank adaptation, and a bounded subject-03/run-01 frozen onset probe reached test ROC-AUC 0.7569192871 (BCE 0.5903241634). This is useful checkpoint/data-path evidence, not the paper's 0.82 +/- 0.07 result: it omits joint fine-tuning, top-10 electrode selection, seven-session aggregation, and the original HDF5 equivalence. The upstream source also publishes no license while the current Hub metadata claims BSD-3-Clause.

A current-master correction candidate closes the NaN STFT path and removes a false stft_clip equivalence claim, but it remains local because the paper/result and rights gates are still open. Do not merge on architecture tests alone.

braindecode/models/brainbert.py
```diff
@@ -0,0 +1,220 @@
...
+        seq_len = spec.shape[2]
+        spec = spec.reshape(batch_size * n_chans, seq_len, self.freq_cutoff)
+
+        # 2. input encoding + Transformer over the sequence of frames.
+        h = self.input_embedding(spec)
+        z = self.transformer(h)  # (batch * n_chans, n_frames, hidden_dim)
+
+        # 3. pool over channels and frames, then classify.
+        z = z.reshape(batch_size, n_chans, seq_len, self.hidden_dim)
+        pooled = z.mean(dim=(1, 2))
```
julien-gadonneix, line 217: - In their paper, they keep an output per channel, is it compatible with your framework in brandecode?
- Also, it seems also a bit weird to me but before averaging over the time dimension, in their fine-tuning version of the model, they select the 10 center tokens: `outputs = outputs[:,middle-5:middle+5].mean(axis=1)`.
- In their classification head, they do not use a layer normalization layer, but is it something you keep in braindecode on purpose?

braindecode/modules/brainbert_modules.py
```diff
@@ -0,0 +1,216 @@
...
+    normalizing : str
+        ``"zscore"`` (per-bin z-score over time, upstream default) or ``"none"``.
+    """
+
+    def __init__(
+        self,
+        sfreq: float,
+        nperseg: int = 400,
+        noverlap: int = 350,
+        freq_cutoff: int = 40,
```
julien-gadonneix, line 62: Maybe you could call this `idx_freq_cutoff` because it is not a frequency, but an index to select the frequency from the spectrogram. The actual frequency cutoff is then around 200Hz.

julien-gadonneix (commented):
Hey,
Thanks for reimplementing this model !
I'm just jumping in with two minor comments as I went through the paper and original codebase :)

PR #1105, opened 2026-07-24
Add PopulationTransformer (PopT) iEEG/sEEG foundation model

adammounir:
Towards #1097.

Adds a braindecode-native port of **PopulationTransformer (PopT)** (Chau et al. 2024, [arXiv:2406.03044](https://arxiv.org/abs/2406.03044)), the third iEEG/sEEG model from #1097 after `Brant` (#1100) and `BrainBERT` (#1104).

### What it is

Unlike a per-channel encoder, PopT is a **population aggregator**: its input is a set of per-electrode feature vectors (typically the frozen embeddings of a channel-level foundation model such as `BrainBERT`) plus each electrode's integer anatomical coordinates. A `CLS` token summarises the population after a stack of Transformer encoder layers.

### Design

- Following the braindecode convention, the per-electrode feature vector plays the role of the `n_times` axis, so the model keeps the standard `(batch, n_chans, n_times)` signature (`n_chans` = number of electrodes, `n_times` = feature dimension, 768 for BrainBERT `stft` features).
- Electrode coordinates are read from `chs_info` and discretised to integer indices inside the model; without positions the electrodes fall back to distinct sequential indices.
- The input embedding, spatial position encoding and Transformer are ported weight-for-weight; only the classification head is a braindecode-native addition.

### Parity

The input embedding + Transformer are checked **bit-exact** against the upstream `PtModelCustom` (input embedding + spatial encoding + post-norm encoder), with the released `popt_brainbert_stft` weights copied over — `test_popt.py::test_encoder_is_bit_exact_with_upstream` (gated by `POPT_SRC`, skipped in CI without it). Reproduce with:

```bash
POPT_SRC=/path/to/PopulationTransformer pytest test/unit_tests/models/test_popt.py
```

### Notes

- The shared model suites (config / categorization / forward / serialization / HF roundtrip) pass for `PopulationTransformer`.
- Weight re-hosting on `braindecode/popt-pretrained` (so `PopulationTransformer.from_pretrained(...)` works out of the box) is a follow-up, as for the previous models.

cc @bruAristimunha

### DIFF
--- a/braindecode/models/__init__.py
+++ b/braindecode/models/__init__.py
@@ -40,6 +40,7 @@ from .meta_neuromotor import MetaNeuromotorHand
 from .msvtnet import MSVTNet
 from .mvpformer import MVPFormer
 from .patchedtransformer import PBT
+from .popt import PopulationTransformer
 from .reve import REVE
 from .sccnet import SCCNet
 from .shallow_fbcsp import ShallowFBCSPNet
@@ -123,6 +124,7 @@ __all__ = [
     "MSVTNet",
     "MVPFormer",
     "PBT",
+    "PopulationTransformer",
     "REVE",
     "SCCNet",
     "ShallowFBCSPNet",
--- a/braindecode/models/popt.py
+++ b/braindecode/models/popt.py
@@ -1,0 +1,253 @@
+"""PopulationTransformer: a self-supervised aggregator over intracranial electrodes.
+
+Port of PopulationTransformer (PopT, Chau et al. 2024) into a braindecode-native
+model. Upstream code and pretrained weights are released by the authors:
+
+* paper: https://arxiv.org/abs/2406.03044
+* code: https://github.com/czlwang/PopulationTransformer
+* weights: https://huggingface.co/PopulationTransformer/popt_brainbert_stft
+
+Unlike a per-channel encoder, PopT operates on a **population** of electrodes:
+its input is a set of per-electrode feature vectors (the frozen embeddings of a
+channel-level foundation model such as BrainBERT) plus each electrode's integer
+anatomical coordinates. A ``CLS`` token summarises the population after a stack
+of Transformer encoder layers. The input embedding, spatial position encoding and
+Transformer are ported weight-for-weight from the upstream reference (bit-exact,
+see ``scripts/popt_parity_check``); the classification head is a
+braindecode-native addition. The official checkpoint loads directly via
+``PopulationTransformer.from_pretrained("braindecode/popt-pretrained")``.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+import torch
+import torch.nn as nn
+
+from braindecode.models.base import EEGModuleMixin
+from braindecode.modules.popt_modules import (
+    _PopTHead,
+    _PopTInputEmbedding,
+    _PopTSpecPredictionHead,
+)
+
+
+class PopulationTransformer(EEGModuleMixin, nn.Module):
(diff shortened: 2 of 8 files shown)


adammounir:
### Parity & replicability

**Encoder parity.** The `test_encoder_is_bit_exact_with_upstream` gate is skipped in CI (needs `POPT_SRC`, the upstream clone). Ran locally: after copying upstream weights, encoder output matches to **max abs diff 9.5e-7** (`atol=1e-5`), spatial PE buffers to `atol=1e-6`. The ported input embedding + Transformer are bit-exact with upstream.

**Re-hosted weights (`braindecode/popt-pretrained`).** Re-ran the paper's Table 1 protocol (Brain Treebank, frozen BrainBERT features → PopT, 7 subjects, paper's fine-tuning recipe). Mean ROC-AUC:

| task | paper | re-hosted | scratch |
|---|---|---|---|
| volume | 0.84 | 0.82 | 0.55 |
| pitch | 0.69 | 0.75 | 0.53 |
| onset | 0.86 | 0.87 | 0.73 |
| speech | 0.89 | 0.82 | 0.60 |

All four tasks in the paper's range, pretrained≫scratch preserved everywhere (+0.14 to +0.27). Not bit-exact absolute numbers (single seed, subject subset) but a faithful functional reproduction.

bruAristimunha (changes requested):
I reviewed this exact candidate against the PopulationTransformer paper and
its BrainBERT dependency. I am requesting changes because the mandatory
paper-result gate is not met and cannot yet be executed faithfully.

The qualifying target is the primary-paper Table 1 BrainBERT + pretrained PopT
Volume result at 90 electrodes: ROC-AUC `0.87 +/- 0.03`, reported as mean +/-
SEM across subjects. Before any run, the exact electrode ranking, subject
splits, seeds/repeats, aggregation, PopT source/checkpoint lineage, and
compatible reuse rights must be frozen. The gate also depends on a faithful,
licensed BrainBERT checkpoint path, which PR #1104 does not yet provide.

The current implementation is a pooled-classification shell rather than the
complete paper pipeline: it does not integrate BrainBERT, has no padding mask
for variable electrode sets, and accepts a hidden-width configuration that
later crashes. Green synthetic shape tests therefore cannot establish the
published result.

Please keep this behind the BrainBERT gate, reproduce the selected Table 1 row
on one exact candidate SHA within a preregistered tolerance, and reduce the
library PR to the smallest faithful model/masking contract with focused tests.
Data acquisition, checkpoint conversion, and full experiments should remain
external and opt-in. No data or checkpoint bytes were downloaded because the
protocol and provenance gates are incomplete.

Reviewed head: `b717d90e452c3fe3319b641af3bd52d9f5d976b0`.

bruAristimunha:
Paper-replication gate at exact head b717d90e452c3fe3319b641af3bd52d9f5d976b0: hold behind #1104.

A bounded PopulationTransformer Volume run on the Iceberg NEMAR Brain Treebank adaptation completed with validation ROC-AUC 0.9167148370 and test ROC-AUC 0.9335834897 (BCE 0.5066329166). The score is not a paper reproduction: it is one subject/run, uses 91 released clean channels rather than the unpublished task-specific top 90, lacks the paper split/subject aggregation, uses a post-paper checkpoint, and has no proven NEMAR-to-original-HDF5 signal equivalence.

Merge requires BrainBERT first, strict recoverable checkpoint/source provenance and rights, the paper's electrode-selection/split protocol, and the reported aggregation. The current successful run is diagnostic evidence only.

Issue #424, opened 2022-10-18
Some questions about TCN


Ainavo:
Why braindecode TCN uses dropout2D instead of dropout?

![image](https://user-images.githubusercontent.com/57820731/196412709-d1ffbeb0-0163-468c-9b9b-6c1c8534279a.png)
[braindecode/model/tcn.py](https://github.com/braindecode/braindecode/blob/master/braindecode/models/tcn.py)
![image](https://user-images.githubusercontent.com/57820731/196412771-99c248cb-deeb-4186-84ae-1db694f11fda.png)
[TCN/tcn.py](https://github.com/locuslab/TCN/blob/2f8c2b817050206397458dfd1f5a25ce8a32fe65/TCN/tcn.py#L22)

martinwimpff:
Hi @Ainavo, Dropout and Dropout2d are different from each other. [Dropout2d](https://pytorch.org/docs/stable/generated/torch.nn.Dropout2d.html) randomly zeros out entire feature maps, whereas [Dropout](https://pytorch.org/docs/stable/generated/torch.nn.Dropout.html) randomly zeros out single elements. Does that answer you question?

Ainavo:
@martinwimpff  Thanks for your reply, I would like to know if applying Dropout2D on EEG data works better than Dropout?

martinwimpff:
Try it out and let us know ;) 
In general [this paper](https://arxiv.org/abs/1411.4280) describes the benefits of "spatial dropout".

Ainavo:
Thanks, I'll try it later

h1deOnBush:
Hi,
I tried to test the performence of tcn in bcic_iv_2a. But I don't know how to use this model, I simply replace
`model = ShallowFBCSPNet(
    n_chans,
    n_classes,
    input_window_samples=input_window_samples,
    final_conv_length='auto',
)`
in plot_bcic_iv_4_ecog_trial.py
to
`model = TCN(
    n_in_chans=n_chans,
    n_outputs=n_classes,
    n_blocks=2,
    n_filters=12,
    kernel_size=4,
    drop_prob=0.5,
    add_log_softmax=True
)`
but it just can't work.
And the error is
`RuntimeError: Expected target size [64, 1107], got [64]`
I dont't know how to fix it，I'd be appreciate if you can give me some advice about how to use tcn model in bcic_iv_2a.

PR #1111, opened 2026-07-29
Use Dropout1d for the channel-wise dropout in BDTCN, TCN, BENDR

tayal-sarthak:
Problem
`_TemporalBlock` in `braindecode/models/tcn.py` applies `nn.Dropout2d` to activations
shaped `(batch, channels, times)`. `_ConvEncoderBENDR` in `braindecode/models/bendr.py`
does the same after each `nn.Conv1d` stage. PyTorch routes 3D input to `Dropout2d`
through a deprecated compatibility path, which warns at each forward pass:

```
dropout2d: Received a 3D input to dropout2d and assuming that channel-wise 1D dropout
behavior is desired - input is interpreted as shape (N, C, L), where C is the channel dim.
This behavior will change in a future release to interpret the input as one without a batch
dimension, i.e. shape (C, H, W). To maintain the 1D channel-wise dropout behavior, please
switch to using dropout1d instead.
```

On current master a single training forward pass raises six of these warnings in
`BDTCN`, six in `BENDR`:

```python
import warnings, torch
from braindecode.models import BDTCN, BENDR

for cls in (BDTCN, BENDR):
    model = cls(n_chans=8, n_outputs=2, n_times=256, sfreq=100.0).train()
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        model(torch.randn(4, 8, 256))
    print(cls.__name__, len([w for w in caught if "dropout2d" in str(w.message)]))
# BDTCN 6
# BENDR 6
```

Beyond the noise, the announced reinterpretation reads dimension 0 as the channel axis.
Masking would then drop whole batch items in place of feature channels, changing the
training behaviour of these models with nothing surfacing at runtime.

So, the change

`nn.Dropout2d` becomes `nn.Dropout1d` at the call sites receiving 3D activations:

| File | Site | Models |
| --- | --- | --- |
| `braindecode/models/tcn.py` | `_TemporalBlock.dropout1`, `_TemporalBlock.dropout2` | `BDTCN`, `TCN` |
| `braindecode/models/bendr.py` | `_ConvEncoderBENDR` encoder stages | `BENDR`, `InterpolatedBENDR` |

Call sites receiving 4D activations keep `nn.Dropout2d`, namely `ATCNet`, `TIDNet`.

#424 records the reason `Dropout2d` sits in the TCN residual block. The goal is masking
complete channels in place of single points. The BENDR encoder mirrors the reference
`dn3_ext.py` implementation, which places dropout after `nn.Conv1d` with the same
channel-masking intent. `nn.Dropout1d` expresses that semantic directly on `(N, C, L)`
input, with no deprecation attached. It is available since torch 1.12, well below the
`torch>=2.0` floor declared in `pyproject.toml`

### DIFF
--- a/braindecode/models/bendr.py
+++ b/braindecode/models/bendr.py
@@ -1,3 +1,8 @@
+# Authors: Bruno Aristimunha <b.aristimunha@gmail.com>
+#          Sarthak Tayal <sarthaktayal2@gmail.com>
+#
+# License: BSD (3-clause)
+
 import copy
 
 import numpy as np
@@ -435,7 +440,9 @@ class _ConvEncoderBENDR(nn.Module):
                         padding=width
                         // 2,  # Correct padding for 'same' output length before stride
                     ),
-                    nn.Dropout2d(dropout),  # 2D dropout (matches paper specification)
+                    # channel-wise dropout (matches paper specification), on
+                    # (batch, channels, times) activations
+                    nn.Dropout1d(dropout),
                     nn.GroupNorm(
                         encoder_h // 2, encoder_h
                     ),  # Consider making num_groups configurable or ensure encoder_h is divisible by 2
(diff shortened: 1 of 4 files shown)


adammounir (approved):
hey @tayal-sarthak, I just verified locally, full test_models.py passes and the dropout2d warnings are gone; LGTM thank you 👍

PR #1112, opened 2026-07-31
predict_trials and cropped train scoring fail when the module is passed as a name or a class

tayal-sarthak:
`EEGClassifier.predict_trials`, `EEGRegressor.predict_trials` and the cropped train-side scoring callbacks read the `module` constructor argument instead than the initialized `module_` attribute. skorch stores the built model in `module_`, while `module` keeps whatever was passed in. When a model name or an uninstantiated class is passed, those code paths therefore operate on a string or a class instead of a model, and raise before doing any work.

This is important for braind ecode because passing a name or a class is the form the documentation recommends. The `EEGClassifier` docstring states that an uninstantiated class should be preferred when passing a torch module directly, and model names are the documented shorthand.

heres some of the affected code:

- `braindecode/eegneuralnet.py`, `_EEGNeuralNet.predict_with_window_inds_and_ys`, which calls `self.module.eval()` and `self.module.forward(...)`. This method is the one `CroppedTrialEpochScoring` and `CroppedTimeSeriesEpochScoring` call at the end of every epoch when `on_train=True`.
- `braindecode/classifier.py`, `EEGClassifier.predict_trials`, which passes `module=self.module`.
- `braindecode/regressor.py`, `EEGRegressor.predict_trials`, which passes `module=self.module`.

### Steps to reproduce

```python
import numpy as np

from braindecode.datasets import create_from_X_y

n_trials, n_chans, n_times = 6, 4, 400
X = np.random.RandomState(0).randn(n_trials, n_chans, n_times).astype(np.float32)
y = np.random.RandomState(1).randint(0, 2, n_trials)

probe = ShallowFBCSPNet(
    n_chans=n_chans, n_outputs=2, n_times=n_times, final_conv_length=2
)
probe.to_dense_prediction_model()
n_preds = probe.get_output_shape()[2]

windows = create_from_X_y(
    X,
    y,
    drop_last_window=False,
    sfreq=100.0,
    window_size_samples=n_times,
    window_stride_samples=n_preds,
)

clf = EEGClassifier(
    "ShallowFBCSPNet",
    module__n_chans=n_chans,
    module__n_outputs=2,
    module__n_times=n_times,
    module__final_conv_length=2,
    cropped=True,
    batch_size=2,
    train_split=None,
)
clf.initialize()
clf.predict_trials(windows)
```

the current bvehavior:

The snippet above fails with:

```
AttributeError: 'str' object has no attribute 'eval'
```

Passing `ShallowFBCSPNet` as a class instead of the name fails with:

```
TypeError: Module.eval() missing 1 required positional argument: 'self'
```

Fitting in cropped mode with a train-side scoring callback, for example `callbacks=["accuracy"]`, fails with the same `TypeError` at the end of the first epoch.

Expected behaviour

All three cases should run against the initialized model and return the trialwise predictions, exactly as they do today when an already instantiated module is passed.

and some extra context

Only an already instantiated module works today, which is why this has gone unnoticed. `examples/advanced_training/bcic_iv_4_ecog_cropped.py` instantiates the model before handing it to `EEGRegressor`, and `test_predict_trials` in `test/unit_tests/training/test_scoring.py` does the same, so no existing example or test exercises the broken form.

### DIFF
--- a/braindecode/classifier.py
+++ b/braindecode/classifier.py
@@ -228,8 +228,9 @@ class EEGClassifier(_EEGNeuralNet, NeuralNetClassifier):
             if return_targets:
                 return preds, X.get_metadata()["target"].to_numpy()
             return preds
+        self.check_is_fitted()
         return predict_trials(
-            module=self.module,
+            module=self.module_,
             dataset=X,
             return_targets=return_targets,
             batch_size=self.batch_size,
--- a/braindecode/eegneuralnet.py
+++ b/braindecode/eegneuralnet.py
@@ -1,5 +1,6 @@
 # Authors: Bruno Aristimunha <b.aristimunha@gmail.com>
 #          Pierre Guetschel <pierre.guetschel@gmail.com>
+#          Sarthak Tayal <sarthaktayal2@gmail.com>
 #
 # License: BSD (3-clause)
 
--- a/braindecode/eegneuralnet.py
+++ b/braindecode/eegneuralnet.py
@@ -126,7 +126,8 @@ class _EEGNeuralNet(NeuralNet, abc.ABC):
                 self._last_window_inds_ = None
 
     def predict_with_window_inds_and_ys(self, dataset):
-        self.module.eval()
+        # self.module can still be a name or a class, self.module_ is the built one
+        self.module_.eval()
         preds = []
         i_window_in_trials = []
         i_window_stops = []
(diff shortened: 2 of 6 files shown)


adammounir (approved):
hey @tayal-sarthak thank you for your contribution ! this fix corrects the name and class paths well (fails without the fix, tested on my side). Merging once CI is all green (just added minimal change on changelog). Thank you :)

tayal-sarthak:
> hey @tayal-sarthak thank you for your contribution ! this fix corrects the name and class paths well (fails without the fix, tested on my side). Merging once CI is all green (just added minimal change on changelog). Thank you :)

hey @adammounir , no worries, thank you! :)

PR #1113, opened 2026-07-31
FBLightConvNet ignores stride_factor, then dies with ZeroDivisionError on short windows

tayal-sarthak:
`braindecode.models.FBLightConvNet` takes a `stride_factor` argument, stores it on the instance, never reads it again. The temporal segmentation of this model is set by `win_len`, which is missing from the class docstring. The docstring lists `n_bands` with a default of 8. The constructor uses 9.

The reference implementation this model is adapted from, `Ma-Xinzhi/LightConvNet`, has no `stride_factor` argument. Its constructor reads:

```python
def __init__(self, num_classes=4, num_samples=1000, num_channels=22,
             num_bands=9, embed_dim=32, win_len=250, num_heads=4,
             weight_softmax=True, bias=False)
```

Segmentation happens through `out.reshape([*out.shape[0:2], -1, self.win_len])`. The parameter reads as a leftover from the sibling filterbank models `FBCNet`, `FBMSNet`, where `stride_factor` does drive the reshape.

ignored parameter

```python
from braindecode.models import FBLightConvNet

kw = dict(n_chans=8, n_outputs=3, n_times=1000, sfreq=250)
a = FBLightConvNet(stride_factor=1, **kw)
b = FBLightConvNet(stride_factor=999, **kw)
```

Both models come out with the same architecture, producing bit identical outputs under a fixed seed. Anyone tuning `stride_factor` on this model is tuning nothing. The same knob on `FBCNet` moves `final_layer.in_features` from 576 to 1440.

Crash on short windows

The temporal attention kernel is sized `n_times // win_len`. A window shorter than `win_len` makes that expression 0, the attention weight is allocated with a zero length axis, weight initialisation then divides by a zero fan:

```python
FBLightConvNet(n_chans=8, n_outputs=3, n_times=200, sfreq=250)
```

```
File "braindecode/models/fblightconvnet.py", line 181, in __init__
    self.attn_conv = _LightweightConv1d(
File "braindecode/models/fblightconvnet.py", line 287, in _init_parameters
    nn.init.xavier_uniform_(self.weight)
File "torch/nn/init.py", line 465, in xavier_uniform_
    std = gain * math.sqrt(2.0 / float(fan_in + fan_out))
ZeroDivisionError: division by zero
```

The traceback lands inside `torch/nn/init.py`, giving no hint that `win_len` is the knob to change. Any window under one second at 250 Hz hits this.

fix::

Deprecate `stride_factor`, keeping it in the signature to avoid breaking existing calls. Document `win_len`, including the constraint on `n_times`. Correct the `n_bands` default in the docstring. Raise a `ValueError` naming `n_times` alongside `win_len` when the window is too short.

### DIFF
--- a/braindecode/models/fblightconvnet.py
+++ b/braindecode/models/fblightconvnet.py
@@ -219,10 +219,12 @@ class FBLightConvNet(EEGModuleMixin, nn.Module):
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         """
         Forward pass of the FBLightConvNet model.
+
         Parameters
         ----------
         x : torch.Tensor
             Input tensor with shape (batch_size, n_chans, n_times).
+
         Returns
         -------
         torch.Tensor
--- a/braindecode/models/fblightconvnet.py
+++ b/braindecode/models/fblightconvnet.py
@@ -116,7 +116,7 @@ class FBLightConvNet(EEGModuleMixin, nn.Module):
         n_bands=9,
         n_filters_spat: int = 32,
         n_dim: int = 3,
-        stride_factor: int = 4,
+        stride_factor: Optional[int] = None,
         win_len: int = 250,
         heads: int = 8,
         weight_softmax: bool = True,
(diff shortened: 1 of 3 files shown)


adammounir (approved):
@tayal-sarthak  nice catch, clean deprecation + explicit ValueError, tests nail it and full model suite green locally on my side. LGTM thank you for this contribution !

tayal-sarthak:
@adammounir thanks for your kind words! :)

PR #1115, opened 2026-08-05
Make braindecode models directly scriptable with torch.jit.script

adityasingh2400:
`torch.jit.script(model)` currently fails on every model in the zoo with `ValueError: chs_info not specified`, long before `forward` is ever compiled. The test suite works around it by rebuilding each model as a plain `nn.Module` first (`convert_model_to_plain` in `test_integration.py`), but a user who follows the PyTorch docs and scripts a braindecode model directly gets an error that has nothing to do with scripting.

The root cause is in `EEGModuleMixin`. While building a module's concrete type, TorchScript reads every class attribute. It calls `getattr(mod, name, None)` for each name in `dir(type(mod))`, and the three-argument `getattr` only swallows `AttributeError`, so the `ValueError` raised by the signal-related properties escapes. Once that is out of the way, `mapping: Optional[Dict[str, str]]` fails too, because `base.py` uses `from __future__ import annotations` and TorchScript cannot resolve the resulting string annotation.

PyTorch provides `__jit_ignored_attributes__` for exactly this, so the fix declares both groups there. The one wrinkle is that TorchScript collects ignorable properties with `vars(type(module))`, which does not see properties inherited from a mixin, so `__init_subclass__` rebinds the effective property objects onto each concrete model class. That is a no-op at runtime, it is the identical descriptor reachable under the identical name, and the properties keep raising `ValueError` exactly as before. Rich `chs_info` metadata stays available in eager mode while scripted models use the already-resolved scalar channel count.

With this, 27 of the 63 direct model classes script, execute, and match eager output straight from `torch.jit.script(model)`, against zero before. The rest still fail for model-specific reasons unrelated to the base class, mostly `torch.nn.utils.parametrize` (which PyTorch explicitly rejects under scripting), `**kwargs` in `forward`, and einops. Those stay in scope of #732.

Tested with direct scripting, execution, eager parity, scripted save/load, configuration round-trips, real MNE-style `chs_info`, and inherited custom ignored-property coverage. The exhaustive registry scan is 27/63. Locally `test_models.py` (649 passed, 1 skipped), the full integration suite (648 passed, 121 skipped), Hugging Face tests (183 passed, 20 skipped), focused model/configuration/feature/correctness gates, and every configured changed-file hook are green.

Progress on #732.


adityasingh2400:
Thanks for approving the run. It caught a real problem and it was mine.

`pre-commit-codestyle` failed on one hook only:

```
codespell.......................Failed
braindecode/models/base.py:203: inferrable ==> inferable
```

That word is on a comment line this PR adds, and it does not appear anywhere on `master`, so nothing else is involved. `mypy` in the same run reported `Success: no issues found in 254 source files`, its output just looks noisy because of an `annotation-unchecked` note.

Fixed in `19082a5`, a single word. Pushing reset the workflows to `action_required` again, which I am sorry about, so this needs one more approval from you whenever convenient.

bruAristimunha (approved):
Approved at exact head 7d9f4715dfc236e148b772d7215d56c418850950 after independent max review. The 27-model direct TorchScript characterization covers script execution, eager parity, and BytesIO save/load parity without platform skips; the focused gate passed 28 tests, the broader model suite passed 1,714 tests (147 skipped), and current master was integrated additively.

PR #1120, opened 2026-08-10
Add LaBraM TopK SAE prototype

ShahVandit:
## Current Status

  The changes are contained in:

  braindecode/mech-interp/labram-activations.ipynb

  This is currently a demonstration intended to reproduce part of the workflow from Mechanistic Interpretability of EEG
  Foundation Models via Sparse Autoencoders (arXiv:2605.13930) using Braindecode’s LaBraM implementation and the public NMT Scalp EEG Dataset.

  The experiment was limited to 400 recordings because of Kaggle’s disk, memory, and runtime constraints. Intermediate preprocessing and activation files are deleted after use to remain within Kaggle’s available storage.

  ### Results

  - Fine-tuned LaBraM
      - Validation window AUROC: 0.7053
      - Test window AUROC: 0.6128
      - Test recording-level AUROC: 0.6696

  - Layer-11 Top-K SAE
      - Training activation tokens: 400,000
      - Input dimension: 200
      - Dictionary size: 1,600
      - Top-K value: 64
      - Reconstruction R²: 0.9578
      - Dead features: 0%
      - Average L0: 64

  - Native-head faithfulness
      - Window-level AUROC change: -0.0047
      - Recording-level AUROC change: -0.0040

The small AUROC changes suggest that replacing layer-11 activations with SAE reconstructions preserved most of the fine-tuned model’s task behavior in this run. Next steps are to  reproduce TCAV-based concept attribution, concept steering, spectral decoding back to EEG structure, and monosemanticity analysis.

### DIFF
--- a/braindecode/mech-interp/validate_sae_labram_nmt.py
+++ b/braindecode/mech-interp/validate_sae_labram_nmt.py
@@ -1,0 +1,711 @@
+"""Validate the sparse-autoencoder utilities on LaBraM and the NMT corpus.
+
+This follows Lehn-Schioler et al. (arXiv:2605.13930) on public data: fine-tune a
+pretrained EEG transformer on a clinical target, freeze it, fit a Top-K SAE to
+one layer's residual stream, then substitute the reconstruction back in and
+re-measure the task with the model's own head.
+
+LaBraM is used rather than a smaller transformer because its tokens are exact
+non-overlapping one-second patches. That property is what a spectral decoder
+needs later -- it learns ``token embedding -> FFT of the matching raw patch``,
+a target that only exists when tokens are time-aligned. NMT also carries
+``pathological``, ``age`` and ``gender``, three of the paper's five TCAV
+concepts.
+
+Everything imports from ``braindecode.visualization``, so a successful run
+validates the library code rather than a parallel implementation.
+
+The full NMT archive is ~13.4 GB and is downloaded on first use. Pass
+``--data-root`` to point at an existing copy.
+
+Usage
+-----
+    python validate_sae_labram_nmt.py --n-recordings 400
+    python validate_sae_labram_nmt.py --data-root /data/nmt --layer 11
+"""
+
+from __future__ import annotations
+
+import argparse
+import copy
+import json
+from pathlib import Path
+
+import numpy as np
+import torch
+from sklearn.metrics import roc_auc_score
+from torch import nn
+from torch.utils.data import DataLoader, Subset
+
+from braindecode.datasets import NMT
+from braindecode.models import Labram
+from braindecode.models.labram import LABRAM_CHANNEL_ORDER
+from braindecode.preprocessing import (
+    Preprocessor,
+    create_fixed_length_windows,
+    preprocess,
+)
+from braindecode.visualization import (
+    SparseAutoencoder,
+    capture_activations,
+    fit_sparse_autoencoder,
+    run_with_activation_substitution,
+    sae_diagnostics,
+)
+
+HF_REPO = "braindecode/labram-pretrained"
+SFREQ = 200
+EMBED_DIM = 200
+
+# NMT stores these names with the older 10-20 convention; LaBraM's 128-name
+# vocabulary only knows the modern ones, so T3/T4/T5/T6 must be renamed or they
+# resolve to no position embedding at all.
+CH_NAMES = [
+    "FP1",
+    "FP2",
+    "F7",
+    "F3",
+    "FZ",
+    "F4",
+    "F8",
+    "T3",
+    "C3",
+    "CZ",
+    "C4",
+    "T4",
+    "T5",
+    "P3",
+    "PZ",
+    "P4",
+    "T6",
+    "O1",
+    "O2",
+]
+RENAME = {"T3": "T7", "T4": "T8", "T5": "P7", "T6": "P8"}
+
+
+def parse_args():
+    p = argparse.ArgumentParser(description=__doc__)
+    p.add_argument(
+        "--data-root",
+        type=Path,
+        default=None,
+        help="directory holding Labels.csv; downloads if omitted",
+    )
+    p.add_argument("--cache", type=Path, default=Path("nmt_windows"))
+    p.add_argument("--n-recordings", type=int, default=400)
+    p.add_argument(
+        "--delete-raw-after-cache",
+        action="store_true",
+        help="delete --data-root once the window cache is written; "
+        "off by default, since it removes your data",
+    )
+    p.add_argument(
+        "--layers",
+        type=int,
+        nargs="+",
+        default=[11],
+        help="blocks to sweep; the fine-tune is shared across them",
+    )
+    p.add_argument(
+        "--ft-ckpt",
+        type=Path,
+        default=Path("labram_nmt_ft.pt"),
+        help="fine-tuned weights; reused when present",
+    )
+    p.add_argument(
+        "--expansions",
+        type=int,
+        nargs="+",
+        default=[8],
+        help="expansion ratios to sweep; k defaults to 8*E per cell",
+    )
+    p.add_argument("--k", type=int, default=None)  # defaults to 8 * expansion
+    p.add_argument("--finetune-epochs", type=int, default=10)
+    p.add_argument("--freeze-epochs", type=int, default=2)
+    p.add_argument("--sae-epochs", type=int, default=20)
+    p.add_argument(
+        "--sae-log-every",
+        type=int,
+        default=5,
+        help="print SAE loss/dead every N epochs; 0 to silence",
+    )
+    p.add_argument("--batch-size", type=int, default=16)
+    p.add_argument("--head-lr", type=float, default=1e-3)
+    p.add_argument("--full-lr", type=float, default=1e-4)
+    p.add_argument("--max-sae-tokens", type=int, default=2_000_000)
+    p.add_argument("--min-valid-auroc", type=float, default=0.75)
+    p.add_argument("--seed", type=int, default=42)
+    p.add_argument(
+        "--resample-steps",
+        type=int,
+        nargs="+",
+        default=[500],
+        help="resampling intervals to sweep, in optimizer steps",
+    )
+    p.add_argument("--out", type=Path, default=Path("sae_labram_nmt_results.json"))
+    return p.parse_args()
+
+
+def _corpus_root(dataset):
+    """Directory holding the extracted recordings, from a recording path.
+
+    ``NMT`` picks its own download location, so the corpus root is only
+    discoverable after loading. Returns ``None`` if the layout is unfamiliar,
+    in which case nothing is deleted.
+    """
+    paths = dataset.description["path"]
+    if paths.empty:
+        return None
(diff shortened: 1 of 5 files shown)


bruAristimunha (changes requested):
Thanks for putting together the public-data experiment. I reviewed the notebook as a Braindecode contribution and against the workflow it cites. This needs a redesign before it is mergeable, so I am moving it back to draft.

The blocking issues are:

1. Braindecode's authored tutorials are Sphinx-Gallery Python scripts under `examples/`; the docs workflow generates notebooks from those scripts. A 1.27 MB executed Kaggle notebook under `braindecode/mech-interp/` is neither an importable package module nor the project tutorial format. It contains 454 output objects, Papermill/Kaggle/widget metadata, and no Markdown cells. Please move the narrative to a gallery example and strip generated outputs and environment metadata.

2. The example cannot be part of the normal docs execution path as written. It installs Python and system packages, downloads and extracts the 13.4 GB NMT archive, deletes caches/directories, and defaults `FORCE_REBUILD = True`. Library documentation must not mutate the environment or delete user data. Use existing Braindecode dataset/download APIs, explicit user-controlled paths and opt-in cleanup, and make the documented data/runtime cost clear. The CI-facing example needs a small bounded or synthetic path; the full experiment can be an explicitly opt-in external reproduction script.

3. The experiment fails its own stated gate (`0.6696 < 0.75`) but only prints a warning and continues into SAE training and faithfulness analysis. The downstream interpretation therefore violates the notebook's own precondition. A gate must actually stop/skip the dependent stages.

4. Do not use held-out test performance to decide whether to continue the analysis. Select architecture, preprocessing, checkpoint, and any performance threshold from training/validation only, then evaluate the test set once. The text says recording-level AUROC is clinically meaningful, but the checkpoint is selected using window-level validation AUROC; use and justify one validation target consistently. Data-dependent scaling constants must also be estimated from the training split only.

5. Extract reusable behavior (activation capture/substitution and, if this belongs in Braindecode's public API, the Top-K SAE) into normal modules with focused unit tests. The current notebook duplicates capture helpers and leaves core correctness, hook cleanup, checkpoint round-trips, top-k bounds, dead-feature resampling, grouped splitting, and substitution behavior untested.

6. Record clean-room provenance precisely. The cited companion repository says it is not end-to-end reproducible and is PolyForm Noncommercial, while Braindecode is BSD-3-Clause. Please identify the paper/method sources used for each independently implemented component and confirm that no code was copied or adapted from the noncommercial repository. This is a provenance requirement, not an accusation.

7. Current pre-commit failures are real: the repository's JSON hook rewrites the entire notebook, and Ruff cannot parse cells 2/3 because of unexpected indentation. A changelog entry is also required.

The useful mergeable shape is: tested reusable code (if an API addition is intended) + a compact Sphinx-Gallery tutorial that runs on a bounded input + a separate, clearly documented opt-in full NMT reproduction path. Please also report results only after the validation gate passes; the current below-gate run should be described as a failed pilot, not evidence that the method reproduces the cited workflow.

ShahVandit:
Hi @bruAristimunha,

The notebook was just for my own runs. I had used Kaggle for prototyping. It’s temporary and not something I’m proposing to merge. The temporary `braindecode/mech-interp` directory is for the same reason; it’s mainly for demo scripts.

For the library code, I’ve moved the SAE implementation under `braindecode/visualization/` following your comments. I feel that a new `braindecode/interpretability/` may make more sense, since this goes beyond visualization and the next steps are spectral decoding, TCAV,concept steering and more features. I can move it once we have a better sense of how everything should be organized together. We can also potentially extend this to port [transformerlens](https://github.com/TransformerLensOrg/TransformerLens) here.

The example under `examples/advanced_training/` uses the published ShallowFBCSPNet checkpoint for BCI IV 2a subject 3, which `plot_interpretability.py` already loads, so no fine-tuning is needed at doc-build time. It also shows that the method isn’t limited to transformers; here the dictionary decomposes a convolutional feature map. This is mainly a demonstration of the workflow rather than a result; the actual validation is the LaBraM run.

| condition | AUROC | vs baseline |
|---|---:|---:|
| baseline | 0.895 | — |
| SAE | 0.888 | −0.007 |
| random SAE | 0.522 | −0.373 |
| zeroed layer | 0.500 | −0.395 |

`sae.py` adds a Top-K sparse autoencoder (Gao et al. 2024) with `get_config`/`from_config`, a `fit_sparse_autoencoder` training loop with Anthropic-style dead-feature resampling, `sae_diagnostics`, and two forward-hook primitives, `capture_activations` and `run_with_activation_substitution`, which remove their hooks in a `finally` block so nothing leaks if the forward pass raises. I've also written tests in `test/unit_tests/visualization/test_sae.py`.

To validate the method against the paper’s setup (`braindecode/mech-interp/validate_sae_labram_nmt.py`), I fine-tuned LaBraM on NMT (398 recordings, recording-grouped split) and fit the SAE on layer 11. The fine-tuned baseline is 0.774 window AUROC versus 0.939 reported in the paper. The difference seems expected given that the paper uses a private clinical corpus of 3,036 subjects with 27 channels and 60 s windows, while NMT gives me 398 recordings, 19 channels, and 15 s windows.

| E | features | k | dead | R² | ΔAUROC | random SAE |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 200 | 8 | 25.0% | 0.806 | +0.0013 | 0.483 |
| 2 | 400 | 16 | 57.0% | 0.871 | +0.0017 | 0.457 |
| 4 | 800 | 32 | 71.5% | 0.932 | +0.0007 | 0.533 |
| 8 | 1600 | 64 | 79.1% | 0.976 | +0.0008 | 0.542 |

Window-level test AUROC; baseline 0.774. The random-SAE column is a same-shape untrained dictionary, i.e. the control.

The main observation here is that faithfulness stays essentially unchanged across an 8× range of dictionary sizes, even though reconstruction R² increases substantially.

I haven’t added `api.rst`, `whats_new.rst`, or `references.bib` yet. I think those can be added later, once we’re sure how all of this combines and what the final structure should look like.

bruAristimunha:
Thanks for replacing the executed notebook with reviewable Python and focused tests. I re-reviewed the current exact head `6307d53ecafc538cf87cfde805c8e433a51f8da8`. The hook cleanup tests and the validation-before-SAE control flow are real improvements, but this still needs to remain draft and should not be repaired as one 2,233-line contribution.

The paper-result gate is not met. The least-cost LaBraM-specific target I could freeze from the cited paper is Figure 2's last transformer block (zero-based layer 11), `E=1`, `k=8`: SAE-substituted test AUROC `0.9377 ± 0.0191` across five subject-disjoint folds, against baseline `0.9386 ± 0.0171`. This target is not currently externally reproducible: the 3,036-subject BC-1 data, exact fine-tuned LaBraM checkpoints, and SAE caches are private; no immutable split manifest is published; and the paper and its public companion source disagree on the split, fine-tuning schedule, TopK semantics, activation cap, decoder normalization, resampling, evaluation head, and uncertainty calculation.

The NMT run is a useful exploratory pilot toward external validation, but it cannot satisfy that reproduction gate. It changes the cohort, windows, preprocessing, split, checkpoint lineage, and metric protocol. I am therefore not requesting an NMT download or Iceberg run from this PR. A future NMT experiment needs its own immutable inclusion/split/preprocessing/checkpoint manifest and must be labelled external validation.

There are also exact-head blockers independent of the unavailable paper artifacts:

- On cache-building runs, `validate_sae_labram_nmt.py` unconditionally overwrites and recursively deletes a user-derived preprocessing directory without proving that this run created and owns it.
- A stale fine-tuning checkpoint can supply `best_valid` without binding or re-evaluating the current dataset, split, preprocessing, model/weight revision, seed, or metric definition.
- Public SAE inputs silently coerce booleans/non-integral dimensions; empty or one-row training and non-finite normalization can emit plausible-looking NaNs; `batch_size=0` leaks a low-level error.
- The group splitter accepts invalid fractions and can silently omit samples whose group is NaN.
- `fit_sparse_autoencoder(seed=...)` mutates the caller's process-global PyTorch RNG.
- The source cites behavior from a PolyForm Noncommercial companion implementation while the proposed files are BSD-3-Clause. Please provide a component-by-component clean-room declaration or BSD-compatible permission before a public SAE implementation is considered.

Please split this maximally instead of expanding it further:

1. A fresh hook-primitives PR containing only independently sourced, provenance-documented `capture_activations`, `run_with_activation_substitution`, cleanup/nested-output tests, and a changelog entry; no SAE, dataset, training loop, or paper claim. The component-level clean-room or permission gate applies to these helpers too. Aim for well under 250 added lines.
2. A separate, provenance-cleared SAE core with strict finite/type/shape/state contracts and focused tests. It remains paper-gated and should not include an optimizer, dead-feature resampling, diagnostics, splitter, gallery, or data runner.
3. Training/resampling/diagnostics only after the core contract and paper target are independently qualified.
4. Keep the full LaBraM/NMT runner outside the package as an opt-in, non-destructive experiment with an immutable manifest. Remove the generic group splitter and network/data-heavy gallery from this PR.

Please also replace the stale PR body: it still describes a removed notebook and an older below-gate run. The current state should be recorded as a draft prototype/external-validation pilot, not a paper reproduction. No spectral decoding, TCAV, steering, or TransformerLens scope should be added while these foundations remain unresolved.

bruAristimunha:
Exact-head paper/scope audit (`6307d53ecafc538cf87cfde805c8e433a51f8da8`): **SPLIT; keep draft; do not merge this head.**

- The public NMT pilot is workflow transfer, not replication of the EEG-SAE paper result on the private 3,036-subject BC-1 cohort and private fine-tuned checkpoints.
- The companion EEG-SAE code is PolyForm Noncommercial 1.0.0, while this module declares BSD-3. A clean-room/source-lineage statement is required before extracting reusable code.
- The driver evaluates the held-out test set throughout the layer/expansion/resampling sweep and accepts unbound caches/checkpoints; select on train/validation and evaluate test exactly once.
- Keep the 13.4-GB NMT validation external with pinned dataset/checkpoint/environment manifests.
- Split the library work into a minimal SAE/hook utility with edge tests and a separate bounded gallery that is explicitly a workflow demonstration. The current `dead=0.50` output should be described as partial dictionary collapse, not evidence of specialization.

Under the current paper-result-before-merge policy, exact replication is unavailable because the paper data/checkpoints/caches are not released. The draft should remain blocked even if CI is green.

ShahVandit:
the activations PR as requested - https://github.com/braindecode/braindecode/pull/1138

ShahVandit:
The top-k mechanism is [Makhzani & Frey (ICLR 2014](https://arxiv.org/abs/1312.5663)), which the paper cites as [17]. I'd implement a minimal TopK SAE from those two: encode, decode, config round-trip, decoder normalisation, strict input contracts, and focused tests, with no optimizer, resampling, diagnostics, splitter, or data runner. It would cite neither the [reference paper](https://arxiv.org/abs/2605.13930) nor its companion code, and would make no claim about that paper's results. Is that an acceptable plan?

An SAE is a trained `nn.Module` with persisted weights. Should it live in `visualization/sae.py` following the `get_config`/`from_config` pattern from `models/base.py`, or does it belong in `modules/` or a new subpackage? @bruAristimunha

PR #1128, opened 2026-08-24
Write spec-compliant JSON in Hub zarr stores

### LINKED ISSUE #880: zarr.json format incorrect
It seems like the description and info attributes are not json formatted dicts in the `zarr.json` file. See [here](https://huggingface.co/datasets/braindecode/example_dataset-raw/blob/main/derivatives/braindecode/dataset.zarr/recording_0/zarr.json). 
I could not file in the repo where this file was generated

azrabano23:
Closes #880

## Bug

Since #978 removed `_sanitize_for_json`, the MNE `Info` dict is stored verbatim as a zarr attribute. MNE `Info` routinely contains NaN (e.g. unknown channel positions in `chs[i]["loc"]`), and zarr-python serializes attributes with Python's default `json` behavior, which emits bare `NaN` literals. Those are forbidden by [RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259#section-6) and by the [zarr v3 spec](https://zarr-specs.readthedocs.io/en/latest/v3/core/index.html) (metadata documents must be valid JSON), so any pushed dataset with unknown channel locations produces a `zarr.json` that strict JSON parsers and non-Python zarr readers reject — as seen in the [example dataset on the Hub](https://huggingface.co/datasets/braindecode/example_dataset-raw/blob/main/derivatives/braindecode/dataset.zarr/recording_0/zarr.json) linked in #880.

The same file also showed `raw_preproc_kwargs`/`window_kwargs`/`window_preproc_kwargs` stored as double-encoded JSON strings rather than JSON objects, which is the other part of the "not json formatted dicts" report in #880.

## Fix

- Reintroduce `_sanitize_for_json` in `hub_io.py` and apply it to the `info` attribute in all three save paths (raw, windows, EEG windows). NaN/Inf become `null`; the existing `_restore_nan_from_json` load path already converts them back to NaN, so the round trip stays lossless (`mne.Info.from_json_dict` gets proper NaN arrays).
- Store the preprocessing kwargs as native JSON objects instead of `json.dumps` strings; the loader still accepts the stringified form written by older braindecode versions, so existing Hub datasets keep loading.

## Tests

Three regression tests in `test/unit_tests/datasets/test_hub_integration.py`:

- `test_zarr_json_attrs_are_spec_compliant` — saves a dataset whose `Info` contains NaN channel locs and parses every generated `zarr.json` with a strict JSON parser that rejects `NaN`/`Infinity` literals (fails before this fix); also asserts the kwargs attrs are JSON objects, not strings.
- `test_nan_channel_locs_round_trip` — NaN channel locations survive save/load unchanged.
- `test_legacy_stringified_kwargs_still_load` — stores written with the old double-encoded kwargs format still load.

Full `test/unit_tests/datasets/test_hub_integration.py` module passes locally; ruff format/lint clean on the changed files.

### DIFF
--- a/braindecode/datasets/bids/hub.py
+++ b/braindecode/datasets/bids/hub.py
@@ -28,6 +28,7 @@ The format follows a BIDS-inspired sourcedata structure:
 # License: BSD (3-clause)
 
 import contextlib
+import copy
 import json
 import logging
 import tempfile
@@ -89,6 +90,10 @@ def _normalize_kwargs_for_json(value, field_name):
 
     try:
         converted = _convert(value)
+        if isinstance(converted, str):
+            raise ValueError(
+                "a root string is ambiguous with the legacy encoded format"
+            )
         return json.loads(json.dumps(converted, allow_nan=False))
     except (OverflowError, RecursionError, TypeError, ValueError) as error:
         raise ValueError(
@@ -712,9 +717,6 @@ class HubDatasetMixin:
         # Validate uniformity across all datasets using shared validation
         dataset_type, _, _ = hub_validation.validate_dataset_uniformity(self.datasets)
 
-        # Keep reference to first dataset for preprocessing kwargs
-        first_ds = self.datasets[0]
-
         # Normalize every JSON value before opening the output store. The cached
         # values below are the exact values handed to the write helpers.
         prepared_infos = []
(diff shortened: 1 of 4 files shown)


bruAristimunha (approved):
Reviewed the exact integrated head e673efb0539f2c59c46fa7ecf541445bd5e3c2d6. Local max review found no P0-P3 issues; Hub tests passed 112/112, preprocessing passed 367 with 14 established skips, and the configured static gates passed. Hosted exact-head CI remains required before merge.

Issue #1141, opened 2026-08-27
14 model files have no `# License:` header line


bruAristimunha:
## Summary

Most model files start with an `# Authors:` / `# License:` header block. Fourteen do not carry a `# License:` line at all, so their terms are only inferable from the repository default.

This is lower severity than the mismatches in the two issues filed alongside this one, but it shares a root cause: the header convention is a habit rather than something enforced.

## Files

- `braindecode/models/biot.py`
- `braindecode/models/contrawr.py`
- `braindecode/models/eegminer.py`
- `braindecode/models/eegsym.py`
- `braindecode/models/fbcnet.py`
- `braindecode/models/fbmsnet.py`
- `braindecode/models/ifnet.py`
- `braindecode/models/labram.py`
- `braindecode/models/luna.py`
- `braindecode/models/msvtnet.py`
- `braindecode/models/reve.py`
- `braindecode/models/sinc_shallow.py`
- `braindecode/models/sparcnet.py`
- `braindecode/models/tidnet.py`

(`braindecode/models/config.py` is also missing one but is not a model.)

`eegminer.py` is the notable case: `NOTICE.txt` already declares it CC BY-NC 4.0, but the file itself says nothing and its Hub metadata reports `bsd-3-clause`. Its terms currently live in exactly one place, and not the file they govern.

Reproduce:

```bash
for f in braindecode/models/*.py; do
  sed -n '1,25p' "$f" | grep -qE "^#\s*License:" || echo "$f"
done
```

## Suggested fix

Add the header block to each file, matching the surrounding convention:

```python
# Authors: <upstream authors, if a port> (original <name>)
#          <adapter> (Braindecode adaptation)
#
# License: <license>
```

For files that are independent implementations from a paper, BSD-3 is presumably right; for ports of released code the upstream license governs and should be named explicitly. `eegminer.py` should be reconciled with what `NOTICE.txt` already claims about it.

Worth pairing with a lightweight check so new models cannot land without one.

bruAristimunha:
## This is not a mechanical fix — evidence

I originally framed this as "add the missing header block", implying it could be done in one pass. Having fetched the actual `LICENSE` file from every upstream repository cited in `braindecode/models/*.py`, that framing was wrong. The information needed to write these headers is a per-file judgment, not a lookup.

Three distinct situations exist among the files without a header:

**1. The license is stated, just not in a header line.** These only need the text moved into the standard block:

[...]

Issue #1140, opened 2026-08-27
NOTICE.txt and README omit several non-BSD licensed files (incl. two CC BY-NC models)


bruAristimunha:
## Summary

`NOTICE.txt` lists exactly two files under "CC BY-NC 4.0 Licensed Files" (`eegminer.py`, `meta_neuromotor.py`), and `README.rst` names only "Creative Commons Attribution-NonCommercial 4.0 International" under Additional Components. Several other files in the tree carry non-BSD licenses in their own headers and appear in neither.

A redistributor reading `LICENSE` + `NOTICE.txt` — the documented place to look — would not learn that these files exist or that noncommercial terms attach to some of them.

## Files with a non-BSD `# License:` header that are absent from `NOTICE.txt`

**Noncommercial (the ones that actually matter):**

- `braindecode/models/emg2qwerty.py` — Creative Commons Attribution-NonCommercial 4.0 International
- `braindecode/models/brainmodule.py` — Attribution-NonCommercial 4.0 International

**Permissive but still not BSD-3:**

- `braindecode/models/ctnet.py` — MIT
- `braindecode/models/dance.py` — MIT
- `braindecode/models/medformer.py` — MIT
- `braindecode/models/tcformer.py` — MIT
- `braindecode/models/mvpformer.py` — Apache-2.0
- `braindecode/models/zuna.py` — Apache-2.0

Reproduce:

```bash
for f in braindecode/models/*.py; do
  L=$(sed -n '1,25p' "$f" | grep -m1 -E "^#\s*License:" | sed 's/^#\s*License:\s*//')
  case "$L" in ""|BSD*|bsd*) ;; *) grep -q "$(basename $f)" NOTICE.txt || echo "$f -> $L";; esac
done
```

## Suggested fix

- Add `emg2qwerty.py` and `brainmodule.py` to the existing CC BY-NC 4.0 section of `NOTICE.txt`, with the same derivative-provenance wording already used for `meta_neuromotor.py`.
- Add MIT and Apache-2.0 sections listing the remaining files, plus their license links.
- Extend the `README.rst` "Additional Components" paragraph so the licenses it names match what `NOTICE.txt` lists.

A small test that every non-BSD `# License:` header has a matching `NOTICE.txt` entry would keep this from drifting again as models are added.

bruAristimunha:
## `emg2qwerty.py` belongs under ShareAlike, not CC BY-NC 4.0

[...]

Issue #1139, opened 2026-08-27
Hub `license=` metadata defaults to bsd-3-clause for non-BSD models (incl. CC BY-NC ones)


bruAristimunha:
## Summary

`EEGModuleMixin.__init_subclass__` accepts a `license=` keyword (`braindecode/models/base.py:217`) and forwards it to the HuggingFace model-card metadata. It defaults to `bsd-3-clause`. Only one model in the library actually sets it, so every other model — including several that are **not** BSD-3 — publishes `bsd-3-clause` in its model card.

For the noncommercial models this is a licensing problem, not a cosmetic one: pushing them to the Hub advertises permissive terms for code that is CC BY-NC.

## Evidence

Read back at runtime from `cls._hub_mixin_info.model_card_data.license`, compared against each file's own `# License:` header:

| Model | File header says | Hub metadata |
|---|---|---|
| `BrainModule` | CC BY-NC 4.0 | `bsd-3-clause` |
| `EMG2QwertyNet` | CC BY-NC 4.0 | `bsd-3-clause` |
| `MetaNeuromotorHand` | CC BY-NC 4.0 | `bsd-3-clause` |
| `CTNet` | MIT | `bsd-3-clause` |
| `DANCE` | MIT | `bsd-3-clause` |
| `MEDFormer` | MIT | `bsd-3-clause` |
| `TCFormer` | MIT | `bsd-3-clause` |
| `MVPFormer` | Apache-2.0 | `bsd-3-clause` |
| `ZUNA` | Apache-2.0 | `bsd-3-clause` |

`EEGMiner` is listed in `NOTICE.txt` as CC BY-NC 4.0 but has no header line and also reports `bsd-3-clause`.

Reproduce:

```python
import inspect
import braindecode.models as M

for name in sorted(M.__all__):
    cls = getattr(M, name, None)
    if inspect.isclass(cls) and hasattr(cls, "_hub_mixin_info"):
        print(name, cls._hub_mixin_info.model_card_data.license)
```

## Suggested fix

Add the correct `license=` to each affected class declaration, e.g.

```python
class EMG2QwertyNet(EEGModuleMixin, nn.Module, license="cc-by-nc-4.0"):
```

A test asserting that a class's `license=` metadata is consistent with its file's `# License:` header would stop this recurring. Related: the NOTICE/README gaps and the missing-header issue filed alongside this one.

bruAristimunha:
## Correction and expansion — verified against upstream LICENSE files

[...]

PR #1143, opened 2026-08-27
Correct license metadata, NOTICE and README for non-BSD model files

bruAristimunha:
Fixes the parts of #1139, #1140 and #1141 that can be settled from evidence, and deliberately leaves the parts that need a per-file provenance judgement.

## How the values were established

Every GitHub reference in `braindecode/models/*.py` was extracted, all 40 distinct upstream repositories were shallow-cloned, and their licence files were read from disk. The GitHub API's `license` field is not sufficient for this: it returned `NOASSERTION` for the two upstreams whose licences matter most, because neither Creative Commons texts nor the CBCR licence are in its classifier.

## What this changes

**1. `EMG2QwertyNet` was claiming the wrong licence.** Its header said CC BY-NC 4.0 and "Inherits CC BY-NC 4.0". The `LICENSE` in `facebookresearch/emg2qwerty` reads `Attribution-NonCommercial-ShareAlike 4.0 International`. ShareAlike is a stronger obligation, so the file was advertising weaker terms than it actually carries. Corrected to CC BY-NC-SA 4.0.

The two Meta repositories genuinely differ — `generic-neuromotor-interface` really is plain CC BY-NC 4.0, so `meta_neuromotor.py` was correct as written. That is presumably how the mistake arose.

**2. Non-BSD models published `bsd-3-clause` in their Hub model card.** `EEGModuleMixin.__init_subclass__` accepts a `license=` keyword and forwards it to the model-card metadata, defaulting to `bsd-3-clause`. No model set it, so noncommercial models advertised permissive terms. Now set for the twelve models whose own headers already declare a non-BSD licence:

| License | Models |
|---|---|
| `mit` | `CTNet`, `DANCE`, `MEDFormer`, `TCFormer`, `IFNet` |
| `apache-2.0` | `MVPFormer`, `ZUNA`, `LUNA` |
| `cc-by-nc-4.0` | `BrainModule`, `EEGMiner`, `MetaNeuromotorHand` |
| `cc-by-nc-sa-4.0` | `EMG2QwertyNet` |

The remaining 52 models are untouched and still report `bsd-3-clause`.

**3. `NOTICE.txt` recorded two of the twelve.** It listed only `eegminer.py` and `meta_neuromotor.py`. It now has sections for CC BY-NC 4.0, CC BY-NC-SA 4.0, MIT and Apache-2.0, with matching licence links.

Its BSD-3 statement also read "All files within the `src/` directory". There is no `src/` directory in this repository — the package is `braindecode/` — so the notice's central claim pointed at nothing and did not carve out the noncommercial files listed directly beneath it. Corrected, with the missing exception clause added.

**4. `README.rst`** named only CC BY-NC 4.0 under Additional Components. It now names all four and points at `NOTICE` for the per-file list, without a hardcoded count that would go stale.

**5. Three files stated a licence in prose but had no header line** — `luna.py` (Apache-2.0), `ifnet.py` (MIT), `eegminer.py` (CC BY-NC 4.0). The existing claim was moved into the standard header block. No new claim is introduced.

**6. Three upstream URLs were dead** because a sentence-ending period had been swept into the link: `eegconformer.py`, `eegsimpleconv.py`, `sstdpn.py`.

## What this deliberately does NOT change

Around 25 model→upstream references show a licence delta — files declaring BSD-3 while their upstream is MIT, Apache-2.0, GPL or noncommercial. **None of those are touched here**, because resolving them means asserting whether each file was written from the paper or adapted from the code, and only the author of each port knows. A clean-room reimplementation inherits nothing; several files already assert exactly that about themselves. A confidently wrong licence header is worse than a missing one, because people rely on headers.

The two that most deserve attention, both left alone:

- `eegconformer.py` declares BSD-3; `eeyhsong/EEG-Conformer` is **GNU GPL v3**. GPL is copyleft, so if that implementation derives from the code this is an incompatibility rather than a notice defect. Mitigating detail: the only reference to that repository inside the file is a hot-linked figure URL, and the file makes no derivation claim either way.
- `tsinception.py` declares BSD-3; `deepBrains/TSception` is **CBCR License 1.0** — BSD-3 text with "for non-commercial purpose" substituted into the grant.

Nine upstreams declare no licence at all, so their derivatives have nothing to inherit and no permission to rely on: `eegitnet`, `eegsimpleconv`, `fbmsnet`, `msvtnet`, `sinc_shallow`, `sstdpn`, `syncnet`, `tidnet`, and `signal_jepa` via BENDR. Confirmed two ways — a sparse clone finding no licence file, and the contents API listing none at the repository root.

Full evidence, tiered by severity, is in #1141.

## Also worth a maintainer's attention

`eegminer.py` carries a patent notice — GB2609265, Cogitat Ltd, "learnable filters for EEG classification". A patent is a separate grant from a copyright licence, CC BY-NC 4.0 conveys no patent rights, and neither `NOTICE.txt` nor the README mentions it. Out of scope here, but it is a larger question than any of the header mismatches this PR fixes.

## Verification

- Hub metadata read back for all 65 models: the twelve above report their new values, the other 52 still report `bsd-3-clause`.
- `pre-commit` green on all 17 changed files (`mypy`, `ruff`, `codespell`, `docstrfmt`, `sphinx-lint`, `isort`).
- Full `test/unit_tests/models/` suite run against the branch.

Reproduce the upstream side with:

```bash
git clone --depth 1 --filter=blob:none --no-checkout https://github.com/<slug>.git d
git -C d sparse-checkout set --no-cone '/LICENSE*' '/COPYING*' '/NOTICE*'
git -C d checkout && head -3 d/LICENSE*
```

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -12,6 +12,7 @@ The following components are licensed under the Creative Commons Attribution-Non
 
 - `braindecode/models/eegminer.py`
 - `braindecode/models/meta_neuromotor.py`
+- `braindecode/models/brainmodule.py`
 
 As well as class later imported into the `braindecode.models.module` named as GeneralizedGaussianFilter.
 
@@ -19,7 +20,45 @@ The `meta_neuromotor.py` file is a derivative of
 `facebookresearch/generic-neuromotor-interface`, released by Meta Platforms,
 Inc. under CC BY-NC 4.0, and inherits the same noncommercial terms.
 
+### CC BY-NC-SA 4.0 Licensed Files
+
+The following components are licensed under the Creative Commons
+Attribution-NonCommercial-ShareAlike 4.0 International License:
+
+- `braindecode/models/emg2qwerty.py`
+
+The `emg2qwerty.py` file is a derivative of `facebookresearch/emg2qwerty`,
+released by Meta Platforms, Inc. under CC BY-NC-SA 4.0, and inherits the
+same noncommercial ShareAlike terms.
+
+### MIT Licensed Files
+
+The following components are licensed under the MIT License:
+
+- `braindecode/models/ctnet.py`
+- `braindecode/models/dance.py`
+- `braindecode/models/medformer.py`
+- `braindecode/models/tcformer.py`
+- `braindecode/models/ifnet.py`
+
+### Apache-2.0 Licensed Files
+
+The following components are licensed under the Apache License 2.0:
+
+- `braindecode/models/mvpformer.py`
+- `braindecode/models/zuna.py`
+- `braindecode/models/luna.py`
+
 ## License Links
 
 - [BSD-3-Clause License](https://opensource.org/licenses/BSD-3-Clause)
 - [CC BY-NC 4.0 License](https://creativecommons.org/licenses/by-nc/4.0/)
+- [CC BY-NC-SA 4.0 License](https://creativecommons.org/licenses/by-nc-sa/4.0/)
+- [MIT License](https://opensource.org/licenses/MIT)
+- [Apache-2.0 License](https://www.apache.org/licenses/LICENSE-2.0)
+
+## Note
+
+This list covers files whose own headers declare a license other than
+BSD-3-Clause. Per-file provenance review of the remaining models is
+tracked separately.
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -4,7 +4,8 @@
 
 ### BSD-3-Clause Licensed Files
 
-All files within the `src/` directory are licensed under the BSD-3-Clause License.
+All files within the `braindecode/` package are licensed under the BSD-3-Clause
+License, except for those listed in the sections below.
 
 ### CC BY-NC 4.0 Licensed Files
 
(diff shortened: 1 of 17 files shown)


bruAristimunha:
Note on the mechanism used here, since it is recent: #1134 ("Fix license metadata without Hub dependency", merged as `1113d45b`) moved the `license=` handling so a model can declare its licence whether or not the optional `huggingface_hub` dependency is installed. Its test constructs a model with `license="cc-by-nc-sa-4.0"` under a stubbed-out Hub.

This PR is the first to populate that mechanism on real models. It adds no new machinery — it only supplies values to the keyword #1134 made robust, and the value it sets on `EMG2QwertyNet` is the same one that test uses as its example.

Verified after the change, across all 65 models: the twelve listed above report their new licences, and the other 52 still report `bsd-3-clause`.

Separately, and relevant to how the deltas left out of this PR should be read: `eegconformer.py` was introduced by commit `dfbcf36f`, authored by `eeyhsong`, who owns the GPL-3.0 upstream `eeyhsong/EEG-Conformer`. A copyright holder may license their own work under any terms, so that contribution carries no GPL obligation. Several other flagged files follow the same pattern. Details in #1141 — but it is worth recording that reasoning in the files themselves, otherwise every future audit regenerates the same false alarms.

PR #1154, opened 2026-09-07
Fix CombinedConv bias folding when a dimension is 1

julien-gadonneix:
`CombinedConv.forward` folds the time and spatial biases with

```python
self.conv_spat.weight.squeeze().sum(-1).mm(time_bias.unsqueeze(-1)).squeeze()
```

`conv_spat.weight` is `(n_filters_spat, n_filters_time, 1, in_chans)` and the bare `squeeze()` is only meant to drop the singleton kernel dimension. When `in_chans`, `n_filters_time` or `n_filters_spat` is also 1 it drops those too, so `sum(-1)` returns a vector and `mm` fails with `RuntimeError: self must be a matrix`.

This makes `ShallowFBCSPNet` and `Deep4Net` unusable on single-channel data:

```python
ShallowFBCSPNet(n_chans=1, n_times=1536, n_outputs=4)  # RuntimeError
```

Squeezing explicit dimensions (`squeeze(2)` and `squeeze(-1)`) fixes it. The output is unchanged wherever the current code works, and matches the sequential `conv_spat(conv_time(x))` reference in every case.

`test_combined_conv` is parametrized over the singleton shapes; it fails on master and passes here.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/braindecode/modules/convolution.py
+++ b/braindecode/modules/convolution.py
@@ -191,11 +191,13 @@ class CombinedConv(nn.Module):
             time_bias = self.conv_time.bias
             if time_bias is None:
                 raise RuntimeError("conv_time.bias is None despite bias_time=True")
+            # squeeze only the singleton kernel dimension: a bare squeeze() also
+            # drops n_filters_spat / n_filters_time / in_chans when they are 1
             calculated_bias = (
-                self.conv_spat.weight.squeeze()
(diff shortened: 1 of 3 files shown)


bruAristimunha (approved):
Looks good for me @julien-gadonneix

PR #1170, opened 2026-09-16
Add DIVER-1 any-variate iEEG foundation model

julien-gadonneix:
Adds `DIVER1` (encoder + classification head) from [DIVER-1](https://arxiv.org/abs/2512.19097).

**Model.** An any-variate transformer for intracranial EEG. A small patch CNN plus a spectral branch turn each channel into time patches, and attention then runs jointly over all electrodes and time steps, so an arbitrary number and layout of channels is supported. Channel identity comes from `chs_info` metadata instead of a fixed montage: sinusoidal electrode coordinates, modality/subtype embeddings, a sliding-window spatio-temporal conditional positional embedding, rotary temporal offsets and a learned same/cross-channel attention bias. The pretraining-only parts (masking, MDRO heads) are not ported, and no public checkpoint exists yet, so this is architecture-only.

**Checked against the source.** The architecture was double-checked against both the paper and the authors' original codebase: transplanting reference weights into this port reproduces the reference forward pass bit-exactly, with matching encoder parameter counts.

**Tests.** `test_integration.py` and `test_return_features.py` pass for `DIVER1`, as do the categorization / modality / registry suites, and `pre-commit` (ruff, mypy, codespell, sphinx-lint) is clean. TorchScript is exempted, since `einops` and the polymorphic `return_features` return type are not scriptable, consistent with the existing exemptions.

Guidance and advice would be very welcome. Happy to adjust anything.


julien-gadonneix:
Validation against the authors' reference implementation, ahead of braindecode's own suites:

- **Numerical equivalence** — transplanting a randomly initialised reference state dict into this port reproduces the reference encoder features bit-exactly (`max|diff| = 0.0`), for both the 1 s and 0.1 s patch variants, with and without electrode coordinates. This needs the reference's attention dropout zeroed, since it stays active in eval mode there; this port gates it on `self.training`.
- **Parameter counts** (encoder-only, 1 s / 0.1 s) — Tiny 12.67M / 12.66M, Small 50.64M / 50.62M, Base 113.92M / 113.88M: matching the reference exactly, and Table 4 once the pretraining heads and mask token are excluded.
- **Structural** — features are channel-permutation equivariant and pooled logits invariant with the position embedding off, and `chs_info` yields the expected modality/subtype ids for `eeg`/`seeg`/`ecog`.

No checkpoint exists to verify against (the reference ships a placeholder for `weights/ieeg_pretrained_weights.pt`), so this is equivalence of the architecture, not of trained behaviour. Two paper-vs-code disagreements are resolved in favour of the code and noted in the docstring — corrections very welcome.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+
+Reimplementation of DIVER-1 (Han et al., 2025), "DIVER-1: Scaling Intracranial
+EEG Foundation Models for Transferable Representations". The architecture is
+transcribed from the authors' reference implementation, whose Transformer
+encoder is adapted from Salesforce's MOIRAI / ``uni2ts`` (Copyright Salesforce,
+Inc.), released under the Apache License, Version 2.0. The license this port
+should carry follows from that provenance but is still to be confirmed with the
+braindecode maintainers.
+The braindecode reimplementation is pure-PyTorch (no ``mup``, no ``jaxtyping``)
+and covers the downstream encoder only.
```
bruAristimunha: we can cut this

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
+# Authors: Julien Gadonneix <145470783+julien-gadonneix@users.noreply.github.com>
+"""DIVER-1: an any-variate iEEG foundation model.
+
+Reimplementation of DIVER-1 (Han et al., 2025), "DIVER-1: Scaling Intracranial
+EEG Foundation Models for Transferable Representations". The architecture is
+transcribed from the authors' reference implementation, whose Transformer
+encoder is adapted from Salesforce's MOIRAI / ``uni2ts`` (Copyright Salesforce,
+Inc.), released under the Apache License, Version 2.0. The license this port
+should carry follows from that provenance but is still to be confirmed with the
+braindecode maintainers.
```
bruAristimunha: I think Labram is a similar case, with a link list of licenses. I think here we assume the license is from the head, and at most we put some small notes saying that the person needs to check the original implementation.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+
+# Channel-modality vocabulary of the reference ``ChannelTypeEmbedding``: slot 0
+# is scalp EEG, slot 1 intracranial EEG.
+_MODALITIES = ("EEG", "iEEG")
+# Electrode sub-modality vocabulary of the reference ``ChannelSubTypeEmbedding``:
+# ECoG grids and strips, and SEEG depth electrodes.
+_SUBTYPES = ("grid", "strip", "depth")
+# FIFF channel-kind codes (mne.io.constants.FIFF), so chs_info can be read
+# without importing mne.
+_FIFF_SEEG, _FIFF_DBS, _FIFF_ECOG = 802, 803, 902
```
bruAristimunha: Is there any specific reason why not import from MNE? In Braindecode, we have a dependency on MNE.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+    :bdg-info:`Attention/Transformer` :bdg-danger:`Foundation Model`
+    :bdg-dark-line:`Channel`
+
+    .. figure:: ../_static/model/diver1_arch.png
+        :align: center
+        :alt: DIVER-1 architecture overview
+        :width: 1000px
+
+        DIVER-1 architecture, pretraining and downstream evaluation pipeline,
+        reproduced from [Han2025]_.
```
bruAristimunha, line 94: great!

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+            self.n_chans * self.n_patches * d_model if pooling == "flatten" else d_model
+        )
+        self.final_layer = nn.Linear(head_in_features, self.n_outputs)
+
+    def reset_head(self, n_outputs):
+        """Replace the linear classification head for a new ``n_outputs``."""
+        self._n_outputs = n_outputs
+        self.final_layer = nn.Linear(self.final_layer.in_features, n_outputs)
+
+    def _resolve_chan_pos(self, chan_pos) -> tuple[torch.Tensor, torch.Tensor]:
```
bruAristimunha: I think we have an util functions for this.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        """Resolve the electrode modality and sub-modality of each channel.
+
+        Explicit arguments win; otherwise both are derived from the ``"kind"``
+        entries of ``chs_info``, accepting either MNE's FIFF integer codes or
+        plain strings. Unlike the reference implementation, which infers the
+        sub-modality from a site-specific table of electrode-group name
+        prefixes, we key off the channel kind: SEEG contacts are depth
+        electrodes and ECoG contacts are grids. Pass ``chan_subtype`` to
+        distinguish grids from strips.
+        """
```
bruAristimunha: My small push is to see if we can simplify this using mne, or if it gonna be useful in the future move for the utils.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        if kind in ("seeg", "dbs"):
+            return "depth"
+        if kind == "ecog":
+            return "grid"
+    elif isinstance(kind, int):
+        if kind in (_FIFF_SEEG, _FIFF_DBS):
+            return "depth"
+        if kind == _FIFF_ECOG:
+            return "grid"
+    return "unknown"
```
bruAristimunha: let's try to use mne for this

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """Map ``(batch, n_chans, n_patches, patch_size)`` to ``d_model`` tokens."""
+        batch, n_chans, n_patches, _ = x.shape
+        x = F.pad(x, self.pad)
+        # All (channel, patch) pairs are encoded independently: fold them into
+        # the height axis of a single-channel 2D convolution.
+        x = self.proj_in(x.reshape(batch, 1, n_chans * n_patches, -1))
+        x = x.permute(0, 2, 1, 3)  # (batch, chans * patches, hidden, out_size)
+        return x.reshape(batch, n_chans, n_patches, self.d_model)
```
bruAristimunha: We prefer einops layers here, to be more explicit with the operation, naming the axis, and trying to link with the paper definition. 

It is okay to add an extra layer just for channel manipulation, if it will make things clear.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        super().__init__()
+        if kernel_size % 2 == 0:
+            raise ValueError(f"cnn_kernel_size must be odd, got {kernel_size}.")
+        if depth < 1:
+            raise ValueError(f"cnn_depth must be at least 1, got {depth}.")
+
+        padded = 1 << (patch_size - 1).bit_length()
+        if stride is None:
+            stride = padded // (8 if patch_size >= 100 else 16)
+        if stride < 1 or padded % stride:
```
bruAristimunha, line 662: tell why the 100 or 16, but without being too verbose.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        super().__init__()
+        self.spectral_proj = nn.Sequential(
+            nn.Linear(d_model // 2 + 1, d_model), nn.Dropout(drop_prob)
+        )
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # rfft is not implemented for half precision: compute it in float32 and
+        # cast the amplitudes back.
+        amplitude = torch.fft.rfft(x.float(), dim=-1, norm="forward").abs()
+        return self.spectral_proj(amplitude.to(x.dtype))
```
bruAristimunha: Can you unstack the operations? One line for each operation, I got that it is equivalent, but just to be more explicit.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+    scale : float
+        Multiplier applied to the coordinates before encoding.
+    """
+
+    def __init__(
+        self, d_model: int, temperature: float = 2000.0, scale: float = 1 / 256
+    ):
+        super().__init__()
+        n_dim = 3
+        self.n_feats = d_model // n_dim // 2 * 2
```
bruAristimunha, line 835: We need to be a little more didactic and unstack the operations.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        # Full-height, window-wide patches of the (channel, patch) grid. The
+        # feature axis is folded into the batch so unfold/fold only slide over
+        # time; zero-padding by window - 1 keeps a window centred on every
+        # patch.
+        unfold_kwargs = {
+            "kernel_size": (n_chans, self.window),
+            "stride": (1, 1),
+            "padding": (0, self.window - 1),
+        }
+        flat = rearrange(x, "batch chans patches dim -> (batch dim) 1 chans patches")
```
bruAristimunha: define as an einops layer, and you can get back torch script compatibility.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        inner_dim = x.shape[-1]
+        # Full-height, window-wide patches of the (channel, patch) grid. The
+        # feature axis is folded into the batch so unfold/fold only slide over
+        # time; zero-padding by window - 1 keeps a window centred on every
+        # patch.
+        unfold_kwargs = {
+            "kernel_size": (n_chans, self.window),
+            "stride": (1, 1),
+            "padding": (0, self.window - 1),
+        }
```
bruAristimunha: Let's define this in the init?

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+            value,
+            attn_mask=bias,
+            dropout_p=self.drop_prob if self.training else 0.0,
+        )
+        return self.out_proj(
+            rearrange(out, "batch heads seq dim -> batch seq (heads dim)")
+        )
+
+
+class _RotaryEmbedding(nn.Module):
```
bruAristimunha, line 1182: I think we have this already implemented here, but we need to double-check.

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+        self.fc1 = nn.Linear(d_model, d_ff, bias=False)
+        self.fc_gate = nn.Linear(d_model, d_ff, bias=False)
+        self.fc2 = nn.Linear(d_ff, d_model, bias=False)
+        self.activation = activation()
+        self.dropout1 = nn.Dropout(drop_prob)
+        self.dropout2 = nn.Dropout(drop_prob)
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        hidden = self.activation(self.fc_gate(x)) * self.fc1(x)
+        return self.dropout2(self.fc2(self.dropout1(hidden)))
```
bruAristimunha: unstack the operation, could be useful later to inspect the layers

braindecode/models/diver1.py
```diff
@@ -0,0 +1,1250 @@
...
+    eps : float
+        Term added to the mean square for numerical stability.
+    """
+
+    def __init__(self, normalized_shape: int, eps: float = 1e-5):
+        super().__init__()
+        self.eps = eps
+        self.weight = nn.Parameter(torch.ones(normalized_shape))
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
bruAristimunha: seems the perfect excuse that i was looking to increase the pytorch version :) so, we can delete.

bruAristimunha (commented):
Small details, and please don't forget to put your name in the citation.cff please.

braindecode/models/diver1.py
julien-gadonneix: I will correct it !

braindecode/models/diver1.py
julien-gadonneix: Indeed, I will use it but I had to add a `fill_missing` argument to it that defaults to `False`. When `True`, it outputs `nan` for the channels without info or loc

braindecode/models/diver1.py
julien-gadonneix: I moved the "generalizable" part that extract the MNE channel types to utils. I shortened and kept in the `__init__` function some of the logic or grouping that seemed specific to this model.
Slight modification from the original implementation: they distinguish grid and strip electrodes which MNE (and thus this implementation) doesn't

braindecode/models/diver1.py
julien-gadonneix: Here it is a bit hard-coded. In the paper they only have two possible `patch_size` that are 0.1 or 1 second. And they adapt some hyperparams based on this choice such as this stride. I added a comment to clarify it

braindecode/models/diver1.py
julien-gadonneix: It is also used by other models (e.g. LUNA) but always redefined. Do you want me to factor it out in the modules?

bruAristimunha: Let's import what is possible, and we open an issue to remove it in another PR.

PR #1171, opened 2026-09-17
Add BaRISTA model

julien-gadonneix:
Architecture verified against original codebase and paper.
Tests to load pretrained weights and reproduce results on the way.


julien-gadonneix:
# BaRISTA pretrained weight loading
It corresponds. The released checkpoint is a flat state dict of the pretraining model, and nothing but module naming separates it from the braindecode port: after renaming, every ported tensor matches a model tensor by name and shape, with zero left over. Coverage is 839,144 of 839,286 parameters — 100.0% to one decimal.

## What gets renamed
| Count | Reference name | Ported name |
|---|---|---|
| 60 | `backbone.layers.N.attention.*` | `backbone.layers.N.self_attn.*` |
| 24 | `tokenizer.temporal_encoder.feature_extractor.net.N.*` | `temporal_encoder.blocks.N.*` |
| 1 | `tokenizer.temporal_pooler.final_layer.weight` | `temporal_pooler.weight` |
| 1 | `tokenizer.spatial_encoder.subcomponent_embeddings.0.weight` | `spatial_emb.tables.0.weight` |

## What is new
- `final_layer.weight` / `final_layer.bias` (2, 64), (2,) — the task head, sized by n_outputs, so necessarily absent from a pretraining checkpoint.
- `token_pooling.weight` (1, 12) — the learned read-out over the 12 time patches, shaped by the window.

## What is not used
12 tensors, one per transformer layer: backbone.layers.N.attention.rotary_emb.inv_freq. These are rotary position frequencies, which the port recomputes from the layer geometry rather than storing, so dropping them loses nothing.

julien-gadonneix:
Paper result reproduction
image
Paper result on sentence onset classification: 0.853 (0.007 sem)
Reproduction: 0.910 (0.003 sem)

Difference may come from the fact that they enforce balanced classes over all, whereas in the reproduction, the classes are imbalanced but I use a training sampler to have balanced training batches and ensure that in test and validation splits, all classes are present.

bruAristimunha:
Not related to the issue with docs! Seems very good for me! many thanks @julien-gadonneix 🙏🏽

bruAristimunha:
My bad, I got confused with another model, I will open again the PR, and I will finish this one for you.

PR #1173, opened 2026-09-21
Add BaRISTA iEEG model

bruAristimunha:
Adds BaRISTA, an intracranial EEG encoder/classifier with temporal patch tokenization and joint space-time rotary attention. Restores @julien-gadonneix's contribution from #1171 after its accidental merge and revert in #1172.

Spatial indices come from dataset metadata. The merged [nemarDatasets/nm000253#3](https://github.com/nemarDatasets/nm000253/pull/3) supplies Brain Treebank's coordinate, parcel and lobe indices. The model contains no atlas-name tables or runtime dataset downloads. MNE-coordinate binning is an optional fallback; released weights should use the dataset's indices.

Includes #1175: callers can supply each recording's indices to `forward`; mean pooling accepts different channel counts and window lengths. Learned pooling requires a fixed total token count. Spatial tables and optional defaults are built by one helper. The model reuses `PatchTokenizer`, `GatedLinearUnit` and native RMSNorm; #1174 supplies the PyTorch >= 2.4 minimum.

`scripts/convert_barista_weights.py` downloads all three official checkpoints from pinned source revision `83b27375eba60e9eba9da4e7dd8fb283baace376`, verifies SHA-256 hashes, renames tensors and fuses the gated projections. Every learned encoder tensor must load; only downstream pooling/classification tensors may be missing. Rotary-frequency buffers are checked before omission. The script compares encoder tokens with the released forward equations and saves local Hub directories, license notices and a JSON report. These models load through `BaRISTA.from_pretrained(..., strict=True)`.

Adds the authors' original overview figure and rewrites the model documentation around its inputs, conversion steps and validation limits. The USC Academic License note stays brief; full terms are in `NOTICE.txt`. No dedicated `test_barista.py` is added.

Validation:
- All three releases: 149 learned encoder tensors loaded for coordinates, 147 each for parcels and lobes. CPU float32 maximum token differences from the independent equation check were 4.41e-6, 8.52e-6 and 6.20e-6 on PyTorch 2.4.0 (atol=1e-5, rtol=1e-4).
- The actual upstream model classes produced identical tokens after explicit CPU adaptations for xformers attention and CUDA-only spatial allocation. This does not validate the CUDA xformers kernel.
- PyTorch 2.6.0 also passed with mean pooling, two channels and 1,025 samples. Strict local Hub reloads reproduce logits exactly. Missing encoder tensors, unexpected tensors and changed rotary frequencies are rejected.
- Spatial helper simplification: eight configurations matched the previous implementation exactly; variable montages, scripting, gradients and twelve invalid-input cases passed. Dataset-index config/Hub/script/export checks passed.
- Three existing BaRISTA feature tests, all configured pre-commit checks, standalone converter lint and a focused Sphinx documentation build passed.
- Applied avoid-ai-writing and tom-neurips-review: removed repetitive Operations/Role phrasing, corrected the fixed-token-count description, and made the checkpoint validation limits explicit.

The original releases contain no downstream head. Converted heads are newly initialized and need fine-tuning. No raw dataset evaluation, downstream accuracy, GPU or mixed-precision equivalence is claimed; masked pretraining is outside this implementation.


braindecode/models/barista.py
```diff
@@ -0,0 +1,1168 @@
...
+        Returns
+        -------
+        torch.Tensor
+            Class logits of shape ``(batch, n_outputs)``.
+        """
+        if x.shape[1] != self.n_chans_grid:
+            raise ValueError(
+                f"BaRISTA was built for {self.n_chans_grid} channels but got input "
+                f"with {x.shape[1]}; rebuild the model for this montage."
+            )
```
julien-gadonneix: Maybe we could remove this check as the model is supposed to accept different subjects with different `n_chans`

braindecode/models/barista.py
```diff
@@ -0,0 +1,1168 @@
...
+        self.n_layers = n_layers
+        self.num_heads = num_heads
+        self.mlp_ratio = mlp_ratio
+        self.pooling = pooling
+        self.drop_prob = drop_prob
+        self.activation = activation
+
+        # Kept as plain ints because EEGModuleMixin hides its signal properties
+        # from TorchScript, so a scripted forward cannot read self.n_chans.
+        self.n_chans_grid = self.n_chans
```
julien-gadonneix: Then I think we can remove this

braindecode/models/barista.py
```diff
@@ -0,0 +1,1168 @@
...
+
+        self.spatial_emb = self._build_spatial_embedding(spatial_scale, spatial_regions)
+
+        # All channels of a patch share one rotary position, so the temporal
+        # index is constant over each block of n_chans consecutive tokens.
+        self.register_buffer(
+            "position_ids",
+            torch.arange(self.n_patches).repeat_interleave(self.n_chans),
+            persistent=False,
+        )
```
julien-gadonneix: And we should keep the `repeat_interleave` operation for the `forward` method

braindecode/models/barista.py
```diff
@@ -0,0 +1,1168 @@
...
+        self.temporal_encoder = _DilatedConvEncoder(
+            patch_size=patch_size,
+            hidden_channels=cnn_channels,
+            depth=cnn_depth,
+            kernel_size=cnn_kernel_size,
+        )
+        self.unfold_grid = Rearrange("batch 1 seq time -> batch seq time")
+        self.temporal_pooler = nn.Linear(patch_size, d_model, bias=False)
+
+        self.spatial_emb = self._build_spatial_embedding(spatial_scale, spatial_regions)
```
julien-gadonneix:
```suggestion
        self.spatial_emb = self._build_spatial_embedding(spatial_scale)
```
This then could also be defined in the `forward`

braindecode/models/barista.py
```diff
@@ -0,0 +1,1168 @@
...
+            raise ValueError(
+                f"spatial_regions has {len(spatial_regions)} labels but the model "
+                f"has {self.n_chans} channels."
+            )
+        indices = _region_indices(list(spatial_regions), vocabulary)
+        # Slot 0 is the reference's UNKNOWN, kept as the padding index so that
+        # electrodes of unknown region contribute nothing.
+        return _SpatialEmbedding(
+            indices.unsqueeze(0), self.d_model, n_regions, padding_idx=0
+        )
```
julien-gadonneix: I think for this function we should just define the embeddings table here
And then select the indices in the `forward`

